PR #8, opened 2020-09-29
Revert "Use a cmake target alias"

Capital-Asterisk:
Reverts TheOpenSpaceProgram/osp-magnum#7

It seems ENTT thinks the build configuration is pretty crazy; I guess they're probably right. Further changes are needed elsewhere for this to work.

```
CMake Error at 3rdparty/entt/CMakeLists.txt:12 (message):
  Prevented in-tree built.  Please create a build directory outside of the
  source code and call cmake from there.  Thank you.
```


jonesmz:
I recommend against this.

You should always do your build directory outside of your source directory.

That being said, entt shouldn't be checking that.

Capital-Asterisk:
I assumed the build directory was specified in the CMakeLists, seems I'm wrong about that. This means it's required to specify a build directory manually, such as `cmake -B build`?

The build and run instructions on the wiki needs to be updated then.

Should I revert this pull request again, or is there a better way?

jonesmz:
When I'm not using an IDE (Personally, on Linux, I use QtCreator, but that's mostly out of long standing habit, not because it's the best or anyhting), I build and configure projects like this:

```
cmake -DCMAKE_BUILD_TYPE=Release -B build
cmake --build build --parallel --config Release
```

I don't think any CMake aware IDE ever builds in the source directly by default.

So, I think you'll want to edit the wiki and revert this PR.

That being said, the EnTT library shouldn't be checking for this and barfing on it. So maybe open a PR to remove that check from the EnTT upstream?

PR #11, opened 2020-10-02
Shrink the sanitizer flags, fixup cmake minimum version.

jonesmz:


z-adams (approved):
I have no idea what this all means, therefore I approve

Capital-Asterisk (commented):
Is there any specific reason why 3.10 is picked? Of all the 3rd party submodules used, EnTT sets the limit to a minimum version 3.12.4, so perhaps we should use this version number too? I don't know what the rest of the stuff do either, so I'd approve of them as well.

jonesmz:
No specific reason, just thought 3.10 was new enough to be worth it.

3.12.4 is fine with me.

PR #12, opened 2020-10-12
Rework Floating Origin into SysAreaAssociate, Generalize Rigid body translation

Capital-Asterisk:
These changes will rework Floating Origin, where entities containing an ACompFloatingOrigin in the ActiveScene are translated towards the origin of the scene when the viewer moves too far away. This prevents significant floating point precision errors.

The translation code is in SysAreaAssociate, the system class that pairs the ActiveScene with a UCompActiveArea, so that the scene can represent an area in the Universe. When translating the scene, the UCompActiveArea has to be translated too. Call `SysAreaAssociate::area_move` to translate both the UCompActiveArea, and every entity with a ACompFloatingOrigin in the ActiveScene.

Floating origin was implemented previously, but all the old code is removed. `ACompTransform::m_enableFloatingOrigin` no longer works. Use `scene.reg_emplace<ACompFloatingOrigin>(entity);` instead to enable Floating Origin on an entity. This should only be added to level-1 entities, parented to the scene's root.

### Rigid bodies and 'controlled entities'

Entities with an ACompRigidBody are controlled by SysNewton; their ACompTransform is updated each frame. This behaviour of 'transform updated each frame' will definitely end up repeated throughout other parts of the code, now named 'controlled' for the time being. 

It would be best if there was a common way to modify the transform from elsewhere while making the controlling system aware of the change. A few new bool members were added to ACompTransform for this purpose:
* m_controlled - Set true from a controlling system to control the entity
* m_mutable - Set true from a controlling system to allow other systems to modify m_transform 
* m_transformDirty Set true from elsewhere, to notify the controlling system when m_transform is modified

SysAreaAssociate uses this interface to translate ACompRigidBody entities without being coupled to SysNewton.

~~Setting a rigid body's transform will modify it AFTER Newton has calculated a new transform for it. This means a frame will be lost, which will cause a bit of a 'twitch' during floating origin translations. SysNewton deals this a little hackily by saving the previous transform of rigid bodies. When transform is set dirty, it will apply the difference between its previous and new/desired transform.~~

Previously, physics was drawn 1 frame behind, as SysNewton will update rigid body's ACompTransform with that of the previous frame. A loop after the physics world update was added to update ACompTransform separately.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -404,7 +404,7 @@ void create_solar_system()
     auto &stationary = uni.trajectory_create<universe::TrajStationary>(
                                         uni, uni.sat_root());
 
-    for (int i = 0; i < 0; i ++)
+    for (int i = 0; i < 10; i ++)
     {
         // Creates a random mess of spamcans
         universe::Satellite sat = debug_add_random_vehicle("TestyMcTestFace Mk"
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -268,8 +268,7 @@ void magnum_application()
     auto &cameraComp = scene.get_registry().emplace<active::ACompCamera>(camera);
 
     cameraTransform.m_transform = Matrix4::translation(Vector3(0, 0, 25));
-    //scene.reg_emplace<active::ACompFloatingOrigin>(camera);
-    //cameraTransform.m_enableFloatingOrigin = true;
+    scene.reg_emplace<active::ACompFloatingOrigin>(camera);
 
     cameraComp.m_viewport
             = Vector2(Magnum::GL::defaultFramebuffer.viewport().size());
@@ -404,9 +403,10 @@ void create_solar_system()
     auto &stationary = uni.trajectory_create<universe::TrajStationary>(
                                         uni, uni.sat_root());
 
+    // Create 10 random vehicles
     for (int i = 0; i < 10; i ++)
     {
-        // Creates a random mess of spamcans
+        // Creates a random mess of spamcans as a vehicle
         universe::Satellite sat = debug_add_random_vehicle("TestyMcTestFace Mk"
                                                  + std::to_string(i));
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -268,7 +268,8 @@ void magnum_application()
     auto &cameraComp = scene.get_registry().emplace<active::ACompCamera>(camera);
 
     cameraTransform.m_transform = Matrix4::translation(Vector3(0, 0, 25));
-    cameraTransform.m_enableFloatingOrigin = true;
+    //scene.reg_emplace<active::ACompFloatingOrigin>(camera);
+    //cameraTransform.m_enableFloatingOrigin = true;
 
     cameraComp.m_viewport
             = Vector2(Magnum::GL::defaultFramebuffer.viewport().size());
@@ -403,7 +404,7 @@ void create_solar_system()
     auto &stationary = uni.trajectory_create<universe::TrajStationary>(
                                         uni, uni.sat_root());
 
-    for (int i = 0; i < 20; i ++)
+    for (int i = 0; i < 0; i ++)
     {
         // Creates a random mess of spamcans
         universe::Satellite sat = debug_add_random_vehicle("TestyMcTestFace Mk"
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -105,6 +105,18 @@ void SysAreaAssociate::disconnect()
     m_areaSat = entt::null;
 }
 
+void SysAreaAssociate::area_move(Vector3s translate)
+{
+    auto &areaPosTraj = m_universe.get_reg()
+            .get<universe::UCompTransformTraj>(m_areaSat);
+
+    areaPosTraj.m_position += translate;
+
+    Vector3 meters = Vector3(translate) / gc_units_per_meter;
+
+    floating_origin_translate(-meters);
+}
+
 void SysAreaAssociate::sat_transform_update(ActiveEnt ent)
 {
     auto const &entAct = m_scene.reg_get<ACompActivatedSat>(ent);
@@ -118,7 +130,8 @@ void SysAreaAssociate::sat_transform_update(ActiveEnt ent)
             .get<universe::UCompTransformTraj>(m_areaSat);
 
     // 1024 units = 1 meter
-    Vector3s posAreaRelative(entTransform.m_transform.translation() * 1024.0f);
+    Vector3s posAreaRelative(entTransform.m_transform.translation()
+                             * gc_units_per_meter);
 
 
 
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -10,8 +10,8 @@ using namespace osp;
 SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni) :
        m_scene(rScene),
        m_universe(uni),
-       m_updateScan(rScene.get_update_order(), "areascan", "", "",
-                    std::bind(&SysAreaAssociate::update_scan, this))
+       m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
+                    [this] { this->update_scan(); })
 {
 
 }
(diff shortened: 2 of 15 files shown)


src/osp/Active/SysNewton.cpp
```diff
@@ -9,35 +9,47 @@ using namespace osp::active;
...
+        // difference with the previous transform. This ensures that the changes
+        // to the rigid body from Newton this frame will still apply.
+
+        // TODO: there's probably a better way to do this. if not, account for
+        //       rotation too, since this is translation only.
 
-        matrix[3].xyz() += scene.floating_origin_get_total();
+        Vector3 diffTranslate = transform.m_transform.translation()
+                              - bodyComp->m_prevTransform.translation();
+        matrix.translation() += diffTranslate;
```
z-adams: This callback runs before the actual position is updated, correct? Otherwise this would cause a frame of motion to be dropped.

src/osp/Satellites/SatActiveArea.h
```diff
@@ -32,6 +32,9 @@ struct UCompActiveArea
     //active::ActiveEnt m_camera;
 
     unsigned m_sceneIndex;
+
+    // true when the ActiveArea is moving
+    bool m_inMotion;
```
z-adams, line 37: Does "ActiveArea is moving" mean a discrete movement or that the frame of reference is actually moving with respect to the system frame?

src/osp/Active/SysNewton.cpp
```diff
@@ -19,37 +19,19 @@ void cb_force_torque(const NewtonBody* body, dFloat timestep, int threadIndex)
...
+        NewtonBodyGetMatrix(body, matrix.data());
 
-        Vector3 diffTranslate = transform.m_transform.translation()
-                              - bodyComp->m_prevTransform.translation();
-        matrix.translation() += diffTranslate;
-
-        NewtonBodySetMatrix(body, matrix.data());
+        // Set matrix without Newton 'conveniently' sleeping for a frame
+        NewtonBodySetMatrixNoSleep(body,
+                                   matrix.data());
```
z-adams, line 30: What happened to the origin translation? I don't see who is responsible for the transformation now

src/osp/Active/SysNewton.cpp
Capital-Asterisk: Yes, but this was janky and removed at the latest commit. The positions are now updated after the physics update in the update_world() function in a separate loop.

src/osp/Satellites/SatActiveArea.h
Capital-Asterisk: This bool doesn't do anything yet. It is suppose to mean that the entire scene frame of reference is moving through the universe. This will likely be removed in favour of a more descriptive enum enum of possible states an active area can be in.

The two possible states planned is:
* Locked to surface: frame of reference is locked to the surface of a planet, and will rotate with it.
* In Deep space: Frame of reference can have a velocity. Orbital velocity can be like 9000m/s; without a velocity, a floating origin translation has to be done every frame.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: The floating origin stuff was removed from SysNewton, and generalized into setting transform. The translation is now done in SysAreaAssociate, and is requested from DebugCameraController.

z-adams (approved):

src/main.cpp
```diff
@@ -403,7 +404,7 @@ void create_solar_system()
     auto &stationary = uni.trajectory_create<universe::TrajStationary>(
                                         uni, uni.sat_root());
 
-    for (int i = 0; i < 20; i ++)
+    for (int i = 0; i < 10; i ++)
```
jonesmz, line 407: magic number?

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -148,6 +147,32 @@ void SysAreaAssociate::activator_add(universe::ITypeSatellite const* type,
     m_activators[type] = &activator;
 }
 
+
+void SysAreaAssociate::floating_origin_translate(Vector3 translation)
+{
+    auto view = m_scene.get_registry()
+            .view<ACompFloatingOrigin, ACompTransform>();
+
+    for (ActiveEnt ent : view)
```
jonesmz, line 156: instead of iterating by value, does it make sense to iterate by const& ?

src/osp/Active/SysNewton.cpp
```diff
@@ -9,33 +9,27 @@ using namespace osp::active;
 
 void cb_force_torque(const NewtonBody* body, dFloat timestep, int threadIndex)
 {
-    SysNewton* sysNewton = static_cast<SysNewton*>(
+    auto* sysNewton = static_cast<SysNewton*>(
```
jonesmz, line 12: Static cast sounds risky.

Not dynamic cast?

src/osp/Active/SysNewton.cpp
```diff
@@ -9,33 +9,27 @@ using namespace osp::active;
...
     ActiveScene &scene = sysNewton->get_scene();
 
-    ACompNwtBody *bodyComp
-            = static_cast<ACompNwtBody*>(NewtonBodyGetUserData(body));
+    auto *bodyComp = static_cast<ACompNwtBody*>(NewtonBodyGetUserData(body));
     DataPhyRigidBody &bodyPhy = bodyComp->m_bodyData;
 
-    //Matrix4 matrix;
+    auto& transform
+            = scene.get_registry().get<ACompTransform>(bodyComp->m_entity);
```
jonesmz, line 20: Why the new line?

src/osp/Active/SysNewton.cpp
```diff
@@ -200,6 +194,9 @@ void SysNewton::create_body(ActiveEnt entity)
 
         // Set inertia and mass
         NewtonBodySetMassMatrix(entBody.m_body, entBody.m_bodyData.m_mass, 1, 1, 1);
+
+        Vector3 tmpOset{0.0f, 0.0f, 0.0f};
+        NewtonBodySetCentreOfMass(entBody.m_body, tmpOset.data());
```
jonesmz, line 199: NewtonBodySetCentreOfMass(entBody.m_body, Vector3{0.0f, 0.0f, 0.0f}.data());

src/osp/Active/activetypes.h
```diff
@@ -27,6 +27,11 @@ using UpdateOrderHandle = FunctionOrderHandle<void(void)>;
 using RenderOrder = FunctionOrder<void(ACompCamera const&)>;
 using RenderOrderHandle = FunctionOrderHandle<void(ACompCamera const&)>;
 
+struct ACompFloatingOrigin
+{
+    bool m_dummy;
```
jonesmz, line 32: ?

src/osp/types.h
```diff
@@ -27,6 +27,10 @@ using Magnum::Rad;
 // An int for space
 typedef int64_t SpaceInt;
 
+// 1024 space units = 1 meter
+// TODO: this should vary by trajectory, but for now it's global
+const float gc_units_per_meter = 1024.0f;
```
jonesmz, line 32: constexpr

src/osp/Active/SysAreaAssociate.cpp
z-adams: ActiveEnt is just a `uint32_t` so I don't think there's any benefit

jonesmz: Makes sense.

Capital-Asterisk:
The biggest mistake here is that rigid body translations don't work at all. I thought it was working, because nothing was glitching out :P. Note how cb_force_torque in SysNewton doesn't read the any new transforms applied to it.

src/osp/Active/activetypes.h
z-adams: Ideally there would be some sort of label or tag concept in EnTT that allows you to tag entities without actually requiring a stored component, but I'm not sure if such a thing exists and if not there's not much choice but to include a dummy member to make it a valid struct.

Capital-Asterisk: EnTT supports tag structs just fine, just that it happened to not compile for this project specifically. I decided to work around it for in that case.

I looked throught the erorrs, and the compile issues are caused by the functions reg_emplace and reg_get in ActiveScene. They're short hands for calling emplace and get in its EnTT registry. Their return values are references, and changing it to decltype(auto), happens to fix it.

src/main.cpp
Capital-Asterisk: Most of main.cpp is magical. The entire file is like a config, or a simple demonstration of using OSP through a C++ interface. Perhaps a comment explaining the loop is sufficient?

src/osp/Active/SysNewton.cpp
Capital-Asterisk: Newton objects allow a single custom 'userdata' void pointer to be stored in them. NewtonWorlds will always store a pointer to the SysNewton that initialized it, and all NewtonBody objects are created from an already-existing ACompNwtBody. There isn't any chance at this point that different types will be assigned as userdata. I've added a few comments in the latest commits to make this more clear. RTTI had also been avoided, and isn't used anywhere in the code.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: There sort of is an 80 column limit rule, but it isn't very well enforced. Perhaps this calls upon more complete coding guidelines.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: This is temporary code used to test something z-adams was trying, removed in the latest commit.

PR #15, opened 2020-11-18
Add instructions to build for Visual Studio and add TOML control configuration

EhWhoAmI:
Wasn't obvious spent 2 hours on it, so I'm trying to save people time with this.


EhWhoAmI:
Also added config files for key presses. But it's not error redundant, so will have to add it. We could have a function that fixes this

.gitignore
```diff
@@ -10,3 +10,4 @@ compile_commands.json
 CTestTestfile.cmake
 OSP-MAGNUM.cbp
 
+/build/*
```
jonesmz, line 13: Lets not add this at this time.

.gitmodules
```diff
@@ -19,3 +19,6 @@
 [submodule "3rdparty/entt"]
 	path = 3rdparty/entt
 	url = https://github.com/skypjack/entt.git
+[submodule "3rdparty/toml11"]
+	path = 3rdparty/toml11
+	url = https://github.com/ToruNiina/toml11
```
jonesmz, line 24: Add newline at end plz.

README.md
```diff
@@ -42,6 +42,18 @@ Build instructions and more details can be found in the
 * Step 1: Make a space flight simulator
 * Step 2: Bloat it with features
 
+## Compiling for Windows on Visual Studio
+You will need: cmake, Visual Studio
+* Clone the repo with `https://github.com/TheOpenSpaceProgram/osp-magnum`
+* Open the repo `cd osp-magnum`
+* Get submodules `git submodule init && git submodule update`
```
jonesmz, line 49: I recommend

```
git submodule update --init --recursive --depth=1
```

Alternatively, you can also do an all in one:

```
git clone --recursive --shallow-submodules https://github.com/TheOpenSpaceProgram/osp-magnum
```

README.md
```diff
@@ -42,6 +42,18 @@ Build instructions and more details can be found in the
...
+* Create the build folder `mkdir build && cd build`
+* Download [SDL2 visual studio development binaries](https://www.libsdl.org/release/SDL2-devel-2.0.12-VC.zip), and extract it into the build folder
+* Run the command `cmake -DCMAKE_PREFIX_PATH=SDL2-2.0.12 -DSDL2_LIBRARY_RELEASE=SDL2-2.0.12/lib/x64/*.lib -DSDL2_INCLUDE_DIR=SDL2-2.0.12/include ..` to create project files
```
jonesmz, line 52: I recommend doing this instead: (Note,  this is bash syntax. Not sure on powershell)

SDLFOLDER=SDL2-2.0.12
cmake -DCMAKE_PREFIX_PATH=$SDLFOLDER -DSDL2_LIBRARY_RELEASE=$SDLFOLDER/lib/x64/*.lib -DSDL2_INCLUDE_DIR=$SDLFOLDER/include ..-DSDL2_INCLUDE_DIR=$SDLFOLDER/include ..

README.md
```diff
@@ -42,6 +42,18 @@ Build instructions and more details can be found in the
...
+* Open the solution `OSP-MAGNUM.sln` in file explorer, and visual studio should open up
+* Shift-control-B to build for debug. 
+* Copying files TODO
```
jonesmz, line 55: @EhWhoAmI Maybe you'd like to add some logic to cmake to make it automatically copy the files in question, instead of the developer having to do it manually?

src/main.cpp
```diff
@@ -372,7 +371,7 @@ void config_controls()
              {0, (int) Key::C, VarTrig::PRESSED, false, VarOp::OR},
              {0, (int) Key::LeftShift, VarTrig::HOLD, false, VarOp::AND},
              {0, (int) Key::A, VarTrig::PRESSED, false, VarOp::OR}});
-
+
```
jonesmz, line 374: Extra whitespace

src/osp/Resource/AssetImporter.cpp
```diff
@@ -350,5 +350,11 @@ void AssetImporter::proto_add_obj_recurse(TinyGltfImporter& gltfImporter,
     }
 }
 
+    void AssetImporter::load_text_to_toml(std::string const& filePath, Package& package)
```
jonesmz, line 353: filePath should be of type std::filesystem::path

jonesmz:
Good work so far! I left you some feedback, and added some other folks as reviewers. I'm excited to see this get merged in a few days!

src/osp/UserInputHandler.h
```diff
@@ -352,6 +372,9 @@ class UserInputHandler
     std::vector<ButtonMap::iterator> m_btnReleased;
 
     //std::map<std::string, int> m_controlActive;
+
+    std::vector<ButtonVarConfig> parse_config_string(std::string text);
```
jonesmz, line 376: probably you want this to be an std::string_view, instead of an std::string

.gitignore
```diff
@@ -9,4 +9,3 @@ install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
 OSP-MAGNUM.cbp
-
```
jonesmz, line 12: Please don't modify the file if you aren't making changes to it.

My advice is to use "git diff" prior to submitting the code just as a sanity check to avoid unexpected changes.

3rdparty/CMakeLists.txt
```diff
@@ -38,3 +38,5 @@ SET(NEWTON_BUILD_SANDBOX_DEMOS OFF)
 ADD_SUBDIRECTORY(newton-dynamics)
 
 ADD_SUBDIRECTORY(entt)
+
+ADD_SUBDIRECTORY(toml11)
```
jonesmz, line 42: newline at end please

README.md
```diff
@@ -42,6 +42,18 @@ Build instructions and more details can be found in the
 * Step 1: Make a space flight simulator
 * Step 2: Bloat it with features
 
+## Compiling for Windows on Visual Studio
+You will need: cmake, Visual Studio
+* Clone the repo with `https://github.com/TheOpenSpaceProgram/osp-magnum`
```
jonesmz, line 47: You can re-write these three lines as

* Clone the repo with `git clone --recursive --shallow-submodules https://github.com/TheOpenSpaceProgram/osp-magnum`
* Navigate to the repo `cd osp-magnum`

src/osp/UserInputHandler.cpp
```diff
@@ -455,4 +466,113 @@ void UserInputHandler::scroll_delta(Vector2i offset)
...
+        config.m_device = osp::sc_mouse;
+    }
+    else if (text == "LMouse") {
+        config.m_devEnum = (int)Mouse::Left;
+        config.m_device = osp::sc_mouse;
+    }
+    else if (text == "MMouse") {
+        config.m_devEnum = (int)Mouse::Middle;
+        config.m_device = osp::sc_mouse;
+    }
```
z-adams, line 565: Switch statement might be more appropriate here?

jonesmz: Unfortunately, C++ doesn't support switch-case for string values.

But what you COULD do, is this.
```
enum class EButton : std::uint8_t
{
    RShift,
    LShift,
    RCtrl,
    LCtrl,
    ....
    Unknown
};
std::string_view button_to_string(EButton button)
{
    switch(button)
    {
        case EButton::RShift: return "RShift";
        case EButton::LShift: return "LShift";
        case EButton::RCtrl: return "RCtrl";
        ....
        default: return "Unknown";
    }
}
EButton string_to_button(std::string_view button)
{
    if(button == "RShift") return EButton::RShift;
    if(button == "LShift") return EButton::LShift;
    if(button == "RCtrl") return EButton::RCtrl;
    return EButton::Unknown;
}
```

jonesmz: There's also various "Magic Enum" libraries out there that might be worth taking a look at.

README.md
z-adams: There ought to be a better way, it's basically a quirk of how VS decides to structure its project data. The binaries from the 3rd party libraries have to be copied to the Build/ directory where VS puts `osp-magnum.exe`, we just have to figure out how to get VS to use the right directories; I can try to figure it out this weekend. On linux it all just works out of the box.

z-adams: I found the solution. I'm not sure why VS treats the projects differently, but osp-magnum doesn't follow the normal CMake build paths.

By default, the variable `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is set to `<osp-magnum>/build/$<CONFIG>/bin`, where `$<CONFIG>` is whatever build configuration you're using in VS (e.g. Debug, Release). If you change the path to `/build/bin/$<CONFIG>` the binaries will all be built in the right place.

Before it would produce a directory structure like
```
build/
|--  bin
|     |-- Debug
|     |     +-  osp-magnum.exe
|     |-- Release
|           +- osp-magnum.exe
|-- Debug
|    |-- bin
|    |      +- <binaries>
|    |-- lib
|           +- <libraries>
|-- Release
     |-- <same as Debug>
```
so by switching the order of `$<CONFIG>/bin` to `bin/$<CONFIG>` it matches the osp-magnum project path and everything goes into the right folder

z-adams:
If you're doing a PR like this, people are going to have a bunch of comments about little things like formatting and naming, so rather than making a commit for every single change, you can just make the changes locally and push them as a single commit. Obviously if someone suggests real changes to source code you'll want to separate things, but your last 4 commits could have been 1 or 2 without being confusing. Maybe @jonesmz will disagree but imo for single-line changes it keeps the commit history cleaner to lump them together into a small "PR fixes" commit.

src/osp/UserInputHandler.cpp
EhWhoAmI: I could make a function that converts it back and forth, and I don't think we need an extra enum, because `OSPMagnum::KeyEvent::Key` already provides what we want, we just have to convert it into an int.

Capital-Asterisk: A map of [string] to structs of {device, enum} is the better to use here. Also good to use that name `string_to_button` as `get_button` sounds like a getter. Something like this should work.

```cpp
struct DeviceButton
{
    int deviceId, int deviceEnum;
}
// ...

// pretend this is a class member
std::map<std::string, DeviceButton> m_strToButton = {
    {"RMouse", {osp::sc_mouse, Mouse::Right}}
    {"LMouse", {osp::sc_mouse, Mouse::Left}}
}

ButtonVarConfig string_to_button(std::string_view buttonName)
{
    // ... something

    auto foundButton = m_strToButton.find(buttonName);

    if (/* device button not found */)
    {
        // do some error thing
    }

    DeviceButton &button = foundButton->second;

    return ButtonVarConfig(button.deviceId, button.deviceEnum, ButtonVar::VarTrigger::HOLD,
                           false, ButtonVar::VarOperator::AND);

}

```

Also consider how other kinds of input devices, like game controllers, can be configured some time in the future. A map would also allow for new keywords can also be added at runtime.

For example, a configuration system dealing with a PS4 controller can do something like this:
1. Read **settings.toml**, and see that the user has configured Controller1 as a PS4 Controller on [OS-specific identifier]
2. Tell some other system to listen to this device in-game, and feed event_raw calls to UserInputHandler
3. Read PS4Controller configs, This contains the device enums
4. Add new keywords to the UserInputHandler Loader's map, such as "Controller1-DPadDown"
5. Read the controls, which can now support stuff like:
```toml
[vehicle_pitch_up]
primary = "S"
secondary = "Controller1-DPadDown"
```

Capital-Asterisk:
About the UserInputHandler: loading files is not one of its duties. Making it load things bloats the class, and makes it rely on AssetImporter. The class should be kept relatively simple and entirely configurable from the outside using plain C++ functions.

All the loading functions: `save_config`, `load_config`, 'parse_config_string', and `get_button` should be separated into their own class belonging in the Resource folder, maybe name it ControlConfigLoader or something. The Loader should be managed by something application-level, like OSPMagnum or main.cpp.

When saving the the config into settings.toml, don't even bother. Converting UserInputHandler's controls back into keyboard/mouse strings will be buggy hell. UserInputHandler controls is the final transformation of config data. Kind of like a compiled binary, where all the human-friendly control configs are converted into machine-only operations.

Editing settings.toml is the job of a todo-later settings gui which should instead edit settings.toml as a toml, without having to parse them into UserInputHandler controls.

jonesmz:
> If you're doing a PR like this, people are going to have a bunch of comments about little things like formatting and naming, so rather than making a commit for every single change, you can just make the changes locally and push them as a single commit. Obviously if someone suggests real changes to source code you'll want to separate things, but your last 4 commits could have been 1 or 2 without being confusing. Maybe @jonesmz will disagree but imo for single-line changes it keeps the commit history cleaner to lump them together into a small "PR fixes" commit.

Typically I advocate for squashing all of the commits of a PR into one, or a very small number, of "logical" changes. E.g. each commit should represent a single, complete, change that the project can compile and run properly with or without. PRs that have more than one commit should represent a "series of logical, complete, changes" that build on each other. Ideally, a PR for a commit series should be set up in such a way that if the reviewers of the PR say "Well, out of your N changes, we only want to merge the first N/2 right now", then that's considered great progress, and the commits after N/2 can be dropped (for now) without much issue and re-worked and re-proposed soon after.

EhWhoAmI:
So, I made `osp/Resources/settings.h`, which consists of a settings class. This will contain all the configuration, and the code to load it to and from file.
A settings class would allow a more robust system, where we do not have to rely on the existence of a settings file.

src/osp/UserInputHandler.cpp
```diff
@@ -455,4 +466,113 @@ void UserInputHandler::scroll_delta(Vector2i offset)
...
+    else if (text.length() == 1) {
+        //Ascii
+        int ascii = (int)text.at(0);
+        if (isalpha(ascii)) {
+            ascii = tolower(ascii);
+        }
+       config.m_devEnum = ascii;
```
z-adams, line 572: I hate to break it to you but this isn't going to work, the key enums can't be assumed to map to ASCII values. The `Magnum::Key` enums are mapped to the key enums from SDL2 and GLFW or any other OpenGL context manager that's implemented, so this part isn't safe. We'll have to figure out another way to map it.

z-adams (changes requested):

src/osp/UserInputHandler.cpp
EhWhoAmI: That's disappointing.
So the question now is, do I create a whole map for ascii values, because I'm afraid that that would look very ugly.

Capital-Asterisk: It's usually reliable to use the enum:
```cpp
int keyEnum = (int)KeyEvent::Key::A + (tolower(text.at(0)) - 'a');
```
A map of ASCII values is not ugly, it's perfectly fine if it's nicely arranged and has a clear reason to be there. Since ASCII values and keycodes will never change, then a switch(char) is the best option here.

There are many other weird keys that don't have single-character names, and for that; a string to keycode&device map would be practical. See my other comment above on this.

z-adams: Good idea, that should work in most cases, even if it's not watertight. The main issue is that the keycodes map to keys and not letters, so GLFW chooses to call the keycodes by their lowercase ASCII value, and SDL2 chooses to call them by uppercase values. Indexing off of the enum value fixes it for those cases, at least.

README.md
Capital-Asterisk: It's also possible to just set the working directory to `osp-magnum/bin`

z-adams: That's necessary too. I forget how it works on linux since I haven't build the repo there in a while but on windows you need this separate build/ directory and VS creates a bunch of folders for the various build configurations. The issue is that for some reason it separates the main project (osp-magnum) from the rest of the projects in the solution, this change just fixes that issue but you still need to change the working directory to osp-magnum/bin. I think on linux I just run `../build/bin/osp-magnum.exe` from that directory or something.

Capital-Asterisk (changes requested):
See comments above. summary:
* Build instructions can be changed to avoid copying files
* Issues with UserInputHandler::get_button: ascii keycodes and long if-else series
* UserInputHandler doing resource-related things when it really shouldn't. Separate loading into a different class.

src/osp/UserInputHandler.cpp
EhWhoAmI: So this is a solution I came up with, which should take into account normal letters, except for special characters. We may need some sort of way to detect special characters such as &,*,@, etc.
```c++

std::tuple<int, int> settings::parse_key(std::string string)
{
    using namespace osp;
    using Key = OSPMagnum::KeyEvent::Key;
    using Mouse = OSPMagnum::MouseEvent::Button;
    using VarOp = ButtonVarConfig::VarOperator;
    using VarTrig = ButtonVarConfig::VarTrigger;
    
    //Device, key
    std::map<std::string, std::tuple<int, int>> m_strToButton = {
        {"RShift", {(int)osp::sc_keyboard, (int)Key::RightShift}},
        {"LShift", {(int)osp::sc_keyboard, (int)Key::RightShift}},
        {"LCtrl",{(int)osp::sc_keyboard, (int)Key::LeftCtrl}},
        {"RCtrl", {(int)osp::sc_keyboard, (int)Key::RightCtrl}},
        {"LAlt", {(int)osp::sc_keyboard, (int)Key::LeftAlt}},
        {"RAlt", {(int)osp::sc_keyboard, (int)Key::RightAlt}},
        {"Up", {(int)osp::sc_keyboard, (int)Key::Up}},
        {"Down", {(int)osp::sc_keyboard, (int)Key::Down}},
        {"Left", {(int)osp::sc_keyboard, (int)Key::Left}},
        {"Right", {(int)osp::sc_keyboard, (int)Key::Right}},
        {"RMouse", {(int)osp::sc_mouse, (int)Mouse::Right}},
        {"LMouse", {(int)osp::sc_mouse, (int)Mouse::Left}},
        {"MMouse", {(int)osp::sc_mouse, (int)Mouse::Middle}}
    };
    std::tuple<int, int> tup = {-1, -1};
    //Select device
    if (m_strToButton.count(string)) {
        //Contains, so get the button
        tup = m_strToButton[string];
    }
    else if(string.length() == 1){
        //Ascii
        if(isalpha(string.at(0)))
            tup = { (int)osp::sc_keyboard, ((int) Key::A + tolower(string.at(0)) - 'a') };
        else if (isdigit(string.at(0))) {
            tup = { (int)osp::sc_keyboard, ((int)Key::One + tolower(string.at(0)) - '1') };
        }
    }
    return tup;
}
```

z-adams: Honestly I think the best thing to do is just put all the keys in the map. Yeah it's kind of ugly, but it's a one-time thing, is guaranteed to work for all platforms that Magnum implements, and none of this code is performance sensitive or anything. 

In my opinion it's not worth the headache just to avoid 100 lines of keymap definition. Sometime interfaces just need a big list of mapping definitions and there's nothing you can do to avoid it, so it's not the end of the world, and I'm not sure if it's worth it to try and build something clever and programmatic if it's not guaranteed to work for every platform. Circumventing Magnum's abstraction layer is something to be avoided unless absolutely necessary.

Handling the special characters will require this treatment anyways so just putting everything in one big ugly map will simplify things substantially.

Capital-Asterisk: There's a few issues here:

* A new map gets allocated every single time the function is called. This can be fixed my making the map static const. However, the [] operator does not work with const, and will modify maps if a value isn't found. Note the use of 'find' in other parts of the code, where it also searches the map only once.
* "m_" denotes a member variable. "sc_" would be the one for static const. There is a style guide that should have been put somewhere more accessible that we ~~very strictly~~ usually follow:
https://docs.google.com/document/d/1dJW-E6SN41XQuvidHRhp41O_1mJmP90bar1egZ31c10/edit?usp=sharing

I'd suggest adding some tab-aligned spaces to the map initialization to vertically align the columns into a more readable table.

Overall, I think the function is okay for the time being; it's already a couple hundred times faster than JS. It's best care more about the structure of the program vs. the insides of individual functions. Judging by the 'settings::' in the declaration, I'd assume loading functions have already been separated into this. My concern is that It's common in OOP to put everything an object can do in a single class, but here I'd highly recommended separating the loading functions from the loaded data. For example, Magnum has ImageData which can be created by multiple importers like StbImageImporter. ImageData is independent of StbImageImporter, and knows nothing of it.

z-adams:
So I had another realization about keymap stuff when thinking about the number row: the numpad breaks the assumption that there is a direct key-char mapping, so that's something we'll have to take into account as well. Here's basically how I suggest doing things:

```cpp
using keydef_t = std::pair<UserInputHandler::DeviceId, OSPMagnum::KeyEvent::Key>;

std::map<std::string, keydef_t> m_strToButton = {
    {"RShift", {osp::sc_keyboard, Key::RightShift}},
    {"LShift", {osp::sc_keyboard, Key::RightShift}},
    // ...
    // The map stores keys which don't have a clear ordering,
    // including named keys or special characters like []\;',./`
};

constexpr std::array<keydef_t, 26> alphabet = 
{
    {osp::sc_keyboard, Key::A},
    {osp::sc_keyboard, Key::B},
    // ...
    {osp::sc_keyboard, Key::Z}
};

// Allows array to be indexed like k = get_alphakey('A');
constexpr keydef_t get_alphakey(const char c)
{
    static_assert(c >= 'A' && c <= 'Z', "Use uppercase please"); // or handle lowercase 
    return alphabet[c - 'A'];
}

// This can either be indexed like k = numrow(3);
// but I would suggest creating a function like above so it can be used like numrow('3')
constexpr std::array<keydef_t, 10> numrow =
{
    {osp::sc_keyboard, Key::Zero},
    {osp:sc_keyboard, Key::One},
    // ...
    {osp::sc_keyboard, Key::Nine}
};
```
The keypad is the tricky bit, since it has `/*-+` that are considered different from the normal keyboard keys, and also repeats the same numbers as the number row. You'll likely want to just use a separate map for these like with the named keys, since the control config will have to contain something like `KP_3`, `KP_.`, `KP_*`, etc. to encode the keypad buttons in control bindings. If you implement a function that allows the number row to be indexed like `numrow('3')` then all the key names can just be done with strings, which makes parsing everything easiest.

EhWhoAmI:
Closed, will split this into 2 prs

PR #16, opened 2020-12-16
Fix CI Errors

Capital-Asterisk:
This PR will simply change a few template function calls in Universe.h that the CI's compiler did not like.


Capital-Asterisk:
Fixed in a different PR

PR #17, opened 2020-12-18
Refactor main.cpp into a more structured Test Application

Capital-Asterisk:
`main.cpp` is going to become a huge mess as more contributions come in. This PR will:
* Move `main.cpp`, `DebugObject.*`, and `OSPMagnum.*` into `src/test_application`. This makes separation between core components, optional components, and runnable test application very clear.
* Created `src/test_application/universe` directory for functions related to creating test universes
* Separate functions of `main.cpp`
  * `src/test_application/flight.*` for starting the Magnum application initialized to the flight scene, originally the `magnum_application()` function. It is very likely that other ways to initialize the scene will be added, such as a map view or vehicle editor.
  * `src/test_application/universe/simple.*` for creating a solar system
  *  `src/test_application/universe/vehicle.*` for creating vehicles, and possibly for creating a test universe with only vehicles
* Add CLI commands to choose which universe to load, as well as choose which kind of application to run
* Add README for test application


src/test_application/flight.cpp
```diff
@@ -0,0 +1,167 @@
...
+using planeta::universe::SatPlanet;
+
+
+void test_flight(std::unique_ptr<OSPMagnum>& magnumApp,
+                 osp::OSPApplication& ospApp, OSPMagnum::Arguments args)
+{
+
+    // Get needed variables
+    Universe &uni = ospApp.get_universe();
+    SatActiveArea *satAA = static_cast<osp::universe::SatActiveArea*>(
```
jonesmz, line 75: You should assert that a dynamic cast results in a non-nullptr value. For an extra sanity check.

src/test_application/flight.cpp
```diff
@@ -0,0 +1,167 @@
...
+    // Configure the controls
+    config_controls(*magnumApp);
+
+    // Create an ActiveScene
+    osp::active::ActiveScene& scene = magnumApp->scene_add("Area 1");
+
+    // Register dynamic systems for that scene
+
+    auto &sysPhysics = scene.dynamic_system_add<osp::active::SysPhysics>(
+                "Physics");
```
jonesmz, line 92: I'm wary of raw strings like this. Strings are expensive to do comparisons.

Could you make the strings be static global variables so that a comparison of the pointer values is sufficient for unique identification?

src/test_application/flight.cpp
```diff
@@ -0,0 +1,167 @@
...
+    // Add a ACompDebugObject to camera to manage camObj's lifetime
+    scene.reg_emplace<ACompDebugObject>(camera, std::move(camObj));
+
+    // Starts the game loop. This function is blocking, and will only return
+    // when the window is  closed. See OSPMagnum::drawEvent
+    magnumApp->exec();
+
+    // Close button has been pressed
+
+    std::cout << "Magnum Application closed\n";
```
jonesmz, line 155: Probably want something like boostlog or other logging framework. Using std::court is yucky

src/test_application/flight.h
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Start flight scene. This will create an ActiveArea in the universe and start
+ * a Magnum application with an ActiveScene. All the necessary systems will be
+ * registered to this scenes for in-universe flight, such as SysAreaAssociate.
+ *
+ * @param magnumApp [out] Magnum application created
+ * @param OSPApp [in,out] OSP universe and resources to run the application on
+ * @param args [in] Arguments to pass to Magnum
+ */
```
jonesmz, line 38: In this situation you should probably be returning the unique pointer by value, and not using an "out" parameter.

src/test_application/main.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+#include <osp/Satellites/SatVehicle.h>
+
+#include <planet-a/Satellites/SatPlanet.h>
+
+#include <iostream>
+#include <memory>
+#include <thread>
+
+using namespace osp;
```
jonesmz, line 41: Recommend against using namespace osp

src/test_application/main.cpp
```diff
@@ -0,0 +1,350 @@
...
+{
+    osp::universe::Universe &uni = g_osp.get_universe();
+    uni.type_register<osp::universe::SatActiveArea>(uni);
+    uni.type_register<osp::universe::SatVehicle>(uni);
+    uni.type_register<planeta::universe::SatPlanet>(uni);
+}
+
+void debug_print_help()
+{
+    std::cout
```
jonesmz, line 225: Recommend boost program options, or similar arg parsing library.

src/test_application/main.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+void debug_print_update_order()
+{
+    if (!g_ospMagnum)
+    {
+        std::cout << "Can't do that yet, start the magnum application first!\n";
+        return;
+    }
+
+    osp::active::UpdateOrder &order = g_ospMagnum->get_scenes().begin()
```
jonesmz, line 249: Const-ref

jonesmz (approved):

src/test_application/DebugObject.h
```diff
@@ -22,6 +22,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
+
```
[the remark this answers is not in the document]
Capital-Asterisk: I'm following the style guide you've written a few years back; it seems you've changed your mind on this. Which 'weird filesystem setup' specifically does this not work? Is there even a C++17 capable compiler that doesn't support `#pragma once`? Perhaps replacing all the `#pragma once`s in the entire codebase can be automated with a bit of bash in a different PR.

src/test_application/DebugObject.h
```diff
@@ -30,9 +31,6 @@
 #include <osp/UserInputHandler.h>
 #include <osp/types.h>
 
-namespace osp
-{
```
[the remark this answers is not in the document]
Capital-Asterisk: I'll put the entire test application into its own namespace then.

src/test_application/DebugObject.h
```diff
@@ -30,9 +31,6 @@
...
-
```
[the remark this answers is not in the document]
Capital-Asterisk: The "Abstract" prefix in the class was there to match Magnum's naming convention which I somehow decided to follow for the first couple of months for some reason; this file is pretty old. Most of the new code in OSP follows the "I" standard.

The virtual destructors is a bit of a problem though. DebugObject heavily relies on having its lifetime managed by something that doesn't know its implementation. I'm not sure how this can even be fixed, or if it's even worth restructuring this class which is kind of made to be this disgusting. This is not a core component, and the use of this class is discouraged.

The purpose of the DebugObject is to be able to quickly add functionality to entities in the ECS for testing only. One of the downsides of ECS is that it takes a bit more effort to add quick and unique functions like the camera controller. An entirely new system has to be created to implement it. DebugObject along with ACompDebugObject pretty much wraps an OOP-style class into ECS.

Other than that, IDynamicSystem from `osp/Active/activetypes.h` relies on virtual destruction too, and is a topic for future discussion.

src/test_application/DebugObject.h
```diff
@@ -48,23 +46,23 @@ class DebugObject : public AbstractDebugObject
...
 
-struct CompDebugObject
+struct ACompDebugObject
 {
-    CompDebugObject(std::unique_ptr<AbstractDebugObject> ptr) :
+    ACompDebugObject(std::unique_ptr<AbstractDebugObject> ptr) :
         m_obj(std::move(ptr)) {}
-    CompDebugObject(CompDebugObject&& move) = default;
-    ~CompDebugObject() = default;
-    CompDebugObject& operator=(CompDebugObject&& move) = default;
```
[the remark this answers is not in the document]
Capital-Asterisk: Ok, sounds good. Many other ECS component structs in other places have non-constexpr move constructors too; these can be dealt with as well.

src/test_application/OSPMagnum.cpp
```diff
@@ -0,0 +1,209 @@
...
+{
+    m_userInput.mouse_delta(event.relativePosition());
+}
+
+void OSPMagnum::mouseScrollEvent(MouseScrollEvent & event)
+{
+    m_userInput.scroll_delta(static_cast<osp::Vector2i>(event.offset()));
+}
+
+osp::active::ActiveScene& OSPMagnum::scene_add(const std::string &name)
```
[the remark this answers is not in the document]
Capital-Asterisk: There is `const std::string&` all over that can be replaced too. I'll keep this in mind for all future code.

src/test_application/flight.cpp
Capital-Asterisk: This line actually turns out to be a mistake that happens to compile and work. The entire line is:
```cpp
SatActiveArea *satAA = static_cast<osp::universe::SatActiveArea*>(
            uni.sat_type_find("Vehicle")->second.get());
```
It's searching for "Vehicle" even though it's expecting a SatActiveArea. It just happens that none of the subsequent functions don't access any member variables. The Sat* classes already has a get_name function, which means that specifying the name in this function isn't even needed, avoiding these mistakes entirely.

RTTI is avoided and isn't used anywhere. Not even any of the dependencies use it. Perhaps just an assert would do? I might also put the casting into a template and use the get_name function, to be able to do this instead:

```cpp
auto &satAA = uni.sat_type_find<osp::universe::SatActiveArea>()
```

src/test_application/DebugObject.h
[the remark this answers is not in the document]
jonesmz: Using both is better than using only one :)

The weird filesystem issues are things like NFS mounts that map the file to multiple locations.

src/test_application/flight.cpp
Capital-Asterisk: These functions aren't performance critical; there's not much to worry about. This name can be stored as a static constant in SysPhysics and other system classes, similar to the Sat* classes mentioned in the previous comment. The name wouldn't be needed by the function at all:
```cpp
auto &sysPhysics = scene.dynamic_system_add<osp::active::SysPhysics>();
```

src/test_application/DebugObject.h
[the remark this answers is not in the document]
jonesmz: Are you managing the lifetime of it using std::shared_ptr or std::unique_ptr? If so you can specialize the smart pointers with custom deleters that know of a virtual destructor in the hierarchy of the object that can be called.

But if its too complicated then its too complicated.

src/test_application/flight.cpp
Capital-Asterisk: Adding boost log should probably be in a separate PR. I haven't got to doing it yet, as it's a task that I'm thinking a contributor can get easily started on.

src/test_application/DebugObject.h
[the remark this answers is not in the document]
jonesmz: I also recommend annotating functions with noexcept where possible.

That helps with code gen because the compiler doesn't need to worry about exception handling for that function call.

src/test_application/OSPMagnum.cpp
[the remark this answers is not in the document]
jonesmz: Sounds good

src/test_application/main.cpp
Capital-Asterisk: This part of the code is made to be as simple as possible. I'd like to keep this close to plain c++ to make it easy on new contributes.

jonesmz: I mean... Argument parsing libraries make things pretty easy, but its not a big deal either way.

src/test_application/DebugObject.h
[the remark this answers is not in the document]
Capital-Asterisk: The constexpr move constructor doesn't compile. "defaulted definition of move constructor is not constexpr"

src/test_application/OSPMagnum.cpp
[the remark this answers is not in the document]
Capital-Asterisk: This function has to convert the std::string_view to an std::string anyways, so it's probably not ideal for this function in particular.
```cpp
osp::active::ActiveScene& OSPMagnum::scene_add(const std::string &name)
{
    auto pair = m_scenes.try_emplace(name, m_userInput, m_ospApp);
    return pair.first->second;
}
```

src/test_application/DebugObject.h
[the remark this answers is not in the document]
jonesmz: Need to do

constexpr CompDebugObject() = default;

= default; doesn't imply constexpr. (though, i wish it did).

src/test_application/OSPMagnum.cpp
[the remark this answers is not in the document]
jonesmz: Ah, then this would be an "owning parameter usage".

I recommend making two versions of the function then.

One that accepts the string by const&, and the other that accepts it by && + std::move()

That allows the consumer of the function to get move semantics if they want them, which can allow the consumer to skip an allocation-and-copy that they always have to pay for with the const& version.

src/test_application/flight.h
Capital-Asterisk: I'm not sure how well this fits with best practices, but this function is used to start a new thread. It uses the unique_ptr to pass the application it creates back to the main thread.

jonesmz: My point was just that you can change the function from

void foo(std::unique_ptr<>& pOut);

to

std::unique_ptr<> foo(void);

Returning a std::unique_ptr by value is basically free, and literally free if you write the function to take advantage of return value optimization.

Accepting a reference to an std::unique_ptr, and then assigning a value to that passed in reference, can't be optimized out in any meaningful way.

Same outcome, different ways of doing it.

Plus, if you return the std::unique_ptr, you can mark the function as [[nodiscard]] and get a nice warning if the variable isn't used after being received.

src/adera/Machines/Rocket.cpp
```diff
@@ -34,6 +34,8 @@ using namespace adera::active::machines;
 using namespace osp::active;
 using namespace osp;
 
+const std::string SysMachineRocket::smc_name = "Rocket";
```
jonesmz, line 37: With C++17, you don't need to define the variable in the cpp file.

You can use "static inline const std::string SysMachineRocket::smc_name = "Rocket";" in the class body itself.

src/adera/Machines/Rocket.h
```diff
@@ -38,6 +38,8 @@ class SysMachineRocket :
 {
 public:
 
+    static const std::string smc_name;
```
jonesmz, line 41: static inline const std::string smc_name = "Rocket";

Though, that being said for string constants that are used in most situations for comparisons should probably be std::string_view.

Functions that accept some kind of string, but don't need to transfer ownership, should accept a std::string_view.

If ownership needs to be transfered, you can either:

take an std::string by value, and std::move() it into place.
take an std::string by both const& and &&, and then where the caller has an std::string_view, and not an std::string, have the caller construct the std::string that's needed.
If ownership only transfers *sometimes* then also provide an std::string_view overload that only constructs a std::string when absolutely necessary. E.g. for functions that do a lookup against an std::map or similar, use the C++17 (14?) thing "std::less<>" to allow "transparent" comparisons, and then if you need to insert a new item into the map, that's when you'd allocate the string.

src/osp/Active/ActiveScene.cpp
```diff
@@ -258,6 +258,12 @@ void ActiveScene::draw(entt::entity camera)
     m_renderOrder.call(cameraComp);
 }
 
+void ActiveScene::system_machine_add(std::string const& name,
```
jonesmz, line 261: Since this function is always emplacing a new item, i would personally take both of the parameters by value (not const& and not &&, just by value) and std::move them into place.

Additionally, consider having this function either return what m_sysMachines.emplace is returning, or at least have it return a bool indicating success / failure, unless you either

1) Know for certain that it'll only be called when no item already exists.
or
2) don't care if the item already exists.

src/osp/Active/ActiveScene.cpp
```diff
@@ -258,6 +258,12 @@ void ActiveScene::draw(entt::entity camera)
     m_renderOrder.call(cameraComp);
 }
 
+void ActiveScene::system_machine_add(std::string const& name,
+                                     std::unique_ptr<ISysMachine>&& sysMachine)
+{
+    m_sysMachines.emplace(name, std::move(sysMachine));
+}
+
 MapSysMachine::iterator ActiveScene::system_machine_find(const std::string &name)
```
jonesmz, line 267: If you change "m_sysMachines" to have the "std::less<>" comparitor (e.g. transparent comparison), then you can change this function to take an std::string_view instead of an std::string, and still get the same comparison behavior.

src/osp/Active/ActiveScene.h
```diff
@@ -228,32 +237,31 @@ class ActiveScene
...
-template<class T, typename... Args>
-T& ActiveScene::dynamic_system_add(std::string const& name,
-                                             Args &&... args)
+
+template<class DYNSYS_T, typename... ARGS_T>
+DYNSYS_T& ActiveScene::dynamic_system_add(ARGS_T &&... args)
 {
-    auto ptr = std::make_unique<T>(*this, args...);
-    T &refReturn = *(ptr.get());
+    auto ptr = std::make_unique<DYNSYS_T>(*this, args...);
```
jonesmz, line 252: When passing template parameter packs into a function as &&'s, you have to make sure you use std::forward<ARGS_T>(args)...

if you don't, the move constructor won't ever be called, even when you expect it to.

src/osp/Active/ActiveScene.h
```diff
@@ -228,32 +237,31 @@ class ActiveScene
...
+    system_machine_add(SYSMACH_T::smc_name,
+                       std::make_unique<SYSMACH_T>(*this, args...));
 }
 
-template<class T, typename... Args>
-T& ActiveScene::dynamic_system_add(std::string const& name,
-                                             Args &&... args)
+
+template<class DYNSYS_T, typename... ARGS_T>
+DYNSYS_T& ActiveScene::dynamic_system_add(ARGS_T &&... args)
```
jonesmz, line 250: Is returning a reference your intended semantics? Typically my first tool to reach for is std::shared_ptr, instead of raw-references.

Here you're using std::unique_ptr, which you then store in the map, so you can't let the callers of the function have access to that.

Maybe this is just my ignorance of how the rest of the codebase works, but it seemed reasonable to ask about.

src/osp/Active/ActiveScene.h
```diff
@@ -228,32 +237,31 @@ class ActiveScene
...
 {
-    auto ptr = std::make_unique<T>(*this, args...);
-    T &refReturn = *(ptr.get());
+    auto ptr = std::make_unique<DYNSYS_T>(*this, args...);
+    DYNSYS_T &refReturn = *(ptr.get());
```
jonesmz, line 253: Why not just operator*? Not sure I see the point of calling .get().

src/osp/Satellites/SatActiveArea.cpp
```diff
@@ -40,14 +40,12 @@ using namespace Magnum::Math::Literals;
 
 using Magnum::Matrix4;
 
-namespace osp::universe
-{
+using namespace osp::universe;
```
jonesmz, line 43: I recommend against using the "using namespace ..." syntax like this.

Better would be a namespace alias where you give the namespace a shorter name, but don't just import the entire namespace into the current scope.

e.g. 
namespace fbz = foo::bar::baz;

src/osp/Universe.h
```diff
@@ -135,6 +135,14 @@ class Universe
         return m_satTypes.find(name);
     }
 
+    /**
+     * Tries to locate an element in the map of registered Satellite by type.
+     * @tparam SATTYPE_T Type of ITypeSatellite to find
+     * @return Map iterator directly from std::map::find()
```
jonesmz, line 141: Is it returning an iterator? or a reference?

If returning an iterator, i don't think it makes sense to templatize it?

src/osp/Universe.h
```diff
@@ -173,6 +183,14 @@ TYPESAT_T& Universe::type_register(ARGS_T&& ... args)
     return toReturn;
 }
 
+template<typename SATTYPE_T>
+SATTYPE_T& Universe::sat_type_find()
+{
+    MapSatType::iterator it = sat_type_find(SATTYPE_T::smc_name);
+    assert(it != m_satTypes.end());
+    return static_cast<SATTYPE_T&>(*it->second);
```
jonesmz, line 191: assert on a dynamic cast as well. I provided an example in another comment.

src/osp/Universe.h
```diff
@@ -173,6 +183,14 @@ TYPESAT_T& Universe::type_register(ARGS_T&& ... args)
...
+}
+
```
jonesmz, line 193:
```
template<typename TRAJECTORY_T, typename ... ARGS_T>
TRAJECTORY_T& Universe::trajectory_create(ARGS_T&& ... args)
{
    std::unique_ptr<TRAJECTORY_T> newTraj
            = std::make_unique<TRAJECTORY_T>(std::forward<ARGS_T>(args)...);
    TRAJECTORY_T& toReturn = *newTraj;
    m_trajectories.push_back(std::move(newTraj)); // this invalidates newTraj
    return toReturn;
}
```

I feel like this would simpler 

```
template<typename TRAJECTORY_T, typename ... ARGS_T>
TRAJECTORY_T& Universe::trajectory_create(ARGS_T&& ... args)
{
    return *( m_trajectories.emplace_back(std::make_unique<TRAJECTORY_T>(std::forward<ARGS_T>(args)...)) );
}
```

m_trajectories is an std::vector, so the emplace_back function returns a reference to the newly inserted item.

jonesmz, line 193: struct UCompTransformTraj
{
    // Relative to m_parent
    Vector3s m_position;
    // In 'global' space
    Quaternion m_rotation;
    Satellite m_parent; // Set only by trajectory
    ISystemTrajectory *m_trajectory{nullptr}; // get rid of this some day
    unsigned m_index; // index in trajectory's vector
    // set this to true when you modify something the trajectory might use
    bool m_dirty{true};
    // move this somewhere else eventually
    std::string m_name;
};


Probably use size_t instead of "unsigned".

If you don't want size_t, then better would be e.g. std::uint64_t or something similar.

jonesmz: Additionally, consider struct-packing.

you'll always get the least amount of padding added by the compiler if you order your member variables with largest alignas() to smallest alignas(), meaning your m_dirty bool would go last, and std::string very well may go first.

src/test_application/DebugObject.cpp
```diff
@@ -111,22 +126,21 @@ void DebugCameraController::update_physics_pre()
...
+    tra *= osp::gc_units_per_meter;
 
     if (!tra.isZero())
     {
         std::cout << "Floating origin translation!\n";
 
         // Move the active area to center on the camera
-        m_scene.dynamic_system_get<active::SysAreaAssociate>("AreaAssociate")
-                .area_move(-tra);
+        m_scene.dynamic_system_find<SysAreaAssociate>().area_move(-tra);
```
jonesmz, line 143: Using std::negate would be more readable here: https://en.cppreference.com/w/cpp/utility/functional/negate

My eyesight sucks, almost missed that there was a minus sign there.

src/test_application/OSPMagnum.cpp
```diff
@@ -138,11 +139,11 @@ void OSPMagnum::mouseScrollEvent(MouseScrollEvent & event)
 
 osp::active::ActiveScene& OSPMagnum::scene_add(const std::string &name)
 {
-    auto pair = m_scenes.try_emplace(name, m_userInput, m_ospApp);
-    return pair.first->second;
+    auto const& [it, success] = m_scenes.try_emplace(name, m_userInput, m_ospApp);
```
jonesmz, line 142: Consider leaving a comment saying that if the scene is already there, don't want to add a new one, just return the existing one.

Which, that being said, maybe the name of the function should be "acquire_scene" or something similar, so that it doesn't sound like it's allocating a new scene.

src/test_application/main.cpp
```diff
@@ -317,26 +320,28 @@ void debug_print_hier()
...
-    universe::Universe &universe = g_osp.get_universe();
+    osp::universe::Universe const &universe = g_osp.get_universe();
 
-    auto view = universe.get_reg().view<universe::UCompTransformTraj,
-                                        universe::UCompType>();
+    auto const view = universe.get_reg().view<const UCompTransformTraj,
+                                              const UCompType>();
 
-    for (universe::Satellite sat : view)
+    for (osp::universe::Satellite sat : view)
```
jonesmz, line 331: Are osp::Universe::Satellite's cheap to copy-construct? That's what the for-loop is doing for every item in the view.

If not cheap, then recommend const&

z-adams: The "view" object in these situations is a collection of entities that meet a certain criterion based on the components they have, but the iteration is over the entities themselves. The entities themselves (e.g. `Satellite`) are just an ID represented by a `uint32_t` so there's no performance concern when dealing with them.

src/adera/Machines/Rocket.h
jonesmz: To provide some more background information on this.

complex objects, like std::string incur an application startup initialization cost that involves a few mutexes and some runtime operations.

On the other hand, doing

"static constexpr std::string_view smc_name = "MyString";

Puts the string data into the read-only-data section of the binary, which incurs literally zero initialization cost, and allows the machine code to just refer to the address in the ROData section directly (with a single fixup by the runtime linker for position-independent-code related sillyness).

It also enables constant propagation to be more effective, and as a result can allow for same-code-folding to pick up on identical code where the read-only-data string is used that it might not be able to if you were using an std::string.

src/test_application/flight.h
Capital-Asterisk: I should put this in the doxygen block; this function blocks execution until the window is closed. It blocks at `pMagnumApp->exec();` The function will not return anything for quite a while.

src/adera/Machines/Rocket.h
Capital-Asterisk: How well does this work across shared libraries? In a previous comment, you mentioned "comparison of the pointer values," and it really seems like this wouldn't work if the variable is defined in multiple places.

src/test_application/OSPMagnum.cpp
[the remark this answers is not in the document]
Capital-Asterisk: Won't the same functionality be achieved with passing by value, and an std::move() into the emplace function? I'm also going to change the name of the function to "scene_create", as it's initializing a new scene.

```cpp
osp::active::ActiveScene& OSPMagnum::scene_create(std::string name)
{
    auto const& [it, success] = m_scenes.try_emplace(std::move(name), m_userInput, m_ospApp);
    return it->second;
}
```

src/osp/Active/ActiveScene.h
Capital-Asterisk: Yes, the dynamic systems are stored in the scenes. They cannot be moved out of their scene, as they store a reference member to it. Returning an iterator to the dynamic system is probably more practical.

The "System" in ECS refers to a stateless function that iterates entities and components. But here, they're classes with non-deterministic update functions and manage the lifetime of UpdateOrderHandles.

As the code evolves, the m_scene member in most dynamic systems will probably be removed. The update functions can be made static where the ActiveScene can be passed as a function argument. This will turn system classes into just structures of UpdateOrderHandles. Further refactoring can store all the UpdateOrderHandles in a single container in ActiveScene, removing the system class entirely and getting rid of those virtual destructors.

src/osp/Satellites/SatActiveArea.cpp
Capital-Asterisk: This is quite a huge change to ask. `using namespace` is used in pretty much every source file. I've already touched almost every file in this PR, but I feel that this should go on a separate PR.

I'd argue that the use of `using namespace` across the sources are already quite restricted in most cases here:
* only used if a source file is part of the module the namespace implies
* only a few of the included files show up in the scope. Including the entire osp namespace would require manually including each individual files
* generic names aren't used anywhere and are mostly avoided

src/osp/Universe.h
Capital-Asterisk: I can't find any comment with dynamic_cast example code. Again, RTTI is not used anywhere. The only time this will fail is if there is two different satellite types with the same name constant, which can only happen if the program was compiled that way. There is also the possibility of using entt's unique class naming functions instead of string constants.

z-adams:
This all seems fine to me. Unless you want to try and leverage this PR to tackle the DependRes crash due to the resource free order, you can merge whenever you see fit. This will make it much easier to merge our diverging branches into master without being forced to immediately integrate them together.

z-adams (approved):

src/adera/Machines/Rocket.h
jonesmz: Like many things, it depends on many things :-)

Inline variable is going to be a weak symbol in the shared library, so the platform runtime linker might consolidate them all to a single memory address.

Nevertheless, you can't rely on comparing only pointers to determine if two strings have the same contents. That's great for positive identification, but can result in false-negatives.

src/osp/Active/ActiveScene.h
jonesmz: > Returning an iterator to the dynamic system is probably more practical.
Careful about iterator invalidation!

Further, returning a reference does run the risk of the underlying storage being moved as well, so careful about that too.

src/osp/Satellites/SatActiveArea.cpp
jonesmz: Instead of "using namespace osp::universe", do "namespace uni = osp::universe", or something like it.

Don't need to change this for this PR. It can be done some other time.

src/osp/Universe.h
jonesmz: > The only time this will fail is if there is two different satellite types with the same name constant, which can only happen if the program was compiled that way.

How can you statically assert that to ensure it's not possible?

> Again, RTTI is not used anywhere

Is RTTI disabled as a cli option to the compiler?

src/test_application/OSPMagnum.cpp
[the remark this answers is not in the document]
jonesmz: try_emplace implies that emplacement might fail.

If emplacement fails, I, as the consumer of the function, don't want to pay the price of allocating/copying the string parameter.

That's why I said the optimal way to do this is to have 2 functions. In my previous comment I said a const& and a && version, but I realized that std::string_view is a perfect super-set of the const& case, so I recommend that instead.

```
// The caller of the function doesn't have a std::string, so the std::string_view provides a way to call the function
// that only causes an allocation if emplacement will succeed.
osp::active::ActiveScene& OSPMagnum::scene_create(std::string_view name)
{
    auto const& [it, success] = m_scenes.try_emplace(name, m_userInput, m_ospApp);
    return it->second;
}

// In the case where we have an actual std::string, and we know we're done with it (because it's been std::moved() into the function), we can save an allocate and copy by providing this overload.
osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
{
    auto const& [it, success] = m_scenes.try_emplace(std::move(name), m_userInput, m_ospApp);
    return it->second;
}
```

jonesmz:
As far as I'm concerned you can merge this, any additional improvements can always be made later.

PR #19, opened 2020-12-22
Various minor cleanups, nothing substantial

jonesmz:


Capital-Asterisk (approved):
I kind of hate to say this, but I was already thinking of re-doing many parts of the wire system. Most of these changes would probably be removed soon.

The current implementation is not friendly to the cache, and doesn't really use the ECS properly. Machines are large classes containing every wire input and output. WireOutputs use Magnum's linked list to connect to WireInputs. This makes the move constructors rather expensive, as each WireOutput moved has to update every input connected to it.

I haven't fully worked it out, but I should put an issue up to discuss it a bit further.

jonesmz:
Well if you want to merge this PR even if you're planning to rewrite it later, go ahead.

Otherwise you can close it.

I don't mind either way.

z-adams (approved):

PR #20, opened 2020-12-22
Copy graphics data into the bin folder for development convienience.

jonesmz:


Capital-Asterisk (approved):
This fixes @EhWhoAmI's issue in PR #15. I'm kind of against the idea of having duplicate files; but now that I think of it, it adds separation between the source files and the build output.

jonesmz:
Improvements can always be made later. At least this directly allows the binary to be run right away.

PR #24, opened 2020-12-29
Add resource package prefixes

z-adams:
Adds the resource package prefixes Cap planned originally.

I recommend looking at Package.h/Package.cpp first, then OSPApplication.h/.cpp. The rest of the changes are simply updating other code to use the new system. I wish there were a way to order the files in the "files changed" tab so that the logical cascade of changes was easier to follow.

### DIFF
--- a/src/osp/Active/SysVehicle.cpp
+++ b/src/osp/Active/SysVehicle.cpp
@@ -280,9 +280,7 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
 
             // Current prototype is a mesh, get the mesh and add it
 
-            // TODO: put package prefixes in resource path
-            //       for now just get the first package
-            Package& package = m_scene.get_application().debug_get_packges()[0];
+            Package& package = m_scene.get_application().debug_find_package("lzdb");
             const DrawableData& drawable =
                 std::get<DrawableData>(currentPrototype.m_objectData);
 
--- a/src/osp/OSPApplication.cpp
+++ b/src/osp/OSPApplication.cpp
@@ -30,3 +30,18 @@ OSPApplication::OSPApplication()
 {
 
 }
+
+void OSPApplication::debug_add_package(Package&& p)
+{
+    auto const& [it, success] = m_packages.emplace(p.get_prefix(), std::move(p));
+    assert(success);
+}
+
+Package& OSPApplication::debug_find_package(std::string_view prefix)
+{
+    if (auto it = m_packages.find(prefix); it != m_packages.end())
+    {
+        return it->second;
+    }
+    throw std::out_of_range("Package not found");
+}
--- a/src/osp/OSPApplication.h
+++ b/src/osp/OSPApplication.h
@@ -24,6 +24,7 @@
  */
 #pragma once
 
+#include <map>
 #include "Universe.h"
 #include "Resource/Package.h"
 
@@ -39,12 +40,30 @@ public:
 
     OSPApplication();
 
-    std::vector<Package>& debug_get_packges() { return m_packages; };
+    /**
+     * Add a resource package to the application
+     *
+     * The package should be populated externally, then passed via rvalue
+     * reference so the contents can be moved into the application resources
+     *
+     * @param p [in] The package to add
+     */
+    void debug_add_package(Package&& p);
+
+    /**
+     * Get a resource package by prefix name
+     *
+     * @param prefix [in] The short prefix name of the package
+     * @return The resource package
+     */
+    Package& debug_find_package(std::string_view prefix);
+
+    size_t debug_num_packages() const { return m_packages.size(); }
 
     universe::Universe& get_universe() { return m_universe; }
 
 private:
-    std::vector<Package> m_packages;
+    std::map<ResPrefix_t, Package, std::less<>> m_packages;
     universe::Universe m_universe;
 };
 
--- a/src/osp/Resource/Package.cpp
+++ b/src/osp/Resource/Package.cpp
@@ -29,12 +29,27 @@
 namespace osp
 {
 
-Package::Package(std::string const& prefix, std::string const& packageName) :
-    // not sure about this but it compiles
+Package::Package(std::string prefix, std::string packageName) :
     m_groups(),
-    //m_prefix{prefix[0], prefix[1], prefix[2], prefix[3]},
-    m_packageName(packageName)
+    m_prefix(std::move(prefix)),
+    m_packageName(std::move(packageName))
 {
 }
+
+StrViewPair_t decompose_str(std::string_view path, const char delim)
+{
+    size_t pos = path.find(delim);
+    return {
+        path.substr(0, pos),
+        path.substr(pos + 1, path.length())
+    };
+}
+
+Path decompose_str(std::string_view path)
+{
+    StrViewPair_t pair = decompose_str(path, ':');
+    return { pair.first, pair.second };
+}
+
 
 }
--- a/src/osp/Resource/Package.h
+++ b/src/osp/Resource/Package.h
@@ -61,24 +61,47 @@ namespace osp
 
 // supported resources
 
+// Just a string, but typedef'd to indicate that it represents a prefix
+using ResPrefix_t = std::string;
+
+// Stores a split path as <prefix, rest of path>
+struct Path
+{
+    std::string_view prefix;
+    std::string_view identifier;
+};
+
+using StrViewPair_t = std::pair<std::string_view, std::string_view>;
+/**
+ * Split a string_view at the first instance of a delimiter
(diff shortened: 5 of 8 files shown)


src/osp/OSPApplication.cpp
```diff
@@ -30,3 +30,13 @@ OSPApplication::OSPApplication()
 {
 
 }
+
+void OSPApplication::debug_add_package(Package&& p)
```
jonesmz, line 34: Careful with assuming that .emplace() will succeed.

1) Functions that just call a function on a member variable should either A) Return what the called function returns (Or a part of it, like the success / fail) or B) should assert at runtime that the inner call succeeded.
2) Do you really want to accept by rvalue-reference only? If so, that's fine, but idomatically most people assume you'll accept a copy as well as a move. If you need it to be rvalue-reference only, consider explicitly deleting the copy constructor and assignment operator off of the Package class.
3) There's a good chance that you want try_emplace over emplace, because try_emplace only actually move constructs the provided value if-and-only-if the insert operation will succeed.
4) If you don't want try_emplace, and you don't want to assert / return the bool, then maybe you want insert_or_assign to overwrite the existing key (if exists)

src/osp/OSPApplication.cpp
```diff
@@ -30,3 +30,13 @@ OSPApplication::OSPApplication()
...
+{
+    m_packages.emplace(p.get_prefix(), std::move(p));
+}
+
+Package& OSPApplication::debug_get_package(resprefix_t prefix)
```
jonesmz, line 39: resprefix_t is typedefed as std::string

but here you're taking it by value for use with a function that doesn't require ownership semantics.

It's far far more likely that what you want to do here is take an std::string_view (Give it an appropriate typedef if you want), and then use std::map::find(), and if the result of find() == std::map::end(), then throw the exception that std::map::at() would throw.

If you don't want to use string_view (I strongly recommend it for this situation), then at the least, take the argument as const&

src/osp/OSPApplication.h
```diff
@@ -39,12 +40,32 @@ class OSPApplication
 
     OSPApplication();
 
-    std::vector<Package>& debug_get_packges() { return m_packages; };
+    //std::vector<Package>& debug_get_packges() { return m_packages; };
```
jonesmz, line 43: Remove commented out code.

src/osp/OSPApplication.h
```diff
@@ -39,12 +40,32 @@ class OSPApplication
...
+     */
+    Package& debug_get_package(resprefix_t prefix);
+
+    size_t debug_num_packages() const { return m_packages.size(); }
 
     universe::Universe& get_universe() { return m_universe; }
 
 private:
-    std::vector<Package> m_packages;
+    std::map<resprefix_t, Package> m_packages;
```
jonesmz, line 68: std::map<resprefix_t, Package, std::less<>>

src/osp/Resource/Package.cpp
```diff
@@ -32,9 +32,18 @@ namespace osp
 Package::Package(std::string const& prefix, std::string const& packageName) :
```
jonesmz, line 32: Most likely you want these constructor parameters to be passed in by value and then moved, not by const&.

src/osp/Resource/Package.cpp
```diff
@@ -32,9 +32,18 @@ namespace osp
...
     // not sure about this but it compiles
     m_groups(),
-    //m_prefix{prefix[0], prefix[1], prefix[2], prefix[3]},
+    m_prefix(prefix),
     m_packageName(packageName)
 {
 }
 
+path_t decompose_path(std::string const& path)
```
jonesmz, line 40: std::string_view. The caller of the function may not inherently have an std::string that they can pass in. They might have some other string type.

src/osp/Resource/Package.h
```diff
@@ -61,12 +61,26 @@ namespace osp
 
 // supported resources
 
+// Just a string, but typedef'd to indicate that it represents a prefix
+typedef std::string resprefix_t;
```
jonesmz, line 65: Use the using keyword over typedef. Typedef sucks.

src/osp/Resource/Package.h
```diff
@@ -61,12 +61,26 @@ namespace osp
...
 
-class Package
-{
-
+// Stores a split path as <prefix, rest of path>
+typedef struct Path { resprefix_t prefix; std::string identifier; } path_t;
```
jonesmz, line 68: There's no reason to make this a typedef at all. Referencing it by the name of the struct type is perfectly fine.

jonesmz: Also, man, that's a lot of string copying. You sure? I feel like what you want here is actually two std::string_views and then a single string object that owns the storage (Careful with copy / move constructors though).

Or maybe even more accurately, you might want something like an std::string_range (doesn't exist, far as I know) that's basically just two size_t's that indicate the range of indicies in some std::string, applied as needed to get back an std::string_view.

src/osp/Resource/Package.h
jonesmz, line 65: I may be misremembering, but I thought the project's coding guidelines document encouraged capitalizing the first letter in the typedef name?

jonesmz (approved):

src/osp/Resource/Package.h
Capital-Asterisk: Very C-style indeed. There's many ways to go upon doing this, depending on how this struct will be used. Will this struct own a string or is it just a return type for decompose_path?

src/osp/Resource/Package.cpp
Capital-Asterisk: This can be turned into a more generic string_view parsing function: inputting a string view and returning a pair of string views. If Path is intended to own a string value, then the pair can be used to initialize a Path.

Capital-Asterisk (approved):

src/osp/Active/SysVehicle.cpp
```diff
@@ -282,7 +282,7 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
 
             // TODO: put package prefixes in resource path
             //       for now just get the first package
-            Package& package = m_scene.get_application().debug_get_packges()[0];
+            Package& package = m_scene.get_application().debug_get_package("lzdb");
```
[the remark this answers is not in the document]
z-adams: I'll change it to "find," it conveys the behavior more clearly

src/osp/OSPApplication.h
z-adams: Is this not the default behavior?

src/osp/Resource/Package.h
z-adams: The primary reason why I did that was so that `OSPApplication::m_packages` was a `std::map<resprefix_t, Package>` which makes it more obvious that the key is not just any old string but rather a specific engine construct. Sort of like how `size_t` (which is just `uint64_t`) conveys a specific meaning. I'll leave it up to you whether or not I should leave it this way.

src/osp/Resource/Package.cpp
z-adams: I can do that, although I'll probably end up making another `decompose_path` function that just wraps that one. The purpose of this one is to make it easy to parse resource names. I just realized that this is not used anywhere right now, but it will be once the rest of my branch is merged.

src/osp/OSPApplication.h
Capital-Asterisk: This "transparent comparator" allows string views to access the map. If I understand correctly, it will allow any type comparable to an std::string.

src/osp/Resource/Package.h
z-adams: Can't remember how it ended up this way, I suspect I originally typedef'd a `std::pair` or something and then forgot to get rid of it when I changed to a struct.

src/osp/OSPApplication.cpp
z-adams: We do indeed want move-only behavior, the Package class's copy constructor is already deleted (although the copy assign is missing, I'll delete that one too). I don't think it makes sense to allow the same key to be added multiple times so it should probably just fail. I'll assert for now; if @Capital-Asterisk thinks overwriting is better I can change it.

src/osp/Resource/Package.h
z-adams: Looks like string_view will work pretty much everywhere, I will either just change everything to string_view or make an equivalent typedef for the resprefix, depending on whether cap thinks it's reasonable to use the typedef in the first place.

jonesmz:
I'm fine with merging this, so whenever you're happy @Capital-Asterisk

Capital-Asterisk (approved):

PR #25, opened 2020-12-29
Add a string_buffer class to provide read-only string fancyness

jonesmz:


Capital-Asterisk:
I'm not quite sure how to properly use this class yet, but it looks very handy. Admittedly, this level of C++ is still kind of alien language to me, I can't really judge it that well.

The only issue I have with this is the name. "StringBuffer" is a well known Java class that does the exact opposite of this: a mutable string. "Buffer" doesn't really hint at the data being read-only. Perhaps "readonly_string," "fixed_string," "final_string," or something.

Capital-Asterisk (approved):

PR #26, opened 2020-12-30
Unique mesh identities

z-adams:
Mesh data in blender files are given autogenerated names like
"Cylinder.004" and are usually not touched by the user. Thus, two
separate glTF files can easily end up containing meshes that share the
same name. To solve this, we use the data filepath to generate a unique
name prefix for meshes, which is used by the various loading functions
to discriminate between mesh resources.

---

As yet these changes are untested; it compiles and the program runs, but I haven't tried to induce a name collision yet. I'll try it sometime before merging, but for now I'm throwing this up here so it can be reviewed.

### DIFF
--- a/src/osp/Resource/AssetImporter.cpp
+++ b/src/osp/Resource/AssetImporter.cpp
@@ -44,6 +44,7 @@
 
 #include "Package.h"
 #include "AssetImporter.h"
+#include "osp/string_concat.h"
 
 using Corrade::Containers::Optional;
 using Magnum::Trade::ImageData2D;
@@ -51,42 +52,46 @@ using Magnum::Trade::MeshData;
 using Magnum::Trade::SceneData;
 using Magnum::GL::Texture2D;
 using Magnum::GL::Mesh;
+using Magnum::UnsignedInt;
 
 namespace osp
 {
 
-void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package& pkg)
+void osp::AssetImporter::load_sturdy_file(std::string_view filepath, Package& pkg)
 {
     PluginManager pluginManager;
     TinyGltfImporter gltfImporter{pluginManager};
 
     // Open .sturdy.gltf file
-    gltfImporter.openFile(filepath);
+    gltfImporter.openFile(std::string{filepath});
     if (!gltfImporter.isOpened() || gltfImporter.defaultScene() == -1)
     {
         std::cout << "Error: couldn't open file: " << filepath << "\n";
     }
 
-    load_sturdy(gltfImporter, pkg);
+    // repurpose filepath into unique identifier for file resources
+    std::string dataResourceName =
+        string_concat(filepath.substr(0, filepath.find('.')), ":");
+
+    load_sturdy(gltfImporter, dataResourceName, pkg);
 
     gltfImporter.close();
 }
 
 void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
-    Package& pkg, unsigned id)
+    Package& pkg, UnsignedInt id, std::string_view resPrefix)
 {
     // It's a part
     std::cout << "PART!\n";
 
     // Recursively add child nodes to part
     PrototypePart part;
-    proto_add_obj_recurse(gltfImporter, pkg, part, 0, id);
+    proto_add_obj_recurse(gltfImporter, pkg, resPrefix, part, 0, id);
 
     // Parse extra properties
-    tinygltf::Node const& node = *static_cast<tinygltf::Node const*>(
-        gltfImporter.object3D(id)->importerState());
+    tinygltf::Value const& extras = static_cast<tinygltf::Node const*>(
+        gltfImporter.object3D(id)->importerState())->extras;
 
-    tinygltf::Value const& extras = node.extras;
     if (!extras.Has("machines"))
     {
         std::cout << "Error: no machines found in "
@@ -122,7 +127,22 @@ void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
     pkg.add<PrototypePart>(gltfImporter.object3DName(id), std::move(part));
 }
 
-void osp::AssetImporter::load_sturdy(TinyGltfImporter& gltfImporter, Package& pkg)
+/* Explanation of resPrefix:
+   When a mesh is created in blender, the object itself has a name (the one that
+   shows up in the scene hierarchy), but the underlying mesh data within that
+   object actually has a unique name, usually the name of the primitive that was
+   used initially, unless it was changed. These names will be something like
+   "Cylinder.004" and are numbered to prevent name collisions within a blend
+   file. The issue is that multiple blend files can have a "Cylinder.004" and
+   unless the author renames the mesh object itself (which most people never
+   touch), when you export the scenes to glTF files and load them both here, you
+   end up with a resource key collision due to the repeated name. Passing
+   resPrefix around allows us to specify a unique prefix to prepend to the mesh
+   name (or any other resource that has the same problem) that is used
+   internally to avoid name conflicts.
+*/
+void osp::AssetImporter::load_sturdy(TinyGltfImporter& gltfImporter,
+        std::string_view resPrefix, Package& pkg)
 {
     std::cout << "Found " << gltfImporter.object3DCount() << " nodes\n";
     Optional<SceneData> sceneData = gltfImporter.scene(gltfImporter.defaultScene());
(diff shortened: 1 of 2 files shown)


src/osp/Resource/AssetImporter.cpp
```diff
@@ -122,7 +125,22 @@ void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
...
+   used initially, unless it was changed. These names will be something like
+   "Cylinder.004" and are numbered to prevent name collisions within a blend
+   file. The issue is that multiple blend files can have a "Cylinder.004" and
+   unless the author renames the mesh object itself (which most people never
+   touch), when you export the scenes to glTF files and load them both here, you
+   end up with a resource key collision due to the repeated name. Passing
+   resPrefix around allows us to specify a unique prefix to prepend to the mesh
+   name (or any other resource that has the same problem) that is used
+   internally to avoid name conflicts.
+*/
```
z-adams, line 141: Not sure if this is the best place for this; I think it's worth having somewhere, and I put it near the implementation since this detail is mostly hidden from the user

src/osp/Resource/AssetImporter.cpp
```diff
@@ -151,7 +169,8 @@ void osp::AssetImporter::load_sturdy(TinyGltfImporter& gltfImporter, Package& pk
     {
         using Magnum::MeshPrimitive;
 
-        std::string const& meshName = gltfImporter.meshName(i);
+        std::string const& meshName = static_cast<std::string>(resPrefix)
+            + gltfImporter.meshName(i);
```
z-adams, line 173: This kinda sucks but I'm not sure how to avoid it, as the string needs to be modified. Open to suggestions.

jonesmz: Definately cannot be casting from std::string_view to std::string. Big no-no. Call the std::string constructor directly.

Regardless, this is the wrong approach.

Use string_concat.

Copy it from here: https://github.com/splinter-build/splinter/blob/15-default-values/src/string_concat.h

Or wait for me to open a PR with a simplified version tomorrow.

src/osp/Resource/AssetImporter.cpp
```diff
@@ -67,20 +67,23 @@ void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package&
         std::cout << "Error: couldn't open file: " << filepath << "\n";
     }
 
-    load_sturdy(gltfImporter, pkg);
+    // repurpose filepath into unique identifier for file resources
```
jonesmz, line 70: filePath looks an aweful lot like it should be an std::string_view.

src/osp/Resource/AssetImporter.cpp
```diff
@@ -67,20 +67,23 @@ void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package&
...
+    std::string dataResourceName = filepath.substr(0, filepath.find('.')) + ':';
```
jonesmz, line 71: This is expensive.

substr is an allocate + copy all by itself, and then you're appending data to it, which is another allocate and copy.

Probably this should use the string_concat function, and that substr call should be against a string_view to avoid paying for an unnecessary allocate and copy.

src/osp/Resource/AssetImporter.cpp
```diff
@@ -67,20 +67,23 @@ void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package&
...
+
+    load_sturdy(gltfImporter, dataResourceName, pkg);
 
     gltfImporter.close();
 }
 
 void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
-    Package& pkg, unsigned id)
+    Package& pkg, unsigned id, std::string_view resPrefix)
```
jonesmz, line 79: Please change the type of id from "unsigned" to, well, anything else. e.g. std::uint64_t or something.

src/osp/Resource/AssetImporter.cpp
```diff
@@ -67,20 +67,23 @@ void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package&
...
     // It's a part
     std::cout << "PART!\n";
 
     // Recursively add child nodes to part
     PrototypePart part;
-    proto_add_obj_recurse(gltfImporter, pkg, part, 0, id);
+    proto_add_obj_recurse(gltfImporter, pkg, resPrefix, part, 0, id);
 
     // Parse extra properties
     tinygltf::Node const& node = *static_cast<tinygltf::Node const*>(
```
jonesmz, line 89: i notice nothing actually ever uses "node" other than the line where you bind node.extras to a reference.

You can use structured bindings for that.

tinygltf::Value const& [extras] = *static_cast<...>(...);

jonesmz (changes requested):
Request that you look into string_concat from the link I provided, to optimally concatenate strings.

src/osp/Resource/AssetImporter.cpp
z-adams: In this case specifically, or in general? Cap used `unsigned` for a lot of things like this, I assume because it's clearer. If you use `uint64_t` or `uint32_t` it sort of falsely implies that there's a specific reason to use that size, when in reality it's just supposed to be a small positive integer and the size doesn't matter.

jonesmz: the size of unsigned changes based on the platform being compiled for. No one wants that.

Specific bit sizes result in the least amount of surprise.

jonesmz: > If you use uint64_t or uint32_t it sort of falsely implies that there's a specific reason to use that size, when in reality it's just supposed to be a small positive integer and the size doesn't matter.

That's solved by using a typedef, and a comment explaining that the choice was arbitrary.

z-adams: In this case the IDs are all very small (like, almost guaranteed to be <100 and usually <10) so changing sizes doesn't matter but a typedef will work too. I propose we make a new PR for this though, as this is just one of many instances of its use in the repo

src/osp/Resource/AssetImporter.cpp
z-adams: It probably should, but since I didn't introduce it in this PR, I think it would be better to make one giant PR where we go and change all the strings to string_views at once, rather than gradually changing them in unrelated PRs

src/osp/Resource/AssetImporter.cpp
z-adams: Can you briefly explain how structured bindings work here? Is `extras` an arbitrary name, or is it accessing the `node.extras` member? The node has a bunch of `Value`s in it, so I don't understand how the compiler would be able to figure out what we're trying to do in this case.

jonesmz: Actually, structured bindings may not work here.

Structured bindings work using the machinery for std::get<>, where the first thing in the structured binding is index 0, and the second is index 1, and so on.

Better would be to instead do

```
tinygltf::Value const& extras = static_cast<...>(...)->extras;
```
directly.

Sorry for the confusion.

src/osp/Resource/AssetImporter.cpp
jonesmz: Sure that's fine.

If they're always smaller than 256, then use std::uint8_t.

src/osp/Resource/AssetImporter.cpp
jonesmz: Sure that's fine.

src/osp/Resource/AssetImporter.cpp
z-adams: Alright makes sense, thanks for the clarification

z-adams:
I rebased my branch off of the new master but I think I forgot to push to update remote before adding my next commit, hopefully I didn't break anything by forcing the remote to accept my local version

src/osp/Resource/AssetImporter.cpp
```diff
@@ -55,38 +56,41 @@ using Magnum::GL::Mesh;
...
 
-void osp::AssetImporter::load_sturdy_file(std::string const& filepath, Package& pkg)
+void osp::AssetImporter::load_sturdy_file(std::string_view filepath, Package& pkg)
 {
     PluginManager pluginManager;
     TinyGltfImporter gltfImporter{pluginManager};
 
     // Open .sturdy.gltf file
-    gltfImporter.openFile(filepath);
+    gltfImporter.openFile(std::string{filepath});
```
jonesmz, line 65: gltifImporter.openFile doesn't support std::string_view?

Does it take an std::filesystem::path or something?

z-adams: I tried leaving it that way, but it said:
```
osp-magnum/src/osp/Resource/AssetImporter.cpp:65:35: error: no matching function for call to ‘Magnum::Trade::TinyGltfImporter::openFile(std::string_view&)’
     gltfImporter.openFile(filepath);
```

It takes a `const std::string&`. Should it work? If not, it's a small price to pay for avoiding the many allocations from `substr`ing the `std::string` as before.

jonesmz: How it is is the best you're going to be able to do unless we can convince the Magnum folks to support std::string_view in their APIs.

So this is fine.

jonesmz:
Looks much better!

I'm fine with this being merged.

Capital-Asterisk (approved):
If I understand correctly, this fixes how multiple gltf files with the same mesh names loaded into the same package will cause a name conflict.

I'm trying to think of things that can go wrong, but so far this is a good solution.

When first writing the sturdy loading, I was thinking that the same mesh can be shared across different gltf files if they had the same name. This is a pretty bad idea. If parts want to use the same meshes, then they belong in the same gltf file.

z-adams:
The force push seems to have broken the github pr comments/interface for some reason, but the commits themselves seem fine.

Yes, the trouble with sharing mesh names is that the actual mesh data names are not the same as the names you assign to objects in the glTF node tree. The mesh names are the autogenerated ones like "Cylinder.004" that blender assigns to the actual mesh data underneath the object, which the artist basically never sees in the blender interface. You can test out the functionality of this PR by changing the instances of
```cpp
std::string const& meshName = string_concat(resPrefix, gltfImporter.meshName(i));
```
to
```cpp
std::string const& meshName = gltfImporter.meshName(i);
```
and then renaming the SpamCan's mesh data to be the same as the Stomper's (`Cylinder.002`) and you'll see that the Stomper's mesh is replaced with the SpamCan's. The issue just never arose because we don't have many models, but in my main trailblazing branch it was one of the first issues I came up against and had to comment out all the other assets. I'll change the `unsigned`s to magnum's equivalent and then I'll merge these changes.

PR #27, opened 2020-12-30
Improve Planet code and Add real-world moon test universe

Capital-Asterisk:
This big PR makes many improvements and additions to Planet-A, as well as few minor changes with other related files.

First off, a quick rundown on what the planet classes do:
* **IcoSphereTree**: A subdividable icosahedron where each triangle face can be subdivided into 4 children. This also keeps track of which triangles are neighbouring each other. Nothing rendering-related happens here.
* **PlanetGeometryA**: Creates 'chunks' to patch overtop of an IcoSphereTree intended to render later. These chunks consist of many smaller triangles, and are stored in a vertex and index buffer. Nothing rendering-related happens here either.
* **SysPlanetA**: Manages the active planets in the scene. It is responsible for dealing with entities assigned with planet data (ACompPlanet). ACompPlanets contain PlanetGeometryAs, and it's up to SysPlanetA to interface its buffers to the GPU for rendering, as well as deal with physics and colliders.

PlanetGeometryA is ridiculously complex, and may need to be explained better in some document or presentation.  God bless whoever attempts to fully understand it at this point.

Notable additions from this PR:
* Vertex sharing between chunks
* Configurable Planet Level-of-Detail falloff based on distance. This is implemented as a lambda function in SysPlanet. This function is used by PlanetGeometryA to determine if triangles should be chunked, subdivided, etc... Configuration is in UCompPlanet.
* Large Seamless surfaces. On a boundary between two chunks of varying detail, the chunk with higher detail are flattened against the lower detail one. Enable the stupid heightmap in PlanetGeometryA line 215 and enter the moon test in flight scene to see this more clearly.
* Gravity attraction force calculated from specified mass of planet
* Test universe with real-world moon simulation. This creates a 'planet' with a radius of 1.74E+6m, and a mass of 7.35E+22kg.
* Improved readability in all planet code

Not done, but possible:
* Generating planet colliders realtime
* Normal calculations for chunks
* Proper way to handle height maps
* Maybe cut up PlanetGeometryA and IcoSphereTree even further into something less OOPy. The state of these two classes are rather difficult to track, which can be improved

### DIFF
--- a/src/osp/Universe.cpp
+++ b/src/osp/Universe.cpp
@@ -51,13 +51,13 @@ void Universe::sat_remove(Satellite sat)
 }
 
 
-Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target)
+Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target) const
 {
-    auto view = m_registry.view<UCompTransformTraj>();
+    auto const view = m_registry.view<const UCompTransformTraj>();
 
     // TODO: maybe do some checks to make sure they have the components
-    auto &framePosTraj = view.get<UCompTransformTraj>(referenceFrame);
-    auto &targetPosTraj = view.get<UCompTransformTraj>(target);
+    auto const &framePosTraj = view.get<const UCompTransformTraj>(referenceFrame);
+    auto const &targetPosTraj = view.get<const UCompTransformTraj>(target);
 
     if (framePosTraj.m_parent == targetPosTraj.m_parent)
     {
@@ -69,7 +69,7 @@ Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target)
     return {0, 0, 0};
 }
 
-Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target)
+Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target) const
 {
     // 1024 units = 1 meter. this can change
     return Vector3(sat_calc_pos(referenceFrame, target)) / 1024.0f;
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -206,6 +206,8 @@ void SysPlanetA::update_geometry()
             planet.m_vrtxBufGL.setData(planet.m_planet.get_vertex_buffer());
             planet.m_indxBufGL.setData(planet.m_planet.get_index_buffer());
 
+            planet.m_planet.get_index_buffer_updates().clear();
+
             using Magnum::Shaders::MeshVisualizer3D;
             using Magnum::GL::MeshPrimitive;
             using Magnum::GL::MeshIndexType;
@@ -224,16 +226,48 @@ void SysPlanetA::update_geometry()
 
         if (m_debugUpdate.triggered())
         {
-            planet.m_vrtxBufGL.setSubData(0, {0.0f, 0.0f, 0.0f});
 
             // subdivide all the triangles
-            planet.m_planet.chunk_geometry_update(
+            planet.m_planet.chunk_geometry_update_all(
                     [ent] (SubTriangle const& tri,
                            SubTriangleChunk const& chunk,
                            int index) -> EChunkUpdateAction
             {
-                return EChunkUpdateAction::Subdivide;
+                if (tri.m_depth == 2)
+                {
+                    if (chunk.m_chunk == gc_invalidChunk)
+                    {
+                        return EChunkUpdateAction::Chunk;
+                    }
+                    else
+                    {
+                        return EChunkUpdateAction::Unchunk;
+                    }
+                }
+                else
+                {
+                    return EChunkUpdateAction::Nothing;
+                }
             });
+
+            using Corrade::Containers::ArrayView;
+
+            std::vector<unsigned> const &indxData = planet.m_planet.get_index_buffer();
+
+            // update GPU buffers
+            for (UpdateRangeSub updRange : planet.m_planet.get_index_buffer_updates())
+            {
+                buindex_t size = updRange.m_end - updRange.m_start;
+                auto arrView = ArrayView(indxData.data() + updRange.m_start, size);
+                //planet.m_indxBufGL.setSubData(updRange.m_start, arrView);
+            }
+
+            planet.m_vrtxBufGL.setData(planet.m_planet.get_vertex_buffer());
+            planet.m_indxBufGL.setData(planet.m_planet.get_index_buffer());
+
+            planet.m_mesh.setCount(planet.m_planet.calc_index_count());
+
+            planet.m_planet.get_index_buffer_updates().clear();
         }
     }
 }
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -296,7 +296,7 @@ void SysPlanetA::update_geometry()
             planet.m_icoTree->subdivide_remove_all_unused();
             planet.m_icoTree->event_notify();
 
-            planet.m_planet->debug_raise_by_share_count();
+            //planet.m_planet->debug_raise_by_share_count();
 
             using Corrade::Containers::ArrayView;
 
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -127,7 +127,7 @@ void SysPlanetA::draw(osp::active::ACompCamera const& camera)
         auto& planet = drawGroup.get<ACompPlanet>(entity);
         auto& transform = drawGroup.get<ACompTransform>(entity);
 
-        if (!planet.m_planet.is_initialized())
+        if (planet.m_planet.get() == nullptr)
         {
             continue;
         }
@@ -168,7 +168,7 @@ void SysPlanetA::debug_create_chunk_collider(osp::active::ActiveEnt ent,
     fishTransform.m_transform = m_scene.reg_get<ACompTransform>(ent).m_transform;
 
     // Get triangle iterators to start and end triangles of the specified chunk
-    auto itsChunk = planet.m_planet.iterate_chunk(chunk);
+    auto itsChunk = planet.m_planet->iterate_chunk(chunk);
 
     // Send them to the physics engine
     physics.shape_create_tri_mesh_static(fishShape,
@@ -188,14 +188,24 @@ void SysPlanetA::update_geometry()
         auto &planet = view.get<ACompPlanet>(ent);
         auto &tf = view.get<ACompTransform>(ent);
 
-        if (!planet.m_planet.is_initialized())
+        if (planet.m_planet.get() == nullptr)
         {
+            planet.m_planet = std::make_shared<PlanetGeometryA>();
+            PlanetGeometryA &planetGeo = *(planet.m_planet);
+
             // initialize planet if not done so yet
             std::cout << "Initializing planet\n";
-            planet.m_planet.initialize(planet.m_radius, 4, 1u << 9, 1u << 13);
+            planet.m_icoTree = std::make_shared<IcoSphereTree>();
+
+            planet.m_icoTree->initialize(planet.m_radius);
+
+            planetGeo.initialize(planet.m_icoTree, planet.m_radius, 4,
+                                       1u << 9, 1u << 13);
+
+            planet.m_icoTree->event_add(planet.m_planet);
 
             // Chunk all the initial triangles
-            planet.m_planet.chunk_geometry_update_all(
+            planetGeo.chunk_geometry_update_all(
                     [] (SubTriangle const& tri, SubTriangleChunk const& chunk,
                         int index) -> EChunkUpdateAction
             {
@@ -205,7 +215,7 @@ void SysPlanetA::update_geometry()
             std::cout << "Planet initialized, now making colliders\n";
 
             // temporary: make colliders for all the chunks
-            for (chindex_t i = 0; i < planet.m_planet.chunk_count(); i ++)
+            for (chindex_t i = 0; i < planetGeo.chunk_count(); i ++)
             {
                 debug_create_chunk_collider(ent, planet, i);
                 std::cout << "* completed chunk collider: " << i << "\n";
@@ -213,10 +223,10 @@ void SysPlanetA::update_geometry()
 
             std::cout << "Planet colliders done\n";
 
-            planet.m_vrtxBufGL.setData(planet.m_planet.get_vertex_buffer());
-            planet.m_indxBufGL.setData(planet.m_planet.get_index_buffer());
+            planet.m_vrtxBufGL.setData(planetGeo.get_vertex_buffer());
+            planet.m_indxBufGL.setData(planetGeo.get_index_buffer());
 
-            planet.m_planet.get_index_buffer_updates().clear();
+            planetGeo.updates_clear();
 
             using Magnum::Shaders::MeshVisualizer3D;
             using Magnum::GL::MeshPrimitive;
@@ -229,13 +239,13 @@ void SysPlanetA::update_geometry()
                                  MeshVisualizer3D::Normal{})
                 .setIndexBuffer(planet.m_indxBufGL, 0,
                                 MeshIndexType::UnsignedInt)
-                .setCount(planet.m_planet.calc_index_count());
-
-
+                .setCount(planetGeo.calc_index_count());
         }
 
         if (m_debugUpdate.triggered() || true)
         {
+            PlanetGeometryA &planetGeo = *(planet.m_planet);
+
             ActiveEnt cam = m_scene.get_registry().view<ACompCamera>().front();
             auto &camTf = m_scene.reg_get<ACompTransform>(cam);
 
@@ -247,12 +257,12 @@ void SysPlanetA::update_geometry()
             using Magnum::Math::sqrt;
 
             // edge length of largest possible triangle
-            float edgeLengthA =  planet.m_planet.get_ico_tree()->get_radius()
+            float edgeLengthA =  planetGeo.get_ico_tree()->get_radius()
                                 / sqrt(10.0f + 2.0f * sqrt(5.0f)) * 4.0;
 
             float threshold = 1.0f;
 
-            planet.m_planet.chunk_geometry_update_all(
+            planetGeo.chunk_geometry_update_all(
                     [edgeLengthA, threshold, tf, camRelative, ent] (
                             SubTriangle const& tri,
                             SubTriangleChunk const& chunk,
@@ -279,17 +289,24 @@ void SysPlanetA::update_geometry()
                 }
             });
 
-            //planet.m_planet.debug_raise_by_share_count();
+            // New triangles were added, notify
+            planet.m_icoTree->event_notify();
+
+            // Remove unused
+            planet.m_icoTree->subdivide_remove_all_unused();
+            planet.m_icoTree->event_notify();
+
+            planet.m_planet->debug_raise_by_share_count();
 
             using Corrade::Containers::ArrayView;
 
             // update GPU index buffer
 
             std::vector<unsigned> const &indxData
-                    = planet.m_planet.get_index_buffer();
+                    = planetGeo.get_index_buffer();
 
             for (UpdateRangeSub updRange
-                 : planet.m_planet.get_index_buffer_updates())
+                 : planetGeo.updates_get_index_changes())
             {
                 buindex_t size = updRange.m_end - updRange.m_start;
                 ArrayView arrView(indxData.data() + updRange.m_start, size);
@@ -300,10 +317,10 @@ void SysPlanetA::update_geometry()
             // update GPU vertex buffer
 
             std::vector<float> const &vrtxData
-                    = planet.m_planet.get_vertex_buffer();
+                    = planetGeo.get_vertex_buffer();
 
             for (UpdateRangeSub updRange
-                 : planet.m_planet.get_vertex_buffer_updates())
+                 : planetGeo.updates_get_vertex_changes())
             {
                 buindex_t size = updRange.m_end - updRange.m_start;
                 ArrayView arrView(vrtxData.data() + updRange.m_start, size);
@@ -313,10 +330,11 @@ void SysPlanetA::update_geometry()
 
             //planet.m_vrtxBufGL.setData(planet.m_planet.get_vertex_buffer());
             //planet.m_indxBufGL.setData(planet.m_planet.get_index_buffer());
-            planet.m_planet.get_index_buffer_updates().clear();
-            planet.m_planet.get_vertex_buffer_updates().clear();
+            planetGeo.updates_clear();
 
-            planet.m_mesh.setCount(planet.m_planet.calc_index_count());
+            planet.m_mesh.setCount(planetGeo.calc_index_count());
+
+            planetGeo.get_ico_tree()->debug_verify_state();
         }
     }
 }
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -181,11 +181,12 @@ void SysPlanetA::debug_create_chunk_collider(osp::active::ActiveEnt ent,
 void SysPlanetA::update_geometry()
 {
 
-    auto view = m_scene.get_registry().view<ACompPlanet>();
+    auto view = m_scene.get_registry().view<ACompPlanet, ACompTransform>();
 
     for (osp::active::ActiveEnt ent : view)
     {
         auto &planet = view.get<ACompPlanet>(ent);
+        auto &tf = view.get<ACompTransform>(ent);
 
         if (!planet.m_planet.is_initialized())
         {
@@ -226,14 +227,48 @@ void SysPlanetA::update_geometry()
 
         if (m_debugUpdate.triggered())
         {
+            ActiveEnt cam = m_scene.get_registry().view<ACompCamera>().front();
+            auto &camTf = m_scene.reg_get<ACompTransform>(cam);
 
-            // subdivide all the triangles
+            // camera position relative to planet
+            Vector3 camRelative = camTf.m_transform.translation()
+                                - tf.m_transform.translation();
+
+            using Magnum::Math::pow;
+            using Magnum::Math::sqrt;
+
+            // edge length of largest possible triangle
+            float edgeLengthA =  planet.m_planet.get_ico_tree()->get_radius()
+                                / sqrt(10.0f + 2.0f * sqrt(5.0f)) * 4.0;
+
+            float threshold = 1.0f;
+
             planet.m_planet.chunk_geometry_update_all(
-                    [ent] (SubTriangle const& tri,
-                           SubTriangleChunk const& chunk,
-                           int index) -> EChunkUpdateAction
+                    [edgeLengthA, threshold, tf, camRelative, ent] (
+                            SubTriangle const& tri,
+                            SubTriangleChunk const& chunk,
+                            int index) -> EChunkUpdateAction
             {
-                if (tri.m_depth == 2)
+                // approximation of triangle edge length
+                float edgeLength = edgeLengthA / float(pow(2, int(tri.m_depth)));
+
+                // distance between viewer and triangle
+                float dist = (tri.m_center - camRelative).length();
+
+                float screenLength = edgeLength / dist;
+
+                bool tooFar = screenLength < threshold;
+
+                if (tooFar)
+                {
+                    return EChunkUpdateAction::Chunk;
+                }
+                else
+                {
+                    return EChunkUpdateAction::Subdivide;
+                }
+
+                /*if (tri.m_depth == 2)
                 {
                     if (chunk.m_chunk == gc_invalidChunk)
                     {
@@ -247,7 +282,7 @@ void SysPlanetA::update_geometry()
                 else
                 {
                     return EChunkUpdateAction::Nothing;
-                }
+                }*/
             });
 
             using Corrade::Containers::ArrayView;
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -91,14 +91,16 @@ StatusActivated SysPlanetA::activate_sat(
     return {0, planetEnt, false};
 }
 
-SysPlanetA::SysPlanetA(osp::active::ActiveScene &scene) :
+SysPlanetA::SysPlanetA(osp::active::ActiveScene &scene,
+                       osp::UserInputHandler &userInput) :
     m_scene(scene),
     m_updateGeometry(scene.get_update_order(), "planet_geo", "", "physics",
-                     std::bind(&SysPlanetA::update_geometry, this)),
+                    [this] { this->update_geometry(); } ),
     m_updatePhysics(scene.get_update_order(), "planet_phys", "planet_geo", "",
-                    std::bind(&SysPlanetA::update_geometry, this)),
+                    [this] { this->update_physics(); }),
     m_renderPlanetDraw(scene.get_render_order(), "", "", "",
-                       std::bind(&SysPlanetA::draw, this, _1))
+                       std::bind(&SysPlanetA::draw, this, _1)),
+    m_debugUpdate(userInput.config_get("debug_planet_update"))
 {
 
 }
@@ -124,6 +126,11 @@ void SysPlanetA::draw(osp::active::ACompCamera const& camera)
     {
         auto& planet = drawGroup.get<ACompPlanet>(entity);
         auto& transform = drawGroup.get<ACompTransform>(entity);
+
+        if (!planet.m_planet.is_initialized())
+        {
+            continue;
+        }
 
         entRelative = camera.m_inverse * transform.m_transformWorld;
 
@@ -211,6 +218,22 @@ void SysPlanetA::update_geometry()
                 .setIndexBuffer(planet.m_indxBufGL, 0,
                                 MeshIndexType::UnsignedInt)
                 .setCount(planet.m_planet.calc_index_count());
+
+
+        }
+
+        if (m_debugUpdate.triggered())
+        {
+            planet.m_vrtxBufGL.setSubData(0, {0.0f, 0.0f, 0.0f});
+
+            // subdivide all the triangles
+            planet.m_planet.chunk_geometry_update(
+                    [ent] (SubTriangle const& tri,
+                           SubTriangleChunk const& chunk,
+                           int index) -> EChunkUpdateAction
+            {
+                return EChunkUpdateAction::Subdivide;
+            });
         }
     }
 }
--- a/src/planet-a/Active/SysPlanetA.cpp
+++ b/src/planet-a/Active/SysPlanetA.cpp
@@ -53,7 +53,6 @@ using osp::universe::Satellite;
 
 const std::string SysPlanetA::smc_name = "PlanetA";
 
-
 StatusActivated SysPlanetA::activate_sat(
         osp::active::ActiveScene &scene,
         osp::active::SysAreaAssociate &area,
@@ -276,7 +275,7 @@ void SysPlanetA::planet_update_geometry(osp::active::ActiveEnt planetEnt)
 
     // Set this somewhere else. Radius at which detail falloff will start
     // This will essentially be the physics area size
-    float viewerRadius = 128.0f;
+    float viewerRadius = 64.0f;
 
     // camera position relative to planet
     Vector3 camRelative = camTf.m_transform.translation()
@@ -290,7 +289,8 @@ void SysPlanetA::planet_update_geometry(osp::active::ActiveEnt planetEnt)
     static const float sc_icoEdgeRatio = sqrt(10.0f + 2.0f * sqrt(5.0f)) / 4.0f;
 
     // edge length of largest possible triangle
-    float edgeLengthA = planetUComp.m_radius / sc_icoEdgeRatio / rPlanetGeo.get_chunk_vertex_width();
+    float edgeLengthA = planetUComp.m_radius / sc_icoEdgeRatio
+                        / rPlanetGeo.get_chunk_vertex_width();
 
     rPlanetGeo.chunk_geometry_update_all(
             [edgeLengthA, planetUComp, camRelative, viewerRadius] (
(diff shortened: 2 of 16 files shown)


z-adams:
Mother of god... this is basically equivalent to my entire trailblazing branch haha. You'll probably feel similarly when I try to merge my dynamic moment of inertia code. In lieu of trying to understand all 2k lines of changes, is there anything I should try when I test it on my machine?

jonesmz:
I say this without digging into the code itself,

but this many changes seem like they should be broken up into 2-3 PRs.

src/osp/Universe.cpp
```diff
@@ -51,13 +51,13 @@ void Universe::sat_remove(Satellite sat)
 }
 
 
-Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target)
+Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target) const
 {
-    auto view = m_registry.view<UCompTransformTraj>();
+    auto const view = m_registry.view<const UCompTransformTraj>();
```
jonesmz, line 56: Note that .view<const ...> probably has no meaningful effect, unless you're asking it to return a const&, which doesn't seem to be the case here.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -75,30 +74,34 @@ StatusActivated SysPlanetA::activate_sat(
     ActiveEnt root = scene.hier_get_root();
     ActiveEnt planetEnt = scene.hier_create_child(root, "Planet");
 
-    auto &planetTransform = scene.get_registry()
-                            .emplace<osp::active::ACompTransform>(planetEnt);
-    planetTransform.m_transform = Matrix4::translation(positionInScene);
+    auto &rPlanetTransform = scene.get_registry()
+                             .emplace<osp::active::ACompTransform>(planetEnt);
```
jonesmz, line 78: Are you following an 80 column rule or something?

I'd argue that if you are, 120 columns might be better. Basically no one has screens that small anymore, and if they do, they're on their own.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -75,30 +74,34 @@ StatusActivated SysPlanetA::activate_sat(
...
-    auto &planetPlanet = scene.reg_emplace<ACompPlanet>(planetEnt);
-    planetPlanet.m_radius = loadMePlanet.m_radius;
+    auto &rPlanetPlanet = scene.reg_emplace<ACompPlanet>(planetEnt);
+    rPlanetPlanet.m_radius = loadMePlanet.m_radius;
+
+    auto &rPlanetForceField = scene.reg_emplace<ACompFFGravity>(planetEnt);
 
-    auto &planetForceField = scene.reg_emplace<ACompFFGravity>(planetEnt);
+    // gravitational constant
+    static const float sc_GravConst = 6.67408E-11f;
```
jonesmz, line 88: Not 9.8?

Should this go into a globals.h file somewhere? You can also make it a template variable, like such.

template<typename T>
inline constexpr T sgc_gravityConstant = T(6.67408E-11);

And then provide more or less accuracy for specializations.

template<>
inline constexpr float sgc_gravityConstant<float> = 6.67408E-11f;

template<>
inline constexpr double sgc_gravityConstant<double> = // some other number here

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -75,30 +74,34 @@ StatusActivated SysPlanetA::activate_sat(
...
     return {0, planetEnt, false};
 }
 
-SysPlanetA::SysPlanetA(osp::active::ActiveScene &scene) :
+SysPlanetA::SysPlanetA(osp::active::ActiveScene &scene,
+                       osp::UserInputHandler &userInput) :
     m_scene(scene),
     m_updateGeometry(scene.get_update_order(), "planet_geo", "", "physics",
-                     std::bind(&SysPlanetA::update_geometry, this)),
+                    [this] { this->update_geometry(); } ),
```
jonesmz, line 99: Do we need to care about lifetime management of this?

Are we guaranteed that whatever m_updateGeometry talks to will always be dead before "this" is dead?

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -75,30 +74,34 @@ StatusActivated SysPlanetA::activate_sat(
...
     m_updatePhysics(scene.get_update_order(), "planet_phys", "planet_geo", "",
-                    std::bind(&SysPlanetA::update_geometry, this)),
+                    [this] { this->update_physics(); }),
     m_renderPlanetDraw(scene.get_render_order(), "", "", "",
-                       std::bind(&SysPlanetA::draw, this, _1))
+                       std::bind(&SysPlanetA::draw, this, _1)),
```
jonesmz, line 103: [this] (auto && arg) { this->draw(std::forward<decltype(arg)>(arg)); }

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -110,11 +113,15 @@ int SysPlanetA::deactivate_sat(osp::active::ActiveScene &scene,
                                osp::universe::Satellite tgtSat,
                                osp::active::ActiveEnt tgtEnt)
 {
+    // TODO
```
jonesmz, line 116: provide a one sentence summary of what the function should do when it's written.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -125,6 +132,11 @@ void SysPlanetA::draw(osp::active::ACompCamera const& camera)
         auto& planet = drawGroup.get<ACompPlanet>(entity);
         auto& transform = drawGroup.get<ACompTransform>(entity);
 
+        if (planet.m_planet.get() == nullptr)
```
jonesmz, line 135: it's sufficient to just do "planet.m_planet == nullptr", you don't need to call .get().

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -174,30 +186,53 @@ void SysPlanetA::debug_create_chunk_collider(osp::active::ActiveEnt ent,
...
+            PlanetGeometryA &rPlanetGeo = *(planet.m_planet);
+
             // initialize planet if not done so yet
             std::cout << "Initializing planet\n";
-            planet.m_planet.initialize(planet.m_radius);
+            planet.m_icoTree = std::make_shared<IcoSphereTree>();
+
+            planet.m_icoTree->initialize(planet.m_radius);
+
+            rPlanetGeo.initialize(planet.m_icoTree, 4, 1u << 9, 1u << 13);
```
jonesmz, line 207: magic numbers.

Perhaps make named constants instead.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -174,30 +186,53 @@ void SysPlanetA::debug_create_chunk_collider(osp::active::ActiveEnt ent,
...
+
+            planet.m_icoTree->event_add(planet.m_planet);
```
jonesmz, line 209: Does it make sense to provide a constructor for IcoSphereTree that can accept the planet to add, so that it can be initialized in one go?

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -174,30 +186,53 @@ void SysPlanetA::debug_create_chunk_collider(osp::active::ActiveEnt ent,
...
+
+            // Chunk all the initial triangles
+            rPlanetGeo.chunk_geometry_update_all(
+                    [] (SubTriangle const& tri, SubTriangleChunk const& chunk,
```
jonesmz, line 213: use [[maybe_unused]] to suppress warnings.

jonesmz: or just replace the argument list with "..." or something.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -210,9 +245,123 @@ void SysPlanetA::update_geometry()
                                  MeshVisualizer3D::Normal{})
                 .setIndexBuffer(planet.m_indxBufGL, 0,
                                 MeshIndexType::UnsignedInt)
-                .setCount(planet.m_planet.calc_index_count());
+                .setCount(rPlanetGeo.calc_index_count());
+        }
+
+        if (m_debugUpdate.triggered() || true)
```
jonesmz, line 251: So... always?

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -210,9 +245,123 @@ void SysPlanetA::update_geometry()
...
+
+    PlanetGeometryA &rPlanetGeo = *(rPlanetPlanet.m_planet);
+
+    // Temporary: use the first camera in the scene as the viewer
+    ActiveEnt cam = m_scene.get_registry().view<ACompCamera>().front();
+    auto const &camTf = m_scene.reg_get<ACompTransform>(cam);
+
+    // Set this somewhere else. Radius at which detail falloff will start
+    // This will essentially be the physics area size
+    float viewerRadius = 64.0f;
```
jonesmz, line 278: looks like a good candidate for global.h for the initial value.

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -210,9 +245,123 @@ void SysPlanetA::update_geometry()
...
+    // camera position relative to planet
+    Vector3 camRelative = camTf.m_transform.translation()
+                        - planetTf.m_transform.translation();
+
+    using Magnum::Math::max;
+    using Magnum::Math::pow;
+    using Magnum::Math::sqrt;
+
+    // Ratio between an icosahedron's edge length and radius
+    static const float sc_icoEdgeRatio = sqrt(10.0f + 2.0f * sqrt(5.0f)) / 4.0f;
```
jonesmz, line 289: This should be constexpr.

Is Magnum::Math::sqrt not constexpr?

If not, then I think there's sufficient justification for us to define a constexpr_sqrt function that uses doubles for the calculation, with a naive implementation (It happens at compile time, so who cares if it's fast, I guess) to use when all the constants are known at compile time.

src/planet-a/IcoSphereTree.cpp
```diff
@@ -159,10 +146,10 @@ void IcoSphereTree::initialize(float radius)
     for (int i = 0; i < gc_icosahedronFaceCount; i ++)
     {
         // Set triangles
-        SubTriangle tri;
+        SubTriangle &rTri = m_triangles.emplace_back();
 
         // indices were already calculated beforehand
-        set_verts(tri,
+        set_verts(rTri,
```
jonesmz, line 152: My first impression is that this should instead be a constructor parameter.

My second impression is that it should be a member function.

Thoughts?

src/planet-a/IcoSphereTree.cpp
```diff
@@ -501,13 +557,153 @@ TriangleSideTransform IcoSphereTree::transform_to_ancestor(
...
+}
+
+void IcoSphereTree::event_notify()
+{
+    if (!m_trianglesAdded.empty())
+    {
+        for (std::weak_ptr<IcoSphereTreeObserver> ob : m_observers)
+        {
+            // TODO: get rid of expired ones
+            ob.lock()->on_ico_triangles_added(m_trianglesAdded);
```
jonesmz, line 587: Bad assumption.

Entirely possible that the weak_ptr points to a pointer that died.

If you know for sure that the weak_ptr always references a valid shared_ptr, then assert that the lock succeeds. E.g.

auto const& sp = ob.lock();
assert(sp != nullptr);
sp->on_ico_triangles_added();

src/planet-a/IcoSphereTree.cpp
```diff
@@ -501,13 +557,153 @@ TriangleSideTransform IcoSphereTree::transform_to_ancestor(
...
+            // TODO: get rid of expired ones
+            ob.lock()->on_ico_triangles_removed(m_trianglesRemoved);
         }
+        m_trianglesRemoved.clear();
+    }
 
-        curT = tri->m_parent;
+    if (!m_vrtxRemoved.empty())
+    {
+        for (std::weak_ptr<IcoSphereTreeObserver> ob : m_observers)
```
jonesmz, line 604: Iterating over the list 3 times is expensive.
Locking the shared_ptr 3 times is expensive.

Better to loop, and lock, once, and then do your if checks inside the loop and lock.

src/planet-a/IcoSphereTree.cpp
```diff
@@ -501,13 +557,153 @@ TriangleSideTransform IcoSphereTree::transform_to_ancestor(
...
+
+        SubTriangle &tri = get_triangle(t);
+
+        // verify hierarchy, for non-root triangles.
+        // the first 20 triangles are the root ones
+        if (t >= gc_icosahedronFaceCount)
+        {
+            SubTriangle &parent = get_triangle(tri.m_parent);
 
+            if (std::find(m_trianglesFree.begin(), m_trianglesFree.end(),
```
jonesmz, line 646: you seem to want std::any, not std::find, for this.

src/planet-a/IcoSphereTree.h
```diff
@@ -33,14 +38,23 @@ namespace planeta
 
 // Index to a triangle
 using trindex_t = uint32_t;
+constexpr trindex_t gc_invalidTri = std::numeric_limits<trindex_t>::max();
```
jonesmz, line 41: I think that constexpr automatically implies the inline keyword, but you might want to double check that you don't want the inline keyword as well.

src/planet-a/IcoSphereTree.h
```diff
@@ -33,14 +38,23 @@ namespace planeta
...
 
 class IcoSphereTree;
 struct SubTriangle;
 struct TriangleSideTransform;
 
+
+
 // The 20 faces of the icosahedron (Top, Left, Right)
 // Each number pointing to a vertex
 static constexpr uint8_t sc_icoTemplateTris[20 * 3] {
```
jonesmz, line 60: Prob these should be inline constexpr instead of static constexpr.

src/planet-a/IcoSphereTree.h
```diff
@@ -67,8 +81,70 @@ static constexpr int gc_icosahedronVertCount = 12;
...
+struct TriangleSideTransform
+{
+    float m_translation{0.0f};
+    float m_scale{1.0f};
+};
+
+class IcoSphereTreeObserver
+{
+public:
+    using Handle_T = std::list<std::weak_ptr<IcoSphereTreeObserver>>::iterator;
```
jonesmz, line 96: An iterator to a  weak_ptr?

Why not just std::shared_ptr, or std::weak_ptr, as a handle? Is knowing the specific position in the list important for a handle? How are you dealing with iterator invalidation?

src/planet-a/IcoSphereTree.h
```diff
@@ -67,8 +81,70 @@ static constexpr int gc_icosahedronVertCount = 12;
...
+
+    virtual void on_ico_triangles_added(std::vector<trindex_t> const&) = 0;
```
jonesmz, line 98: This looks like a good spot for std::span, but sadly we can't use that yet, since C++20 isn't really available anywhere.

src/planet-a/IcoSphereTree.h
```diff
@@ -67,8 +81,70 @@ static constexpr int gc_icosahedronVertCount = 12;
...
+};
+
+
+/**
+ * A triangle on the IcoSphereTree
+ */
+struct SubTriangle
+{
+    trindex_t m_parent;
+    uint8_t m_siblingIndex; // 0:Top 1:Left 2:Right 3:Center
```
jonesmz, line 110: Looks like a good place to use an enum.

enum class Sibling : std::uint8_t
{
    Top = 0,
    Left = 1,
    Right = 2,
    Center = 3
};

src/planet-a/IcoSphereTree.h
```diff
@@ -67,8 +81,70 @@ static constexpr int gc_icosahedronVertCount = 12;
...
+
+    trindex_t m_neighbours[3];
+    buindex_t m_corners[3]; // to vertex buffer, 3 corners of triangle
+
+    osp::Vector3 m_center;
+
+    bool m_subdivided{false}, m_deleted{false};
```
jonesmz, line 117: Personally I'm not a fan of this syntax for declaring variables.

If you like it though, go ahead

src/planet-a/IcoSphereTree.h
```diff
@@ -67,8 +81,70 @@ static constexpr int gc_icosahedronVertCount = 12;
...
+
+    // 0 for first 20 triangles of icosahedron. +1 for each depth onwards
+    uint8_t m_depth{0};
```
jonesmz, line 120: I think you might want to re-arrange your member variables to get better struct packing.

src/planet-a/IcoSphereTree.h
```diff
@@ -90,33 +166,36 @@ class IcoSphereTree
...
+     * @param triInd [in] Index to triangle
      * @return Pointer to triangle
      */
-    SubTriangle& get_triangle(trindex_t t)
-    {
-        return m_triangles[t];
-    }
-
-    trindex_t triangle_count()
+    SubTriangle& get_triangle(trindex_t triInd)
```
jonesmz, line 177: technically this is constexpr as well

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -38,21 +35,35 @@
...
+using Magnum::Math::abs;
+
+void planeta::update_range_insert(std::vector<UpdateRangeSub>& range,
+                                  UpdateRangeSub insert)
+{
+    // TODO: Merge update ranges to decrease number of glBufferSubData calls
+
+    //UpdateRangeSub merged;
+
+    range.push_back(insert);
```
jonesmz, line 48: UpdateRangeSub is a small object? No need for std::move() ?

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -38,21 +35,35 @@
...
-    m_vrtxSharedMax = 10000;
-    m_chunkMax = 100;
+    m_vrtxSharedMax = maxShared;
+    m_chunkMax = maxChunks;
 
     //m_chunkAreaThreshold = 0.04f;
-    m_chunkWidth = 17; // MUST BE (POWER OF 2) + 1. min: 5
-    m_chunkWidthB = m_chunkWidth - 1; // used often in many calculations
+    m_chunkWidthB = pow(2u, chunkDiv);
+    m_chunkWidth  = m_chunkWidthB + 1; // MUST BE (POWER OF 2) + 1. min: 5
```
jonesmz, line 61: Then add an assert to guarantee it.

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -66,14 +77,16 @@ void PlanetGeometryA::initialize(float radius)
...
-            - m_vrtxSharedPerChunk);
+                                                - m_vrtxSharedPerChunk);
 
     // allocate vectors for dealing with chunks
     m_indxBuffer.resize(m_chunkMax * m_indxPerChunk * 3);
     m_vrtxBuffer.resize(m_vrtxMax * m_vrtxSize);
     m_chunkToTri.resize(m_chunkMax);
     m_vrtxSharedUsers.resize(m_vrtxSharedMax);
-
+    m_vrtxSharedIcoCorners.resize(m_icoTree->get_vertex_buffer().size()
```
jonesmz, line 87: This is allocating the space, and then also initializing. Is that your intention?

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -239,27 +185,18 @@ loindex_t PlanetGeometryA::get_index_ringed(unsigned x, unsigned y) const
         // Bottom edge
         return x;
     }
-    else if (x == 0)
```
jonesmz, line 242: you might consider shortening this function like such:

```
if(y == m_chunkWidthB) return x; // Bottom edge
if(x == 0)                          return m_chunkWidthB * 2 + y; // Left Edge
if(x == y).....................
```

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -268,80 +205,161 @@ struct VertexToSubdiv
...
-    SubTriangle& tri = m_icoTree->get_triangle(t);
-
-    trindex_t min = std::max<trindex_t>(m_triangleChunks.size(),
-                                        m_icoTree->triangle_count());
-    m_triangleChunks.resize(min, {gc_invalidChunk, 0, gc_invalidTri, 0, 0});
+    float raise = 0; // add together a bunch of sin waves for terrain
+    raise += (std::sin(pos.x() / 128.0f) + std::sin(pos.y() / 128.0f) + std::sin(pos.z() / 128.0f)) * 64.0f;
+    raise += (std::sin(pos.x() / 500.0f) + std::sin(pos.y() / 500.0f) + std::sin(pos.z() / 500.0f)) * 128.0f;
+    raise += (std::sin(pos.x() / 720.0f) + std::sin(pos.y() / 720.0f) + std::sin(pos.z() / 720.0f)) * 512.0f;
+    return raise * 0.0f; // remove 0.0f for fun on the moon
```
jonesmz, line 215: So always return 0?

src/planet-a/Active/SysPlanetA.cpp
z-adams: With a split window, my laptop only manages 83 columns wide :'(

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -268,80 +205,161 @@ struct VertexToSubdiv
...
+        // Take the space at the end of the chunk buffer
+        rChunk.m_chunk = m_chunkCount;
+        m_chunkCount ++;
+    }
+    else
+    {
+        // Use empty space available. Get lowest value [last element] to make
+        // it less likely that the data will end up getting moved by chunk_pack
+        // when it's closer to the start of the buffer
+        std::set<chindex_t>::iterator chunkToGet = --m_chunkFree.end();
```
jonesmz, line 253: use std::advance(m_chunkFree.end(), -1);

src/planet-a/Active/SysPlanetA.cpp
z-adams: It's newton's gravity constant 6.67*10^-11, but yes it should go into a global constants file somewhere

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -268,80 +205,161 @@ struct VertexToSubdiv
...
+        rChunk.m_chunk = *chunkToGet;
+        m_chunkFree.erase(chunkToGet);
```
jonesmz, line 255: std::set is an unordered container.

What relevance does "last item in iterator order" have for what you're doing here?

If you're trying to do some kind of memory pool, then there are libraries out there that do a great job at that.

src/planet-a/Active/SysPlanetA.cpp
z-adams: It seems like it's specific to the planet rendering stuff and may be changed via graphics settings so I think keeping it in this file is appropriate, or perhaps moving it to the header

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -268,80 +205,161 @@ struct VertexToSubdiv
...
+        else if (neighChunk.m_descendentChunked != 0)
+        {
+            // Neighbour has chunked descendent, but has no chunk itself
+            rNeighbourChunked = neighbourIndex;
+            if (rTri.m_depth == neighTri.m_depth)
+            {
+                // if same depth, then set those descendent's chunked
+                // neighbours to t
+
+                uint8_t neighbourSide = rTri.m_neighbourSide[side];
```
jonesmz, line 313: why bother with declaring a variable?

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -352,70 +370,99 @@ void PlanetGeometryA::chunk_add(trindex_t t)
...
+        {
+            vertIndex = shared_create();
+
+            rSharedCorner = vertIndex;
+            m_vrtxSharedIcoCornersReverse[vertIndex] = triCorner;
+        }
+
+
+        buindex_t vrtxOffset = vertIndex * m_vrtxSize;
+        auto vrtxOffsetIt = m_vrtxBuffer.begin() + vrtxOffset;
```
jonesmz, line 420: use std::next for this.

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -352,70 +370,99 @@ void PlanetGeometryA::chunk_add(trindex_t t)
...
+                    m_icoTree->get_vertex_nrm(icoCornerVrtx));
+
+        Vector3 pos = vrtxPos;
+        Vector3 nrm = pos.normalized();
+
+        // temporary
+        pos = nrm * float(m_icoTree->get_radius());
+        pos += nrm * debug_stupid_heightmap(pos);
+
+        std::copy(pos.data(), pos.data() + pos.Size,
```
jonesmz, line 444: not pos.begin(), pos.end() ?

src/planet-a/IcoSphereTree.h
z-adams: cppreference says it's available on all of GCC libstdc++, Clang libc++, MSVC, and Apple Clang. Not sufficient?

src/planet-a/PlanetGeometryA.cpp
```diff
@@ -584,138 +644,367 @@ void PlanetGeometryA::chunk_add(trindex_t t)
...
+        //SubTriangle &moveTri = m_icoTree->get_triangle(moveT);
+        SubTriangleChunk &moveChunk = m_triangleChunks[moveT];
+
+        // Do move
+
+        // Replace tri's domain location with lastTriangle
+        m_chunkToTri[chunk] = m_chunkToTri[moveChunk.m_chunk];
+
+        // Move lastTri's index data to replace tri's data
+        std::copy(m_indxBuffer.begin() + moveChunk.m_dataIndx,
```
jonesmz, line 991: std::next

src/planet-a/PlanetGeometryA.h
```diff
@@ -42,16 +40,15 @@ using chindex_t = uint32_t;
...
-using vrindex_t = uint32_t;
 
 constexpr chindex_t gc_invalidChunk = std::numeric_limits<chindex_t>::max();
 constexpr loindex_t gc_invalidLocal = std::numeric_limits<loindex_t>::max();
-constexpr trindex_t gc_invalidTri = std::numeric_limits<trindex_t>::max();
-constexpr vrindex_t gc_invalidVrtx = std::numeric_limits<vrindex_t>::max();
 
 
-struct UpdateRange;
+enum class EChunkUpdateAction { Nothing, Subdivide, Unsubdivide,
```
jonesmz, line 48: provide an underlying type, give each value a value.

src/planet-a/PlanetGeometryA.h
```diff
@@ -355,10 +517,116 @@ class PlanetGeometryA::IteratorTriIndexed :
...
+    chunk_triangle_assure();
+
+    // loop through triangles, see which ones to chunk, subdivide, etc...
+    std::vector<trindex_t> toChunk;
+
+    // subdivision/unsubdivision can be done right away
+    // un-chunking can be done right away
+    // chunking should all be done at the end
+
+    // Loop through initial 20 triangles of icosahedron
```
jonesmz, line 533: Well, until the value of gc_icosahedronFaceCount changes.

src/planet-a/Active/SysPlanetA.cpp
jonesmz: O_o?

My laptop can do like 150 columns.

Where's the mismatch?

src/planet-a/IcoSphereTree.h
jonesmz: Is it available in Ubuntu 20.04?

Additionally, turning on C++20 also implies that people can use other C++20 features.

I think it would be better to wait a few more months until most people's dev machines have C++20 available. Even on my gentoo computers, I can't use C++20 yet. Gentoo's still on GCC 9 because GCC changed a default and fucked a bunch of packages up.

z-adams:
Haven't dug down into the detailed icosahedron/tree nitty gritty, but I briefly checked out the GPU data stuff and looked at it with RenderDoc and it looks fine. It might be worth drawing the chunks as a sequence of triangle strips later when we have high detail terrain (it may provide a large performance boost for slower machines), but I'd have to dive much deeper into the code to see what it would take to integrate with your scheme here. Indexed arrays work fine for now.

jonesmz:
I've provided as much of a review as I can, since the actual mathy-math is beyond my current understanding and I don't feel like learning new things today.

If you want to break this up into smaller PRs and ask for more targeted feedback, I'm happy to do that.

If you don't want to break it up into smaller PRs, then I'm not going to object to it being merged as is, (though I do hope you consider the feedback I did leave).

src/osp/Universe.cpp
Capital-Asterisk: Views provide access to a registry, and are fairly cheap to construct. They're still internally linked to the registry they were created from. Marking them const assures that they won't modify the registry.

src/planet-a/Active/SysPlanetA.cpp
z-adams: It's a small (14") 1600x900 display, and Ubuntu cleverly adjusts the text scaling to make the font very readable. I could shrink the scale but it would make it a lot harder to read.

Wait, 150? On both sides of a split screen? My full size 1920x1080 desktop monitor only gives me 103, 117 if I use the default zoom (although VS does have some clutter around the edges of the window). You say your eyesight sucks but I think your code is just half of a comfortably readable size haha

src/planet-a/IcoSphereTree.h
z-adams: I'm not sure, I've only tested C++20 features on MSVC. We can wait though if the latest gcc isn't available everywhere yet.

z-adams (approved):

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: Yes, I am following an 80 column rule. The verbosity of modern C++ code makes this a bit difficult to follow. I usually code with a window vertically split on half of a 1080p screen, and sometimes use a second monitor. Personally, I find it easier to read text that extends more vertically than horizontally.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: This is a good idea. I kept it self-contained to stop this PR from getting any larger. The file name should imply something scientific, like "sciencitifc_globals.h," "si_globals.h," or part of the subdirectory name.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: The dynamic systems added to the scene stay with it throughout its entire lifetime. The scene does not call any updates during destruction.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: All the types are already known, and a more explicit solution is simpler
```cpp
[this] (ACompCamera const& camera) { this->draw(camera); }
```

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: The IcoSphereTree is designed to allow being shared between multiple PlanetGeometryA instances, even between separate scenes. It's likely that the initialization of IcoSphereTree will be moved to an entirely different function some time in the future.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: Yep. It was originally configured in a way that the planet will update only when pressing a button. As it became more complete, the `|| true` was added so it updates every frame.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: Right now, the most practical place to put this is in SysAreaAssociate, as it describes the physics area. However, I'm soon planning on removing states from the IDynamicSystem classes, possibly creating a component struct to store it instead. I'd rather keep this variable as-is, as too many other things around it will change.

src/planet-a/Active/SysPlanetA.cpp
Capital-Asterisk: Unfortunately, there isn't a constexpr sqrt in magnum nor std. The simplest solution possible that will work for sure is to just use the calculated result (0.9510565162). It is extremely unlikely that this equation will ever change. I'm not sure how reliable a constexpr float sqrt function will be.

src/planet-a/IcoSphereTree.cpp
Capital-Asterisk: New triangles are almost never created during runtime; existing triangle in the buffers are reused, so a constructor wouldn't be practical. A member function would be good.

src/osp/Universe.cpp
jonesmz: Right, but .view<const ...> doesn't necessarily mean anything unless you're making it return a const&, because, e.g.

int foo(void);
const int foo(void);

Can both be assigned to a non-const int.

So you'll need to either make it return a reference, or find a different way to provide immutability.

src/planet-a/Active/SysPlanetA.cpp
jonesmz: What is a "split screen" then? I just use my whole screen for my editor.

src/planet-a/Active/SysPlanetA.cpp
jonesmz: Put some documentation related to that somewhere ?

src/planet-a/Active/SysPlanetA.cpp
jonesmz: Sure

src/planet-a/Active/SysPlanetA.cpp
jonesmz: ok

src/planet-a/Active/SysPlanetA.cpp
jonesmz: well, i'd recommend removing the if-check then

src/planet-a/Active/SysPlanetA.cpp
jonesmz: No more or less reliable than a hand-specified outcome, i suppose.

src/planet-a/IcoSphereTree.cpp
jonesmz: correction
std::any_of

src/planet-a/IcoSphereTree.h
Capital-Asterisk: These handles are used to delete an observer from an IcoSphereTree's list, with 
```void IcoSphereTree::event_remove(IcoSphereTreeObserver::Handle_T)```
though this function implemented yet.

File: src/planet-a/IcoSphereTree.h (hunk 22 above)
Comment (Capital-Asterisk) [reply src/planet-a/IcoSphereTree.h:110]: There are similar patterns all over the code, where a uint8_t is used to denote a side of a triangle, or siblings. Arithmetic often is performed on these, like on line 560:
```cpp
out.m_translation = out.m_translation * 0.5f
                              + 0.5f * (side == (pTri->m_siblingIndex + 1) % 3);
```
All of these add and modulus operations can be wrapped into some kind of 'cycle' function.

src/planet-a/IcoSphereTree.h
Capital-Asterisk: The [] operator of std::vector is not constexpr

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: It's a struct of two uint32_t

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: This was previously set with magic numbers, the comment is redundant. The pow should guarantee it. Asserting if chunkDiv is too high is more reasonable.

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: Initializing them to 0 means they're referring to the first vertex, so they're initialized to gc_invalidVrtx. This vector is intended to be parallel with IcoSphereTree::m_vrtxBuffer

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: This code was written with a moon-sized planet in mind. The small planet in the 'simple' test universe gets messed up with this function. It displaces the terrain too much, making the planet into a mess, so it was disabled by adding the `* 0.0f`.

src/planet-a/IcoSphereTree.h
jonesmz: Sure, but it should still be an enum :-P

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: `std::set` is ordered, `std::unordered_set` is unordered for sure. The index buffer here has to be contiguous, as it's being sent to the GPU for rendering. When chunks are marked as deleted, valid chunks at the end of the buffer are moved to fill the deleted spaces in chunk_pack(). New chunks can be added before this moving is done, writing directly over deleted spaces. This small optimization makes sure that new chunks will take a space close to the start of the buffer, as spaces near the end of the buffer might be moved later.

src/planet-a/Active/SysPlanetA.cpp
z-adams: Having two files open side by side on the same monitor?

src/planet-a/IcoSphereTree.h
z-adams: A non-class enum? Enum classes are great but when the primary purpose of the enum is to use it as a named integral constant that is involved in arithmetic, it's sort of irritating to insert a verbose C++ cast every time it's used. I can see making it a C-enum though.

src/osp/Universe.cpp
Capital-Asterisk: It refuses to compile without the const for UCompTransformTraj.

src/planet-a/IcoSphereTree.cpp
Capital-Asterisk: Actually, if std::array was used instead of c-style arrays, then `rTri.m_corners = {a, b, c}` can be done.

src/planet-a/PlanetGeometryA.cpp
Capital-Asterisk: These are Vector3, they don't have iterators.

jonesmz: Gross. Alrighty.

PR #29, opened 2021-01-01
Shader instancing

z-adams:
Adds data-oriented shader instances; fixes crash-at exit. Sorry for the dual-purpose PR but the crash fix is a small change, was necessary for the shader stuff, and I implemented them at the same time. I just split them up into two files because the files they touched were almost completely separate, so it was easy to separate them into two commits.


src/osp/Universe.h
```diff
@@ -77,6 +77,11 @@ class Universe
     Universe(Universe &&move) = delete;
     ~Universe() = default;
 
+    /**
+     * Safely destroys the universe
+     */
+    void destroy();
```
z-adams, line 83: :)

src/osp/Active/SysVehicle.cpp
```diff
@@ -308,14 +306,22 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+            PhongShaderInstance si;
+            si.m_shaderProgram = glResources.get<adera::shader::Phong>("phong_shader");
+            si.m_textures = std::move(textureResources);
+            si.m_lightPosition = Vector3{10.0f, 15.0f, 5.0f};
+            si.m_ambientColor = 0x111111_rgbf;
+            si.m_specularColor = 0x330000_rgbf;
+            m_scene.reg_emplace<PhongShaderInstance>(currentEnt, std::move(si));
+
+            CompDrawableDebug& bBocks = m_scene.reg_emplace<CompDrawableDebug>(
+                        currentEnt, meshRes, adera::shader::Phong::draw, 0x0202EE_rgbf);
```
z-adams, line 324: Step 1. a drawable object is created; `adera::shader::Phong` is passed to its function pointer, and the shader instance component is added to the same entity

src/osp/Active/SysDebugRender.cpp
```diff
@@ -67,16 +70,6 @@ void SysDebugRender::draw(ACompCamera const& camera)
...
-                .setAmbientColor(0x111111_rgbf)
-                .setSpecularColor(0x330000_rgbf)
-                .setLightPosition({10.0f, 15.0f, 5.0f})
-                .setTransformationMatrix(entRelative)
-                .setProjectionMatrix(camera.m_projection)
-                .setNormalMatrix(entRelative.normalMatrix())
-                .draw(*(drawable.m_mesh));
-
-        //std::cout << "render! \n";
+        drawable.m_shader_draw(entity, m_scene, *drawable.m_mesh, camera, transform);
```
z-adams, line 73: Step 2. The render system calls the function pointer for the object

src/adera/Shaders/Phong.cpp
```diff
@@ -0,0 +1,27 @@
...
+    shader
+        .bindDiffuseTexture(*shaderInstance.m_textures[0])
+        .setAmbientColor(shaderInstance.m_ambientColor)
+        .setSpecularColor(shaderInstance.m_specularColor)
+        .setLightPosition(shaderInstance.m_lightPosition)
+        .setTransformationMatrix(entRelative)
+        .setProjectionMatrix(camera.m_projection)
+        .setNormalMatrix(entRelative.normalMatrix())
+        .draw(mesh);
+}
```
z-adams, line 27: Step 3. The static function fetches the correct shaderinstance component, gets the original shader object, and calls the rest of the draw logic

src/osp/Universe.h
jonesmz: But does it destroy *this* universe? Or the other one?

I choose the universe that was created for Zaphod Beeblebrox

z-adams: This one, hopefully more safely than the Trafalmadorian rocket fuel incident.

src/adera/Shaders/Phong.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
jonesmz, line 1: Copyright header

src/adera/Shaders/Phong.h
```diff
@@ -0,0 +1,40 @@
...
+#include <Magnum/Math/Color.h>
+#include "osp/Active/activetypes.h"
+#include "osp/Active/ActiveScene.h"
+#include "osp/Resource/Resource.h"
+
+namespace adera::shader
+{
+
+
+class Phong : public Magnum::Shaders::Phong
```
jonesmz, line 16: If you're only going to have public member functions, maybe just make it a struct?

src/osp/Active/Shader.h
```diff
@@ -0,0 +1,47 @@
...
+ * @param ACompCamera - Camera used to draw the scene
+ * @param ACompTransform - Object transformation data
+ */
+using ShaderDrawFnc_t = void (*)(
+    ActiveEnt,
+    ActiveScene&,
+    Magnum::GL::Mesh&,
+    ACompCamera const&,
+    ACompTransform const&);
+}
```
jonesmz, line 47: Add a new line to the end of files. Almost all programs bitch if there isn't one.

src/osp/Universe.cpp
```diff
@@ -37,6 +37,13 @@ Universe::Universe()
     m_root = sat_create();
 }
 
+void Universe::destroy()
```
jonesmz, line 40: This doesn't seem called by anything else?

jonesmz: Nevermind, found it.

src/osp/OSPApplication.cpp
```diff
@@ -24,9 +24,11 @@
  */
 #include "OSPApplication.h"
 
+#include "osp/Resource/blueprints.h"
+
 using namespace osp;
 
-OSPApplication::OSPApplication()
+OSPApplication::OSPApplication() : m_glResources("gl", "gl-resources")
```
jonesmz, line 31: I'd recommend doing this as an inline default.

E.g.

Package m_glResources{"gl", "gl-resources"};

where the variable is declared.

src/adera/Shaders/Phong.h
z-adams: Is it ok for a struct to inherit from a class? Seems weird

src/osp/Active/SysVehicle.cpp
```diff
@@ -308,14 +306,22 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
             }
 
             // by now, the mesh and texture should both exist
-
-            CompDrawableDebug& bBocks
-                    = m_scene.reg_emplace<CompDrawableDebug>(
-                        currentEnt, &(*meshRes), std::move(textureResources), m_shader.get(), 0x0202EE_rgbf);
+            using adera::shader::PhongShaderInstance;
+            Package& glResources = m_scene.get_application().get_gl_resources();
+            PhongShaderInstance si;
```
jonesmz, line 315: Why bother with the using statement above?

Why not just

adera::shader::PhongShaderInstance si;

?

src/osp/Active/SysVehicle.cpp
```diff
@@ -308,14 +306,22 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+            si.m_shaderProgram = glResources.get<adera::shader::Phong>("phong_shader");
+            si.m_textures = std::move(textureResources);
+            si.m_lightPosition = Vector3{10.0f, 15.0f, 5.0f};
+            si.m_ambientColor = 0x111111_rgbf;
+            si.m_specularColor = 0x330000_rgbf;
```
jonesmz, line 320: This is screaming for a constructor to me ....

src/osp/Active/SysVehicle.cpp
```diff
@@ -308,14 +306,22 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+            m_scene.reg_emplace<PhongShaderInstance>(currentEnt, std::move(si));
```
jonesmz, line 321: Does emplace take a variadic argument list?

If so, and you add a constructor to PhongShaderInstance, you can just pass the parameters to the constructor to the reg_emplace<> function directly.

src/osp/Active/SysVehicle.cpp
z-adams: Could just aggregate-initialize it, I just figured it was more explicit this way. These objects only get used in a couple places so it won't really clutter the codebase or anything, and it makes it easy to view the shader uniform values.

src/osp/Active/SysVehicle.cpp
z-adams: It gets used again on line 321 and makes the line too long, so I figured I'd just do this to shorten both 315 and 321

src/adera/Shaders/Phong.h
jonesmz: Yep.
Literally the only difference between a struct and a class is the default access permissions (public / protected / private)

struct = everything public by default
class = everything private by default

No other differences.

src/adera/Shaders/Phong.cpp
```diff
@@ -13,7 +13,7 @@ void Phong::draw(ActiveEnt e,
     PhongShaderInstance& shaderInstance = scene.reg_get<PhongShaderInstance>(e);
     Phong& shader = *shaderInstance.m_shaderProgram;
 
-    Magnum::Matrix4& entRelative = camera.m_inverse * transform.m_transformWorld;
+    Magnum::Matrix4 entRelative = camera.m_inverse * transform.m_transformWorld;
```
jonesmz, line 16: so if you were to use const& here to bind the entRelative variable to the temporary, and cause lifetime extension, then that technically allows for a situation where "camera.m_inverse * transform.m_transformWorld" returns a reference to a Matrix4 that exists elsewhere, or a situation where "camera.m_inverse * transform.m_transformWorld" returns a polymorphic object that has base class Matrix4.

Not saying you should do that, just pointing out the nifty detail of the language.

You can additionally do && instead of const&, and get lifetime extension that way, and then std::move() it later.

But my understanding is that matrix4 is a small POD type so doing that wouldn't do any good.

src/osp/Active/SysVehicle.cpp
```diff
@@ -310,18 +310,18 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+            shader.m_shaderProgram = glResources.get<Phong>("phong_shader");
+            shader.m_textures = std::move(textureResources);
+            shader.m_lightPosition = Vector3{10.0f, 15.0f, 5.0f};
+            shader.m_ambientColor = 0x111111_rgbf;
+            shader.m_specularColor = 0x330000_rgbf;
+            m_scene.reg_emplace<Phong::InstanceData>(currentEnt, std::move(shader));
 
             CompDrawableDebug& bBocks = m_scene.reg_emplace<CompDrawableDebug>(
-                        currentEnt, meshRes, adera::shader::Phong::draw, 0x0202EE_rgbf);
+                        currentEnt, meshRes, Phong::draw_entity, 0x0202EE_rgbf);
```
jonesmz, line 324: as a personal preference, i'd do 

&Phong::draw_entity

as the way to notate that it's a pointer-to-function.

src/osp/OSPApplication.cpp
```diff
@@ -28,7 +28,7 @@
 
 using namespace osp;
 
-OSPApplication::OSPApplication() : m_glResources("gl", "gl-resources")
+OSPApplication::OSPApplication()
```
jonesmz, line 31: no =default; ?

src/adera/Shaders/Phong.cpp
z-adams: It's fine by-value like this; I have zero idea why it compiles and runs with MSVC as it was. The Windows build passed (and it ran on my machine); I wouldn't have noticed anything if the linux build hadn't failed. I need to always remember to re-enable `permissive-` mode.

src/test_application/flight.cpp
```diff
@@ -153,10 +153,8 @@ void testapp::test_flight(std::unique_ptr<OSPMagnum>& pMagnumApp,
     // Disconnect ActiveArea
     sysArea.disconnect();
 
-    // workaround: wipe mesh resources because they're specific to the
-    // opengl context
-    rOspApp.debug_find_package("lzdb").clear<Magnum::GL::Mesh>();
-    rOspApp.debug_find_package("lzdb").clear<Magnum::GL::Texture2D>();
+    pMagnumApp->free_scenes();
+    rOspApp.shutdown();
```
Capital-Asterisk, line 157: So the application cannot be re-opened afterwards? When the flight scene closes, it saves its changes to the universe, not destroy it.

src/adera/Shaders/Phong.h
```diff
@@ -0,0 +1,63 @@
...
+namespace adera::shader
+{
+
+
+class Phong : protected Magnum::Shaders::Phong
+{
+public:
+    using Magnum::Shaders::Phong::Phong;
+
+    struct InstanceData
```
Capital-Asterisk, line 45: Prefix with "AComp". Maybe even just name it ACompPhongUniforms to be more obvious.

src/adera/Shaders/Phong.h
```diff
@@ -0,0 +1,63 @@
...
+#include <Magnum/Math/Color.h>
+#include "osp/Active/activetypes.h"
+#include "osp/Active/ActiveScene.h"
+#include "osp/Resource/Resource.h"
+
+namespace adera::shader
+{
+
+
+class Phong : protected Magnum::Shaders::Phong
```
Capital-Asterisk, line 40: This class is just a namespace and an alias for Magnum::Shaders::Phong. I don't think it needs to exist anymore.

Capital-Asterisk (changes requested):

src/adera/Shaders/Phong.h
z-adams: Did you read the conversation I had with mishtal on the Discord?

z-adams:
I'm going to close this and sort out the crash stuff in a separate PR first, and then I'll make another PR for the shader stuff and finish the suggested changes there.

PR #30, opened 2021-01-02
Improve GL Context-dependent resource management

z-adams:
This is the cleanest way to handle this issue I've found so far, but let me know if you have any better ideas. A new `Package` is stored inside OSPMagnum as `OSPMagnum::m_context` and stores handles to GL context data. Scenes receive a reference to the context data, so that systems may access meshes, shaders, textures, and so on. When the OSPMagnum instance is destroyed, the GL resources are automatically freed, leaving the rest alone. The main crash that happened even before shaders or anything just required a `destroy_universe()` call in the CLI loop's exit clause.

### DIFF
--- a/src/osp/Active/ActiveScene.cpp
+++ b/src/osp/Active/ActiveScene.cpp
@@ -38,8 +38,9 @@ void ACompCamera::calculate_projection()
 }
 
 
-ActiveScene::ActiveScene(UserInputHandler &userInput, OSPApplication &app) :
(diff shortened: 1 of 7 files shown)


jonesmz (approved):
Looks good to me.

Straight forward, to the point.

z-adams:
Tested this branch alongside my shader branch; had to add one more step since this seems to be the first time any components store a `DependRes<>` to anything. Adding an explicit `~OSPMagnum()` destructor to force scenes to be freed before the GL context data fixes everything.

Capital-Asterisk (approved):

PR #31, opened 2021-01-02
Add data-oriented instancable shaders

z-adams:
Latest version of data-oriented shaders, now built on top of new GL resource management system and with previous PR comments taken into account. Tested; flight scene launches, can be closed and re-launched, and program exits, all without crashing. Yay!

### DIFF
--- a/src/adera/Shaders/Phong.cpp
+++ b/src/adera/Shaders/Phong.cpp
@@ -22,22 +22,30 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-#include "OSPApplication.h"
+#include "Phong.h"
+#include <Magnum/Math/Matrix4.h>
 
-using namespace osp;
+using namespace osp::active;
+using namespace adera::shader;
 
-void OSPApplication::debug_add_package(Package&& p)
+void Phong::draw_entity(ActiveEnt e,
+    ActiveScene& rScene, 
+    Magnum::GL::Mesh& rMesh,
+    ACompCamera const& camera,
+    ACompTransform const& transform)
 {
-    auto const& [it, success] = m_packages.emplace(p.get_prefix(), std::move(p));
-    assert(success);
-}
+    auto& shaderInstance = rScene.reg_get<ACompPhongInstance>(e);
+    Phong& shader = *shaderInstance.m_shaderProgram;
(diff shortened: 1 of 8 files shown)


src/adera/Shaders/Phong.cpp
```diff
@@ -0,0 +1,51 @@
...
+using namespace osp::active;
+using namespace adera::shader;
+
+void Phong::draw_entity(ActiveEnt e,
+    ActiveScene& scene, 
+    Magnum::GL::Mesh& mesh,
+    ACompCamera const& camera,
+    ACompTransform const& transform)
+{
+    ACompPhongInstance& shaderInstance = scene.reg_get<ACompPhongInstance>(e);
```
Capital-Asterisk, line 37: `auto` keyword here, as reg_get already tells the type.

src/osp/Active/Shader.h
```diff
@@ -0,0 +1,47 @@
...
+ * @param ACompCamera - Camera used to draw the scene
+ * @param ACompTransform - Object transformation data
+ */
+using ShaderDrawFnc_t = void (*)(
+    ActiveEnt,
+    ActiveScene&,
+    Magnum::GL::Mesh&,
+    ACompCamera const&,
+    ACompTransform const&);
+}
```
Capital-Asterisk, line 47: Mesh can be made const here. I think there is a small possibility that a draw function needs to modify something in the scene, so it should be left mutable. ActiveScene argument in `Phong::draw_entity` should be renamed to `rScene` to indicate mutable reference.

Capital-Asterisk (approved):
Just a few minor nitpicks; otherwise. all good!

src/osp/Active/Shader.h
z-adams: I tried that already and unfortunately it can't, as `draw()` only accepts a `Mesh&` or `Mesh&&` (non-`const`). I think this is because the `GL::Mesh` type contains state-tracking logic to avoid re-binding the same buffer twice, etc. I'll change the argument name for the scene though.

Capital-Asterisk: Oh whoops. Anyways, all non-const references should have an 'r' prefix; this means the mesh too. Personally, I found this convention a lot more readable than I previously thought, especially with local variables.

z-adams: Sounds good, I'll make these changes and merge.

PR #32, opened 2021-01-03
Add part-based placeholder rocket

z-adams:
This adds a bunch of asset files, which contain (extremely) simple placeholder parts that nevertheless possess all basic config properties that exist thus far. They can be cobbled together into a simple upper-stage rocket which consists of a capsule, a fuselage, a main engine, and RCS thruster blocks. The resulting rocket currently replaces the single-part Stomper rocket.

This also adds some miscellaneous utility functions, and generalizes the main.cpp asset import using the new `string_concat()` function.

The parts are each assigned the appropriate machines, rather than all machines being lumped into the command module. This doesn't do much yet as things like physical thrusters aren't yet implemented; this is mostly in preparation for the upcoming series of commits which will add those features. The RCS thrusters are disabled for the time being until these features are added so that they will function properly.


src/osp/Resource/AssetImporter.cpp
```diff
@@ -243,6 +243,17 @@ DependRes<Mesh> osp::AssetImporter::compile_mesh(
     return pkg.add<Mesh>(meshData.name(), Magnum::MeshTools::compile(*meshData));
 }
 
+DependRes<Magnum::GL::Mesh> AssetImporter::compile_mesh(
+    const std::string& meshDataName, Package& package)
```
z-adams, line 247: Before anyone says anything, I know these should be `string_view`. I'll change it before merging.

src/test_application/main.cpp
```diff
@@ -199,8 +200,34 @@ void load_a_bunch_of_stuff()
...
+    // Load noise textures
+    const std::string noise256 = "noise256";
+    const std::string noise1024 = "noise1024";
+    const std::string n256path = osp::string_concat(datapath, noise256, ".png");
+    const std::string n1024path = osp::string_concat(datapath, noise1024, ".png");
+
+    osp::AssetImporter::load_image(n256path, lazyDebugPack);
+    osp::AssetImporter::load_image(n1024path, lazyDebugPack);
+    /*osp::AssetImporter::compile_tex(n256path, lazyDebugPack);
+    osp::AssetImporter::compile_tex(n1024path, lazyDebugPack);*/
```
z-adams, line 230: These aren't used yet, I'm leaving them here until next commit where I'll find a place to initialize them that exists within the GL context

src/osp/Resource/AssetImporter.cpp
```diff
@@ -243,6 +243,17 @@ DependRes<Mesh> osp::AssetImporter::compile_mesh(
...
+{
+    DependRes<MeshData> meshData = package.get<MeshData>(meshDataName);
+    if (meshData.empty())
+    {
+        return DependRes<Mesh>();
```
jonesmz, line 252: Could just do "return { }:" to signify "Whatever it is, make it the default".

src/osp/Resource/AssetImporter.cpp
```diff
@@ -269,6 +280,17 @@ DependRes<Texture2D> osp::AssetImporter::compile_tex(
     return package.add<Texture2D>(imageData.name(), std::move(tex));
 }
 
+DependRes<Magnum::GL::Texture2D> AssetImporter::compile_tex(
```
jonesmz, line 283: i wonder if it wouldn't be more ergonomic to have these "compile_XYZ" functions instead be template functions "compile" with the internal logic changing based on the type of the package.
```
template<typename PACKAGE_T>
void AssetImporter::compile(std::string_view imageDataName, Package& package) = delete;
template<>
void AssetImporter::compile<ImageData2D>(std::string_view imageDataName, Package& package)
{
    auto imgData = package.get<ImageData2D>(imageDataName);
    if (imgData.empty())
    {
        return { };
    }
    // do the same treatment for the function taking a DependRes<PACKAGE_T>
    return compile(imgData, package);
}
```

src/osp/Resource/Resource.h
```diff
@@ -145,11 +145,14 @@ class DependRes
     }
 
     bool empty() const { return m_empty; }
+    operator bool() const { return !empty(); }
```
jonesmz, line 148: Never make conversion operator to bool non-explicit.

Causes lots of screwups with people passing the type to a function and being surprised by conversions to bool.

src/osp/Resource/Resource.h
```diff
@@ -145,11 +145,14 @@ class DependRes
...
     std::string name() const { return m_it->first; }
 
     TYPE_T const& operator*() const { return m_it->second.m_data; }
     TYPE_T& operator*() { return m_it->second.m_data; }
 
+    TYPE_T const* operator->() const { return &m_it->second.m_data; }
```
jonesmz, line 154: constexpr
noexcept

src/osp/Resource/Resource.h
z-adams: As in make it `explicit`? Requiring a `static_cast<bool>(resource)` is even more cumbersome than `(!resource.empty())` so I'll just remove the bool operator then.

src/test_application/main.cpp
```diff
@@ -199,8 +200,34 @@ void load_a_bunch_of_stuff()
     osp::Package lazyDebugPack("lzdb", "lazy-debug");
 
     // Load sturdy glTF files
-    osp::AssetImporter::load_sturdy_file("OSPData/adera/spamcan.sturdy.gltf", lazyDebugPack);
-    osp::AssetImporter::load_sturdy_file("OSPData/adera/stomper.sturdy.gltf", lazyDebugPack);
+    const std::string_view datapath = {"OSPData/adera/"};
+    const std::vector<std::string_view> meshes =
```
jonesmz, line 204: This is a good example of the confusing way string lifetime works.

This is, of course, fine, since the std::string_view will always point to a string that is alive, since these strings are in the read-only data section

but if you pass these strings to a function that want to take ownership of the string data's lifetime, then it'll incur an allocate and copy.

This is why I made that shared_string class, because it provides explicit ownership behavior that also handles read-only-data (by having a default constructed m_lifetime). That allows for lifetime managing functions to accept the string whether it's in RO-Data or not, and get the optimal outcome in terms of runtime cost.

jonesmz (approved):

src/osp/Resource/AssetImporter.cpp
z-adams: Hmm, I'll consider it. I don't think we need any more versions beyond meshes and textures. Assuming these would be the only two variants needed, is there any point still to do this?

src/test_application/main.cpp
z-adams: I don't understand; is there anything wrong with what's going on here? The strings here only exist to be catted together and passed to the `load_sturdy_file()` function which is going to end up needing an allocate anyways, and whenever a function being called here requires ownership of the string argument, it just converts it into a new `std::string`. Is there something I should change to make this better? I guess you're just saying that if one of these strings were to be copied into multiple places, it could benefit from sharing the underlying memory?

bin/OSPData/adera/plume.sturdy.gltf
```diff
@@ -0,0 +1,138 @@
...
+    "textures" : [
+        {
+            "source" : 0
+        }
+    ],
+    "images" : [
+        {
+            "mimeType" : "image/png",
+            "name" : "plumebase",
+            "uri" : "plumebase.png"
```
z-adams, line 71: Note to self: get rid of plume mesh material data if possible

src/osp/Resource/Resource.h
z-adams: How do I tell when I should be using these? For instance, why do the `operator*()` above not need to be `constexpr`/`noexcept`?

src/osp/Resource/Resource.h
jonesmz: explicit bool operators continue working inside of if statements without needing to use either static cast, or !resource.empty().

src/osp/Resource/Resource.h
jonesmz: They are also good candidates for constexpr and noexcept.

noexcept -> if it can't throw an exception, it should be no except.
constexpr -> if it only does operations that can be conducted at compile time, such as just returning something, then make it constexpr.

src/test_application/main.cpp
jonesmz: Nothing you've got here is wrong, i was just using it as an example of where string ownership can be tricky.

src/osp/Resource/AssetImporter.cpp
jonesmz: If only two versions are needed then it doesn't really matter either way.

src/osp/Resource/Resource.h
z-adams: Oh, then I shouldn't have removed it -_- alright good to know

src/osp/Resource/Resource.h
z-adams: How do I know it can't throw an exception? For instance, if `m_it` doesn't point to a valid object, wouldn't dereferencing it cause an exception? Since I'm not particularly familiar with exceptions, it's sort of hard to know when one will never be thrown.

jonesmz: Whether or not *m_it throws an exception depends on the actual type of m_it.

The real determinant is:

"If this function throws an exception, is there anything meaningful that the program can actually do about it?"

if m_it == std::map::end(), then *m_it is undefined behavior, and you're program *should* crash in order to avoid a security issue. 

So the logical condition that would lead to the situation that might cause an exception to be thrown should, essentially, trigger the program terminating because it strongly implies that your program is in a borked situation anyway.

In that case, the function should be noexcept so that if an exception *is* thrown, the program terminates.

Most functions should be noexcept. They should only ever be not-noexcept if you are actually planning to *catch* the exception thrown from the function. E.g. the function is written with the intention of having a thrown exception *mean* something that can be acted on.

z-adams:
This is pushing the envelope of what should reasonably be combined into one PR but I made a few nearby changes like making Packages operate on `std::string_view` rather than `std::string const&`, by giving the resource map that fancy `std::less<>` thingy. I'll have to fix the plume material stuff later as I've clean forgotten the process of exporting meshes in the right format.

src/osp/Resource/Resource.h
Capital-Asterisk: This is the only real issue I have with this PR. Similarly, more explicit ifs like `a != 0` and `b != nulllptr` are more obvious and are preferred.

Capital-Asterisk (approved):

PR #34, opened 2021-01-04
Improve Newton Dynamics integration further

Capital-Asterisk:
This is a step towards making all systems stateless, along with various other improvements to SysNewton: 
* Move the Newton physics world into a component (ACompNwtWorld)
* Change many functions to static
* Make all communication to SysNewton happen only through components

Other systems that use physics no longer have to obtain an instance of SysNewton through `scene.dynamic_system_find`. They also no longer have to manually initialize rigid bodies with `create_body`, as SysNewton will now initialize rigid bodies when needed.

Notes:
* SysNewton is aliased to SysPhysics_t to act more generic


src/adera/Machines/Rocket.cpp
```diff
@@ -108,18 +107,18 @@ void SysMachineRocket::update_physics()
         else
         {
             // rocket's rigid body not set yet
-            auto body = m_physics.find_rigidbody_ancestor(ent);
+            auto [bodyEnt, pBody]
+                    = SysPhysics_t::find_rigidbody_ancestor(m_scene, ent);
```
jonesmz, line 111: not "auto const&" ?

src/adera/Machines/Rocket.cpp
```diff
@@ -108,18 +107,18 @@ void SysMachineRocket::update_physics()
...
                 std::cout << "no rigid body!\n";
                 continue;
             }
 
-            machine.m_rigidBody = body.first;
-            compRb = body.second;
-            compTf = m_scene.get_registry()
-                    .try_get<ACompTransform>(body.first);
+            machine.m_rigidBody = bodyEnt;
+            compRb = pBody;
```
jonesmz, line 120: If you're not accepting it by value, you probably want to use std::move().

src/adera/Machines/Rocket.h
```diff
@@ -51,7 +51,7 @@ class SysMachineRocket :
 
 private:
 
-    osp::active::SysPhysics &m_physics;
+    //osp::active::SysPhysics &m_physics;
```
jonesmz, line 54: Why comment out instead of remove?

src/osp/Active/SysNewton.cpp
```diff
@@ -28,63 +28,68 @@
...
+
+using osp::active::ACompNwtBody;
+using osp::active::ACompNwtWorld;
+using osp::active::ACompTransform;
 
 const std::string SysNewton::smc_name = "NewtonPhysics";
 
 // Callback called for every Rigid Body (even static ones) on NewtonUpdate
-void cb_force_torque(const NewtonBody* body, dFloat timestep, int threadIndex)
+void cb_force_torque(const NewtonBody* rBody, dFloat timestep, int threadIndex)
```
jonesmz, line 43: this is a pointer, not a reference?

src/adera/Machines/Rocket.cpp
Capital-Asterisk: bodyEnt is an entity, it is really small.

jonesmz: but pBody is a shared_pointer, isn't it?

jonesmz (approved):

z-adams:
Looking at the changes, it seems like mostly style/code improvements, apart from moving the newton state into a component and removing direct references between systems. Can you summarize any other major behavioral differences?

z-adams (approved):

src/adera/Machines/Rocket.cpp
```diff
@@ -63,7 +63,6 @@ std::vector<WireOutput*> MachineRocket::existing_outputs()
 
 SysMachineRocket::SysMachineRocket(ActiveScene &scene) :
     SysMachine<SysMachineRocket, MachineRocket>(scene),
-    m_physics(scene.dynamic_system_find<SysPhysics>()),
     m_updatePhysics(scene.get_update_order(), "mach_rocket", "wire", "physics",
                     std::bind(&SysMachineRocket::update_physics, this))
```
jonesmz, line 67: lambda?

src/adera/Machines/Rocket.cpp
jonesmz: Or is it a raw pointer?

src/osp/Active/SysNewton.cpp
```diff
@@ -28,63 +28,68 @@
...
-    NewtonBodySetForce(body, bodyPhy.m_netForce.data());
-    NewtonBodySetTorque(body, bodyPhy.m_netTorque.data());
+    NewtonBodySetForce(pBody, pBodyComp->m_netForce.data());
+    NewtonBodySetTorque(pBody, pBodyComp->m_netTorque.data());
 
     // Reset accumolated net force and torque for next frame
-    bodyPhy.m_netForce = {0.0f, 0.0f, 0.0f};
-    bodyPhy.m_netTorque = {0.0f, 0.0f, 0.0f};
+    pBodyComp->m_netForce = {0.0f, 0.0f, 0.0f};
+    pBodyComp->m_netTorque = {0.0f, 0.0f, 0.0f};
```
jonesmz, line 72: `{0.0f, 0.0f, 0.0f}` seems like something that should be a constant in a header file.

like

`static constexpr std::initializer_list<float, float, float> sgc_zeroVector = {0.0f, 0.0f, 0.0f}`

or something

src/osp/Active/SysNewton.cpp
```diff
@@ -28,63 +28,68 @@
...
 }
 
-ACompNwtBody::ACompNwtBody(ACompNwtBody&& move) :
-        m_body(move.m_body),
-        m_entity(move.m_entity),
+ACompNwtBody::ACompNwtBody(ACompNwtBody&& move) noexcept
+ : DataRigidBody(std::move(move))
+ , m_body(move.m_body)
+ , m_entity(move.m_entity)
```
jonesmz, line 78: consider doing

````
, m_body(std::exchange(move.m_body, {}))
, m_body(std::exchange(movem_entity, entt::null))
```

to make the "moved-from" object less likely to be accidentally "successfully" used.

src/osp/Active/SysNewton.cpp
```diff
@@ -28,63 +28,68 @@
...
     {
         NewtonBodySetUserData(m_body, this);
     }
 }
 
-ACompNwtBody& ACompNwtBody::operator=(ACompNwtBody&& move)
+ACompNwtBody& ACompNwtBody::operator=(ACompNwtBody&& move) noexcept
 {
     m_body = move.m_body;
     m_entity = move.m_entity;
```
jonesmz, line 91: std::exchange idiom

src/osp/Active/SysNewton.cpp
```diff
@@ -94,66 +99,129 @@ ACompNwtBody& ACompNwtBody::operator=(ACompNwtBody&& move)
...
+
+    // Iterate just the entities with ACompNwtBody
+    for (ActiveEnt ent : viewBody)
+    {
+        auto &entBody = viewBodyTransform.get<ACompNwtBody>(ent);
+
+        // temporary: delete if something is dirty
+        if (entBody.m_colliderDirty)
+        {
+            NewtonDestroyBody(entBody.m_body);
```
jonesmz, line 166: A long time ago I learned a very painful mistake about setting things to nullptr *after* destroying them. The long and short of it is that someone's gonna write some code that, somehow, manages to access your just-destroyed thing *after* it was destroyed but before it was set to nullptr, which then makes the sr. director of a different department call your boss really angerly demanding that you fix the code you broke, even though it turns out it was their code that was broken the whole time but it only managed to work by accident because the operating system didn't clear out the pointed-to-object after destruction so accessing it after it was freed still magically worked until i fixed my code.

jonesmz: As such, i recommend

NewtonDestroyBody(std::exchange(entBody.m_body, nullptr));

src/adera/Machines/Rocket.cpp
Capital-Asterisk: Yes, it is a raw pointer to ACompNwtBody*. The `find_rigidbody_ancestor` function will return nullptr if it doesn't find a ACompNwtBody. Function is defined here:

https://github.com/TheOpenSpaceProgram/osp-magnum/blob/2ccf99891dd4611ebac6b6a010b1f5c1d5056a99/src/osp/Active/SysNewton.cpp#L361

Capital-Asterisk:
> Looking at the changes, it seems like mostly style/code improvements, apart from moving the newton state into a component and removing direct references between systems that you mentioned above. Are there any other major behavioral differences?

Nope, nothing major. The only thing I forgot to mention is how accessing ACompNwtBody is now a bit more convenient. It use to have an `m_bodyData` struct member to store generic non-Newton-specific things, needing code like `body.m_bodyData.m_intertia = ...`. But now it uses inheritance, where one can now do `body.m_inertia = ...`

src/adera/Machines/Rocket.cpp
Capital-Asterisk: Hmm, I thought PR #35 would fix this, but I forgot.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: Hmm, interesting disaster. I hope the accepted fix was not removing the `= nullptr`

jonesmz: Na, we fixed it properly.

Basically
```
auto pOld = std::exchange(m_p, nullptr);
delete pOld;
return;
```
instead of
```
delete m_p;
m_p = nullptr;
return;
```

src/osp/Active/SysNewton.cpp
Capital-Asterisk: I think this is best kept as-is. A `{0.0f, 0.0f, 0.0f}` literal is very clear and obvious. I also considered these other solutions:
```cpp
// using explicit constructor Vector3(float)
pBodyComp->m_netForce = osp::Vector3{0.0f};

// template function for zero
template<typename T>
decltype(auto) zero() { return T(0); }
// ...
pBodyComp->m_netForce = osp::zero<osp::Vector3>();
```

Issue #13, opened 2020-10-12
Change a bunch of std::bind to lambdas


Capital-Asterisk:
Classes that use the FunctionOrder, most of them being system classes named "SysSomething," have something that look like this in their constructor:
```cpp
DebugCameraController::DebugCameraController(active::ActiveScene &scene,
                                             active::ActiveEnt ent) :
// ... stuff here
 m_updatePhysicsPost(scene.get_update_order(), "dbg_cam", "physics", "",
                std::bind(&DebugCameraController::update_physics_post, this)),
// ... more stuff here
```
What does this mean??? This constructs a FunctionOrderHandle member (named m_updatePhysicsPost) with an std::function as an argument, adding the function to the update order of the scene. This is saying that the `update_physics_post` function of this specific instance of DebugCameraController will be called after the "physics" update.

All of these should be changed to a lambda instead, as std::bind is hard to read and is discouraged:
```cpp
m_updatePhysicsPre(scene.get_update_order(), "dbg_cam_pre", "", "physics",
                [this] { this->update_physics_pre(); }),
```

Capital-Asterisk:
These are already all gone, or in pending PRs.

PR #35, opened 2021-01-04
Add ActiveScene& argument to update functons, and clean up related code

Capital-Asterisk:
This PR adds an `ActiveScene&` argument to the UpdateOrder definitions in `src/active/activetypes.h`, adding the new argument to every update function. This is another step towards making systems stateless. Related code was also cleaned up, such as adding an `_t` to typedefs, and fixing the inconsistent use of entt default types in ActiveScene. Issue #13 is also fixed.

### DIFF
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -33,16 +33,16 @@ using namespace osp;
 
 const std::string SysAreaAssociate::smc_name = "AreaAssociate";
 
-SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni) :
-       m_scene(rScene),
-       m_universe(uni),
-       m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
-                    [this] { this->update_scan(); })
+SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni)
+ : m_scene(rScene)
+ , m_universe(uni)
+ , m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
+                [this] (ActiveScene& rScene) { this->update_scan(rScene); })
 {
 
 }
 
-void SysAreaAssociate::update_scan()
+void SysAreaAssociate::update_scan(ActiveScene& rScene)
 {
     // scan for nearby satellites, maybe move this somewhere else some day
 
--- a/src/osp/Active/SysWire.cpp
+++ b/src/osp/Active/SysWire.cpp
@@ -51,15 +51,15 @@ void WireInput::doErase()
      list()->cut(this);
 };
 
-SysWire::SysWire(ActiveScene &scene) :
-        m_updateWire(scene.get_update_order(), "wire", "", "",
-                     std::bind(&SysWire::update_propigate, this))
+SysWire::SysWire(ActiveScene &scene)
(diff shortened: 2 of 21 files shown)


src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -33,16 +33,16 @@ using namespace osp;
...
 
-SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni) :
-       m_scene(rScene),
-       m_universe(uni),
-       m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
-                    [this] { this->update_scan(); })
+SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni)
+ : m_scene(rScene)
+ , m_universe(uni)
+ , m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
```
z-adams, line 39: This is weird, are you sure about this

src/osp/Active/SysWire.h
```diff
@@ -287,12 +287,12 @@ class SysWire : public IDynamicSystem
     SysWire(SysWire const& copy) = delete;
     SysWire(SysWire&& move) = delete;
 
-    void update_propigate();
+    void update_propigate(ActiveScene& rScene);
```
z-adams, line 290: Since this line is already being modified, might be a good time to fix typo `propigate`->`propagate`

z-adams (approved):
Everything looks good, although I'd appreciate an explanation for the commas

src/osp/Active/SysAreaAssociate.cpp
Capital-Asterisk: I've seen it around a few times, it's a pretty common style. Personally, I find it easier to work with. I never though of switching for consistency, but then jonesmz started using it in PR #19 so why not.

src/osp/Active/SysWire.cpp
```diff
@@ -51,15 +51,15 @@ void WireInput::doErase()
      list()->cut(this);
 };
 
-SysWire::SysWire(ActiveScene &scene) :
-        m_updateWire(scene.get_update_order(), "wire", "", "",
-                     std::bind(&SysWire::update_propigate, this))
+SysWire::SysWire(ActiveScene &scene)
+ : m_updateWire(scene.get_update_order(), "wire", "", "",
+                [this] (ActiveScene& rScene) { this->update_propagate(rScene); })
```
jonesmz, line 56: From a "don't repeat yourself" mindset, this same specific function gets called inside of a lambda in multiple places.

Makes me think that it would be better to have a specific struct that takes a pointer to the base class of SysWire and has an operator() function.

src/osp/Active/activetypes.h
```diff
@@ -43,13 +43,14 @@ struct ACompCamera;
 
 constexpr unsigned gc_heir_physics_level = 1;
 
-using ActiveEnt = entt::entity;
+enum class ActiveEnt: entt::id_type {};
```
jonesmz, line 46: Nice.

jonesmz (approved):

src/osp/Active/SysWire.cpp
Capital-Asterisk: With ActiveScene as an argument, many of these update functions will be turned into static functions soon, removing the need for this lambda in the first place.

PR #36, opened 2021-01-05
Plume effects

z-adams:
Rough summary of changes:

1st commit: adds a handy new system called "SysExhaustPlume" which attaches a "plume" component to rocket engines, and screams "ON!" whenever the throttle is greater than zero. Also adds a handy `hierarchy_traverse()` function to ActiveScene that allows an arbitrary callable to be called on each element of the hierarchy.

2nd commit: adds a new function to SysMachineRocket: `attach_plume_effect()`, which attempts to attach a plume effect to each engine on instantiation. Also adds functions to load in plume effect data from a glTF to attach the right one to an engine. Adds a CompVisibleDebug component to the rendering system, and uses all the new features to make a conical mesh with a diffuse texture appear whenever the throttle is on.

3rd commit: adds the plume shader. Modifies the rendering pipeline to support transparency, more convenient drawing of EnTT collections. Adds AssetImpoter functionality to retrieve plume effect parameters from a plume mesh glTF.

4th commit: uses LShift/LCtrl to smoothly vary throttle.

This will all need to be updated again after the existing system PRs get merged, but it shouldn't take much.

Also, since the spamcan aggregations and stomper rocket don't have plumes, it fails to give them one during init. I tested the moon scene and this didn't seem to cause any issues besides printing some messages, but I'll double check tomorrow to make sure it won't break anything.

### DIFF
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -31,6 +31,7 @@
 #include "Rocket.h"
 #include "osp/Resource/AssetImporter.h"
 #include "adera/Shaders/Phong.h"
+#include "adera/Shaders/PlumeShader.h"
 #include "adera/SysExhaustPlume.h"
 #include "adera/Plume.h"
 #include <Magnum/Trade/MeshData.h>
@@ -172,11 +173,6 @@ void SysMachineRocket::update_physics()
 
 void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
 {
-    using Magnum::GL::Mesh;
-    using Magnum::GL::Texture2D;
-    using Magnum::Trade::ImageData2D;
-    using namespace Magnum::Math::Literals;
-
     ActiveEnt plumeNode = entt::null;
 
     auto findPlumeHandle = [this, &plumeNode](ActiveEnt ent)
@@ -209,47 +205,11 @@ void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
     DependRes<PlumeEffectData> plumeEffect = pkg.get<PlumeEffectData>(effectName);
     if (plumeEffect.empty())
     {
-        using Magnum::Trade::MeshData;
-
-        DependRes<MeshData> meshData = pkg.get<MeshData>(plumeEffect->meshName);
-        if (meshData.empty())
-        {
-            std::cout << "ERROR: couldn't find mesh " << plumeEffect->meshName
-                << " for plume effect\n";
-        }
-
-        plumeMesh = AssetImporter::compile_mesh(meshData, glResources);
+        std::cout << "ERROR: couldn't find plume effect " << effectName << "!\n";
+        return;
     }
 
-    // Get plume tex (TEMPORARY: just grab noise1024.png)
-    const std::string texName = "OSPData/adera/noise1024.png";
-    DependRes<Texture2D> plumeTex = glResources.get<Texture2D>(texName);
-    if (plumeTex.empty())
-    {
-        using Magnum::Trade::ImageData;
-
-        DependRes<ImageData2D> imageData = pkg.get<ImageData2D>(texName);
-        if (imageData.empty())
-        {
-            std::cout << "ERROR: couldn't find texture " << texName
-                << " for plume effect\n";
-        }
-
-        plumeTex = AssetImporter::compile_tex(imageData, glResources);
-    }
-
-    // by now, the mesh and texture should both exist
-    using adera::shader::Phong;
-    Phong::ACompPhongInstance shader;
-    shader.m_shaderProgram = glResources.get<Phong>("phong_shader");
-    shader.m_textures = std::vector<DependRes<Texture2D>>{plumeTex};
-    shader.m_lightPosition = Vector3{10.0f, 15.0f, 5.0f};
-    shader.m_ambientColor = 0x111111_rgbf;
-    shader.m_specularColor = 0x330000_rgbf;
-    m_scene.reg_emplace<Phong::ACompPhongInstance>(plumeNode, std::move(shader));
-    m_scene.reg_emplace<CompDrawableDebug>(plumeNode, plumeMesh,
-        &adera::shader::Phong::draw_entity);
-    m_scene.reg_emplace<CompVisibleDebug>(plumeNode, false);
+    m_scene.reg_emplace<ACompExhaustPlume>(plumeNode, ent, plumeEffect);
 }
 
 Machine& SysMachineRocket::instantiate(ActiveEnt ent)
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -26,10 +26,15 @@
 
 #include <osp/Active/ActiveScene.h>
 #include <osp/Active/physics.h>
-
+#include <osp/Active/SysDebugRender.h>
 
 #include "Rocket.h"
+#include "osp/Resource/AssetImporter.h"
+#include "adera/Shaders/Phong.h"
 #include "adera/SysExhaustPlume.h"
+#include "adera/Plume.h"
+#include <Magnum/Trade/MeshData.h>
+#include <Magnum/Math/Color.h>
 
 using namespace adera::active::machines;
 using namespace osp::active;
@@ -165,9 +170,14 @@ void SysMachineRocket::update_physics()
     }
 }
 
-Machine& SysMachineRocket::instantiate(ActiveEnt ent)
+void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
 {
-    ActiveEnt plumeNode;
+    using Magnum::GL::Mesh;
+    using Magnum::GL::Texture2D;
+    using Magnum::Trade::ImageData2D;
+    using namespace Magnum::Math::Literals;
+
+    ActiveEnt plumeNode = entt::null;
 
     auto findPlumeHandle = [this, &plumeNode](ActiveEnt ent)
     {
@@ -199,14 +209,53 @@ Machine& SysMachineRocket::instantiate(ActiveEnt ent)
     DependRes<PlumeEffectData> plumeEffect = pkg.get<PlumeEffectData>(effectName);
     if (plumeEffect.empty())
     {
-        std::cout << "Found MachineRocket "
-            << static_cast<int>(ent) << "'s associated plume: "
-            << static_cast<int>(plumeNode) << "\n";
+        using Magnum::Trade::MeshData;
 
-        m_scene.reg_emplace<ACompExhaustPlume>(plumeNode, ent);
+        DependRes<MeshData> meshData = pkg.get<MeshData>(plumeEffect->meshName);
+        if (meshData.empty())
+        {
+            std::cout << "ERROR: couldn't find mesh " << plumeEffect->meshName
+                << " for plume effect\n";
+        }
+
+        plumeMesh = AssetImporter::compile_mesh(meshData, glResources);
     }
 
-    return m_scene.reg_emplace<MachineRocket>(ent);//emplace(ent);
+    // Get plume tex (TEMPORARY: just grab noise1024.png)
+    const std::string texName = "OSPData/adera/noise1024.png";
+    DependRes<Texture2D> plumeTex = glResources.get<Texture2D>(texName);
+    if (plumeTex.empty())
+    {
+        using Magnum::Trade::ImageData;
+
+        DependRes<ImageData2D> imageData = pkg.get<ImageData2D>(texName);
+        if (imageData.empty())
+        {
+            std::cout << "ERROR: couldn't find texture " << texName
+                << " for plume effect\n";
+        }
+
+        plumeTex = AssetImporter::compile_tex(imageData, glResources);
+    }
+
+    // by now, the mesh and texture should both exist
+    using adera::shader::Phong;
+    Phong::ACompPhongInstance shader;
+    shader.m_shaderProgram = glResources.get<Phong>("phong_shader");
+    shader.m_textures = std::vector<DependRes<Texture2D>>{plumeTex};
+    shader.m_lightPosition = Vector3{10.0f, 15.0f, 5.0f};
+    shader.m_ambientColor = 0x111111_rgbf;
+    shader.m_specularColor = 0x330000_rgbf;
+    m_scene.reg_emplace<Phong::ACompPhongInstance>(plumeNode, std::move(shader));
+    m_scene.reg_emplace<CompDrawableDebug>(plumeNode, plumeMesh,
+        &adera::shader::Phong::draw_entity);
+    m_scene.reg_emplace<CompVisibleDebug>(plumeNode, false);
+}
+
+Machine& SysMachineRocket::instantiate(ActiveEnt ent)
+{
+    attach_plume_effect(ent);
+    return m_scene.reg_emplace<MachineRocket>(ent);
 }
 
 Machine& SysMachineRocket::get(ActiveEnt ent)
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -29,6 +29,7 @@
 
 
 #include "Rocket.h"
+#include "adera/SysExhaustPlume.h"
 
 using namespace adera::active::machines;
 using namespace osp::active;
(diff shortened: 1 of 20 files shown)


src/osp/Resource/Resource.h
```diff
@@ -122,19 +122,15 @@ class DependRes
...
+            m_it->second.m_refCount--;
         }
-        m_empty = false;
+        m_empty = move.m_empty;
+        move.m_empty = true;
+
         m_it = std::move(move.m_it);
-        m_it->second.m_refCount ++;
 
         return *this;
```
z-adams, line 135: I seem to remember this fixing an important bug, but it's been so long that I forget what it was. Does this seem sane?

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+    // Get plume mesh
+    Package& glResources = m_scene.get_context_resources();
+    DependRes<Mesh> plumeMesh = glResources.get<Mesh>(plumeEffect->meshName);
+    if (plumeMesh.empty())
+    {
+        plumeMesh = AssetImporter::compile_mesh(plumeEffect->meshName, pkg, glResources);
+    }
+
+    // Get plume tex (TEMPORARY: just grab noise1024.png)
+    const std::string texName = "OSPData/adera/noise1024.png";
```
z-adams, line 225: If I have time, I may generalize this for this commit, although there's no rush as there's only one plume effect right now

src/osp/Active/SysDebugRender.cpp
```diff
@@ -46,30 +47,51 @@ SysDebugRender::SysDebugRender(ActiveScene &rScene) :
...
     Package& glResources = m_scene.get_context_resources();
-    glResources.add<adera::shader::Phong>("phong_shader",
-        adera::shader::Phong{Magnum::Shaders::Phong::Flag::DiffuseTexture});
+
+    using namespace adera::shader;
+
+    glResources.add<Phong>("phong_shader",
+        Phong{Magnum::Shaders::Phong::Flag::DiffuseTexture});
+
+    glResources.add<PlumeShader>("plume_shader");
```
z-adams, line 56: Until we find a better place to load shaders, I'm going to continue stashing them in the SysDebugRender constructor, but we should find a new place soon

src/adera/SysExhaustPlume.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto plumeView = m_scene.get_registry()
+        .view<ACompExhaustPlume, ShaderInstance_t, CompVisibleDebug>();
+
+    for (ActiveEnt plumeEnt : plumeView)
+    {
+        auto& plume = plumeView.get<ACompExhaustPlume>(plumeEnt);
+        auto& plumeShader = plumeView.get<ShaderInstance_t>(plumeEnt);
+        CompVisibleDebug& visibility = plumeView.get<CompVisibleDebug>(plumeEnt);
+
+        auto& machine = m_scene.reg_get<MachineRocket>(plume.m_parentMachineRocket);
```
Capital-Asterisk, line 66: I'm going to assume most of this is temporary, but rocket and plumes both depend on each other. Rockets have to initialize the plumes, and plumes are reading MachineRockets specifically.

A possible solution would be to add a new component for plume engine power, in the same entity as MachineRocket. MachineRocket can then write its throttle value into it. This allows other systems aside that aren't rockets to use plumes.

Capital-Asterisk (commented):
There's quite a few style issues here, most of them about style such as using `auto` and `r` prefix to mutable references. It doesn't seem practical for me to comment on all of them. I don't mind editing some of these myself; I'm just not sure if this should be merged where it can be edited further by multiple people, merged into a new branch, or I should just PR to your repo?

src/adera/SysExhaustPlume.cpp
z-adams: Systems like what? I suppose jet engines might be a different type of system? I figured that more general effects would have their own system, which is why this system and shader are designed so specifically for rocket exhaust plumes

Capital-Asterisk: I consider it generally cleaner to separate parts of the code when possible. Machines are designed to be replaceable by having the same wire connection names, almost like duck typing. It's likely that MachineRocket won't become a ultra-generalized to fit every rocket model possible. Someone might want their own MachineRealisticRocket.

z-adams: Hm ok. For now, a good middle ground would be to just have `ACompExhaustPlume` store the throttle value, and then make the MachineRocket or whatever other propulsion system write the throttle value, like you mentioned. That makes sense to me since SysExhaustPlume is purely graphics-oriented, so making this change would make it manage only the rendering/shader and be completely agnostic to how it gets its information. Does that sound ok?

z-adams: That's actually a much cleaner and more decoupled way, and I'm not sure how I missed it originally

Capital-Asterisk: If MachineRocket writes directly into its plumes, then this adds the issue of having to loop through them. I suggested a separate component to prevent this from happening. Also, engine power can also be made into a wire output. I might make some changes to the wire system that will make this easier to do, so it might be okay to just leave this as-is.

z-adams: Alright, if you're ok with it as-is I'll leave it and we can discuss/experiment and change it when we have a plan in place, or if any clear solutions present themselves when things progress.

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
     }
 }
 
+void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
```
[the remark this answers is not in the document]
z-adams: Some sort of auto-generated plume mesh can be sorted out, but that's probably worthy of its own PR and some thinking about how an engine part pack or something should specify its engine plumes. The secret purpose of this feature was basically to get myself acquainted with Magnum's graphics API and the engine; when we get to the point of having multiple engine parts and caring about game aesthetics we can take a closer look at some general engine effects generator and vertex shader and make everything look all nice.

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+{
+    using Magnum::GL::Mesh;
+    using Magnum::GL::Texture2D;
+    using Magnum::Trade::ImageData2D;
+    using namespace Magnum::Math::Literals;
+
+    ActiveEnt plumeNode = entt::null;
+
+    auto findPlumeHandle = [&](ActiveEnt ent)
```
jonesmz, line 182: discourage taking reference to all vars. Prone to mistakes.

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+    {
+        ACompHierarchy& node = m_scene.reg_get<ACompHierarchy>(ent);
+        if (node.m_name.compare(0, 9, "fx_plume_") == 0)
```
jonesmz, line 185:
```
static constexpr std::string_view comp = "fx_plume_";
if(node.m_name.compare(0, comp.size, comp.data()) == 0)
{ }
```
Or, my preference, which is yoda-conditionals

```
static constexpr std::string_view comp = "fx_plume_";
if(0 == node.m_name.compare(0, comp.size, comp.data()))
{ }
```

z-adams: Why do you prefer the latter? It looks weird. It's like saying
```cpp
3 = int myInt;
```
I know `==` is a commutative operation, but it still makes more sense to me to put the subject on the left.

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+        {
+            plumeNode = ent;
+            return false;  // terminate search
+        }
+        return true;
+    };
+
+    m_scene.hierarchy_traverse(ent, findPlumeHandle);
```
jonesmz, line 193: Instead of making your functor return a true or false, which is convienient, i know, define an enum class. E.g.
enum class EHierarchyTraverseStatus : bool
{
    Continue = true,
    Stop = false
};
(Or use uint8_t and 0, 1, if that doesn't compile)

Additionally, consider changing your hierarchy_traverse function to return decltype(auto), and have it return the result of your functor, that would let you change your lambda to return `{EHierarchyTraverseStatus::Stop, ent}` or `{EHierarchyTraverseStatus::Continue, entt::null}`

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+
+    if (plumeNode == entt::null)
+    {
+        std::cout << "ERROR: could not find plume anchor for MachineRocket "
+            << static_cast<int>(ent) << "\n";
+        return;
+    }
+    std::cout << "MachineRocket "
+        << static_cast<int>(ent) << "'s associated plume: "
```
jonesmz, line 202: I think it would be better to define an operator<<(std::ostream & os, ActiveEnt val) { return os << static_cast<int>(val); }

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+        << static_cast<int>(plumeNode) << "\n";
+
+    // Get plume effect
+    Package& pkg = m_scene.get_application().debug_find_package("lzdb");
+    std::string plumeAnchorName = m_scene.reg_get<ACompHierarchy>(plumeNode).m_name;
```
jonesmz, line 207: The underlying value from reg_get<...>(...) is stable in memory right? If so, take the plumeAchorName as either a std::string const&, or a std::string_view.

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+    std::string effectName = plumeAnchorName.substr(3, plumeAnchorName.length() - 3);
```
jonesmz, line 208: This allocates, use std::string_view{plumeAnchorName}.remove_prefix(3) instead

src/adera/Machines/Rocket.cpp
z-adams: Just fixed this moments ago in a yet-unpushed commit. Since the name is just used to query the resources, they can all be `std::string_view`

src/adera/Machines/Rocket.cpp
```diff
@@ -163,9 +170,86 @@ void SysMachineRocket::update_physics()
...
+    m_scene.reg_emplace<ACompExhaustPlume>(plumeNode, ent, plumeEffect);
+    
+    // Emplace plume shader instance component from plume effect parameters
+    using adera::shader::PlumeShader;
+    PlumeShader::ACompPlumeShaderInstance shader(
+        glResources.get<PlumeShader>("plume_shader"),
+        n1024,
+        n1024,
+        *plumeEffect);
+    m_scene.reg_emplace<PlumeShader::ACompPlumeShaderInstance>(plumeNode, std::move(shader));
```
jonesmz, line 241: just emplace the shader directly, don't construct it and then move it.

src/adera/Machines/UserControl.cpp
```diff
@@ -99,7 +101,34 @@ void SysMachineUserControl::update_sensor()
...
     {
         MachineUserControl &machine = view.get<MachineUserControl>(ent);
+        auto& throttlePos = std::get<wiretype::Percent>(machine.m_woThrottle.value()).m_value;
 
+        float throttleRate = 0.5f;
+        auto delta = throttleRate * m_scene.get_time_delta_fixed();
+
+        if (m_throttleMore.trigger_hold())
+        {
+            if (throttlePos + delta > 1.0f)
```
jonesmz, line 111: try std::clamp

src/adera/Machines/Rocket.cpp
z-adams: Searching for a certain entity is just one of many things you could do with this function; it could be for any process that needs to recursively touch all the nodes. Or am I misunderstanding the point of your last sentence?

src/adera/Shaders/PlumeShader.h
```diff
@@ -0,0 +1,140 @@
...
+        float m_currentTime;
+        float m_powerLevel;
+
+        // Constructor to auto-initialize from effect data
+        ACompPlumeShaderInstance(
+            osp::DependRes<PlumeShader> parent,
+            osp::DependRes<Magnum::GL::Texture2D> nozzleTex,
+            osp::DependRes<Magnum::GL::Texture2D> combustionTex,
+            PlumeEffectData const& data)
+            : m_shaderProgram(parent),
```
jonesmz, line 82: comma on the left? :-:

src/adera/Shaders/PlumeShader.h
```diff
@@ -0,0 +1,140 @@
...
+    PlumeShader& setMeshZBounds(float topZ, float bottomZ);
+    PlumeShader& bindNozzleNoiseTexture(Magnum::GL::Texture2D& tex);
+    PlumeShader& bindCombustionNoiseTexture(Magnum::GL::Texture2D& tex);
+    PlumeShader& setBaseColor(const Magnum::Color4 color);
+    PlumeShader& setFlowVelocity(const float vel);
+    PlumeShader& updateTime(const float currentTime);
+    PlumeShader& setPower(const float power);
+};
+
+}
```
jonesmz, line 140: // namespace adera::shaders

src/adera/SysExhaustPlume.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <Magnum/GL/Mesh.h>
+#include <Magnum/GL/Texture.h>
+#include <Magnum/Trade/ImageData.h>
+#include <Magnum/ImageView.h>
+#include <Magnum/GL/Sampler.h>
+#include <Magnum/GL/TextureFormat.h>
+
+using namespace osp::active;
+
+const std::string SysExhaustPlume::smc_name = "ExhaustPlume";
```
jonesmz, line 42: This variable can be defined directly in the class body by using "static inline std::string smc_name = "ExhaustPlume";

src/osp/Active/ActiveScene.h
```diff
@@ -332,4 +344,47 @@ struct ACompCamera
 };
 
 
+template<typename FUNC_T>
+void ActiveScene::hierarchy_traverse(ActiveEnt root, FUNC_T callable)
+{
+    using osp::active::ACompHierarchy;
+
+    std::vector<ActiveEnt> parentNextSibling;
```
jonesmz, line 352: For very small "stacks" like this, it seems to me like we need a generalized stack allocator.

Take a look here

 https://github.com/GameFoundry/bsf/blob/master/Source/Foundation/bsfUtility/Allocators/BsStackAlloc.h
 https://github.com/GameFoundry/bsf/blob/master/Source/Foundation/bsfUtility/Allocators/BsStackAlloc.cpp

Stack allocations are essentially free at runtime, so having a mechanism like that would be very nice.

src/osp/Active/ActiveScene.h
```diff
@@ -332,4 +344,47 @@ struct ACompCamera
...
+        if (hier.m_childCount)
+        {
+            // entity has some children
+            currentEnt = hier.m_childFirst;
+
+
+            // save next sibling for later if it exists
+            if (hier.m_siblingNext != entt::null)
+            {
+                parentNextSibling.push_back(hier.m_siblingNext);
```
jonesmz, line 371: Would it make sense to store an std::reference_wrapper to the ACompHierarchy into the vector instead of ActiveEnt?

src/osp/Active/ActiveScene.h
```diff
@@ -332,4 +344,47 @@ struct ACompCamera
...
+        } else if (hier.m_siblingNext != entt::null)
+        {
+            // no children, move to next sibling
+            currentEnt = hier.m_siblingNext;
+        } else if (parentNextSibling.size())
+        {
+            // last sibling, and not done yet
+            // is last sibling, move to parent's (or ancestor's) next sibling
+            currentEnt = parentNextSibling.back();
+            parentNextSibling.pop_back();
```
jonesmz, line 382: Looks like a job for https://en.cppreference.com/w/cpp/container/stack

src/osp/Resource/Resource.h
```diff
@@ -122,19 +122,15 @@ class DependRes
...
         if (!m_empty)
         {
-            // existing iterator is being overwritten, release it
-            m_it->second.m_refCount --;
+            // Existing iterator is being overwritten, decrease ref count
+            m_it->second.m_refCount--;
         }
-        m_empty = false;
+        m_empty = move.m_empty;
+        move.m_empty = true;
```
jonesmz, line 131: instead of m_empty, maybe there's room for std::optional<> to be used in part of this?

jonesmz (approved):

src/adera/Machines/Rocket.cpp
Capital-Asterisk: This is new to me, I like it. It makes it very obvious it's a comparison and not an assignment.

src/osp/Active/ActiveScene.h
Capital-Asterisk: It should be possible, but it would only rearrange the entire function without much improvements. `reg_get` would still have to be called for every entity iterated, which isn't expensive at all. References to components can be invalidated if new components are added.

src/adera/Machines/Rocket.cpp
jonesmz: because
```
if(ImARuntimeError = 4){}
if(4 = ImACompilerError){}
```

And yes, this is a real bug that I fixed in 25 year old code 10 years ago.

src/adera/Machines/Rocket.cpp
jonesmz: Yes, but the magic of templates :-)

If you write your function carefully (Or your lambda, depending on how crafty your metaprogramming game is) your hierarchy_traverse() function can serve double duty.

In the lazy way, you just always have your hierarchy_traverse() function return the second member of the thing returned from the lambda, and for lambdas that don't want to return anything, just have them return like nullptr or something.

In the crafty way, you have your hierarchy_traverse function return a tuple that represents all of the return values from the lambda except for index 0 (e.g. a tuple suffix).

If you write your hierarchy_traverse function carefully, there's very little cost to doing this. E.g. returning a tuple<EHierarchyTraverseStatus, std::reference_wrapper<>, std::reference_wrapper<>,....>, where the cost is basically just shuffling pointers around, which the compiler will heavily optimize for you cause of templates.

src/osp/Resource/Resource.h
Capital-Asterisk: Definitely. m_empty just tracks if its internal std::map iterator is valid or not. I guess using std::optional changes it to "valid iterator or nothing."

src/adera/Plume.h
```diff
@@ -0,0 +1,39 @@
...
+#include <Magnum/Math/Color.h>
+#include <Magnum/Magnum.h>
+
+struct PlumeEffectData
+{
+    float flowVelocity;
+    Magnum::Color4 color;
+    float zMin;
+    float zMax;
+    std::string meshName;
```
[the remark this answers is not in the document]
z-adams: Would the same thing apply to PrototypePart then? Right now PlumeEffectData is set up like ProtoPart, where it captures the names of the resources and then compiles/loads them upon instantiation. I think this is what you're describing, but I'm not positive. Either way, I will probably leave this as-is for now, and we can go change both this and PrototypePart together.

src/adera/Machines/Rocket.cpp
z-adams: What is it with you guys and horrendously ugly yet undeniably practical stylistic habits...

jonesmz: When you spend several years working with octogenarians who wrote a big software product using K&R C on actual Unix in the 70s, and then never learned anything since that point in their lives, and make terrible programming errors constantly that just accidentally work because C89 (Yea, they upgraded... once) is trash as possible...

you learn how to defend yourself from your co-workers.

PR #44, opened 2021-01-09
Simplify build for MSVC

EhWhoAmI:
MSVC creates different directories for Debug and release builds, and it does not copy all the necessary DLLs to the same directory. 

This reduces the need to copy files manually after each build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,14 @@ ELSE()
   ENDIF()
 ENDIF()
 
+# Force Runtime output directory into build/bin folder; force multiprocessor compilation for all projects
+IF(MSVC)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/$<CONFIG> CACHE PATH "")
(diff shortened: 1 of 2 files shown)


EhWhoAmI:
I'm gonna skip adding extra command line arguments because I have no idea how to do that.

jonesmz (approved):
As I said on discord, aside from considering the need to make these changes to be really odd, i don't see anything immediately obviously wrong, so I won't object to this being merged.

@z-adams or @Capital-Asterisk I'll leave it to one of you to merge this.

z-adams (changes requested):
I'd like to tweak this stuff so don't merge it yet please, I just haven't had time on my desktop to test it.

Capital-Asterisk:
I have never built on windows before, but I have a branch that sets up the CI to uploads binaries when finished, this might be interesting:

https://github.com/Capital-Asterisk/osp-magnum/tree/release-artifact

I had to manually merge together `build/bin/Release/*` and `build/Release/lib`, and add `build/bin/OSPData` to get it working (on wine). I'm assuming this change will make this unnecessary, which is great for distribution.

EhWhoAmI:
Yes, it should reduce the need to manually merge the folders together, which is extra annoying when you want a new build.

z-adams (approved):

PR #45, opened 2021-01-10
Physical RCS Part 1

z-adams:
I figured I'll do this one in two parts, and tackle the newton stuff in a second PR since it'll be a bit complicated.

Creates MachineRCSController and SysMachineRCSController, which are used to convert maneuver commands to RCS thruster rocket throttle values. It doesn't doesn't do anything physical yet; it just plays the plume effects on the RCS thrusters according to the maneuver commands. Later it will actually calculate physical thruster forces and get rid of the magical torques currently used.

### DIFF
--- a/bin/OSPData/adera/ph_rcs.sturdy.gltf
+++ b/bin/OSPData/adera/ph_rcs.sturdy.gltf
@@ -58,6 +58,9 @@
                 "manufacturer" : "ACME Placeholders",
                 "machines" : [
                     {
+                        "type" : "RCSController"
+                    },
+                    {
                         "type" : "Rocket"
                     },
                     {
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -1,0 +1,137 @@
+#include "RCSController.h"
+#include "osp/Active/ActiveScene.h"
+
+#include <Magnum/Math/Vector.h>
+
+using namespace adera::active::machines;
+using namespace osp::active;
+using Magnum::Vector3;
+
+
+MachineRCSController::MachineRCSController()
+    : Machine(true),
+    m_wiCommandOrient(this, "Orient"),
+    m_woThrottle(this, "Throttle")
+{
+    using wiretype::Percent;
+    m_woThrottle.value() = Percent{0.0f};
+}
+
+MachineRCSController::MachineRCSController(MachineRCSController&& move)
+    : osp::active::Machine(std::move(move)), m_wiCommandOrient(std::move(move.m_wiCommandOrient)), m_woThrottle(std::move(move.m_woThrottle))
+{
+
+}
+
+MachineRCSController& MachineRCSController::operator=(MachineRCSController&& move)
+{
+    m_enable = std::move(move.m_enable);
+    return *this;
+}
+
+void MachineRCSController::propagate_output(WireOutput* output)
+{
+
+}
+
+WireInput* MachineRCSController::request_input(osp::WireInPort port)
+{
+    return existing_inputs()[port];
+}
+
+WireOutput* MachineRCSController::request_output(osp::WireOutPort port)
+{
+    return existing_outputs()[port];
+}
+
+std::vector<WireInput*> MachineRCSController::existing_inputs()
+{
+    return {&m_wiCommandOrient};
+}
+
+std::vector<WireOutput*> MachineRCSController::existing_outputs()
+{
+    return {&m_woThrottle};
+}
+
+SysMachineRCSController::SysMachineRCSController(ActiveScene& scene)
+    :SysMachine<SysMachineRCSController, MachineRCSController>(scene),
+    m_updateControls(scene.get_update_order(), "mach_rcs", "wire", "controls",
+        std::bind(&SysMachineRCSController::update_controls, this))
+{
+}
+
+float SysMachineRCSController::thruster_influence(Vector3 posOset, Vector3 direction,
+    Vector3 cmdTransl, Vector3 cmdRot)
+{
+    using Magnum::Math::cross;
+    using Magnum::Math::dot;
+    using Magnum::Math::clamp;
+
+    Vector3 thrust = -direction.normalized();
+    
+    float rotInfluence = 0.0f;
+    float translInfluence = 0.0f;
+
+    if (cmdRot.length() > 0.0f)
+    {
+        Vector3 torque = cross(posOset, thrust).normalized();
+        rotInfluence = dot(torque, cmdRot.normalized());
+    }
+
+    if (cmdTransl.length() > 0.0f)
+    {
+        translInfluence = dot(thrust, cmdTransl.normalized());
+    }
+
+    float total = rotInfluence + translInfluence;
+
+    if (total < .01f)
+    {
+        // Ignore small contributions from imprecision
+        // Real thrusters can't throttle this deep anyways
+        return 0.0f;
+    }
+    return clamp(rotInfluence + translInfluence, 0.0f, 1.0f);
+}
+
+void SysMachineRCSController::update_controls()
+{
+    auto view = m_scene.get_registry().view<MachineRCSController, ACompTransform>();
+
+    for (ActiveEnt ent : view)
+    {
+        auto& machine = view.get<MachineRCSController>(ent);
+        auto& transform = view.get<ACompTransform>(ent);
+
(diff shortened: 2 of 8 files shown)


bin/OSPData/adera/ph_rcs.sturdy.gltf
```diff
@@ -57,6 +57,9 @@
                 "description" : "A placeholder linear RCS block",
                 "manufacturer" : "ACME Placeholders",
                 "machines" : [
+                    {
```
jonesmz, line 60: Completely unrelated to your work here, but it would amuse me enormously if someone were to create a pure constexpr gltf file parser. 

There's a pure constexpr json parser out there, so it should be possible (or nearly so anyway)

src/adera/Machines/RCSController.cpp
```diff
@@ -0,0 +1,125 @@
...
+using namespace osp::active;
+using Magnum::Vector3;
+
+
+MachineRCSController::MachineRCSController()
+    : Machine(true),
+    m_wiCommandOrient(this, "Orient"),
+    m_woThrottle(this, "Throttle")
+{
+    using wiretype::Percent;
```
jonesmz, line 16: That's a lot of using declarations....

Really necessary for one line?

src/adera/Machines/RCSController.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+float SysMachineRCSController::thruster_influence(Vector3 posOset, Vector3 direction,
+    Vector3 cmdTransl, Vector3 cmdRot)
+{
+    using Magnum::Math::cross;
+    using Magnum::Math::dot;
+    using Magnum::Math::clamp;
+
+    Vector3 thrust = -direction.normalized();
```
jonesmz, line 59: https://en.cppreference.com/w/cpp/utility/functional/negate

Much less likely for someone to mis-read the operator- there if you use std::negate

src/adera/Machines/RCSController.cpp
z-adams: What's a lot? Yeah here it's not necessary, but I'm curious what others you're referring to. The ones at the top of the file are pretty necessary.

src/adera/Machines/RCSController.cpp
```diff
@@ -0,0 +1,125 @@
...
+    
+    float rotInfluence = 0.0f;
+    float translInfluence = 0.0f;
+
+    if (cmdRot.length() > 0.0f)
```
jonesmz, line 64: What is the length of a rotation...?

src/adera/Machines/RCSController.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    float total = rotInfluence + translInfluence;
+
+    if (total < .01f)
+    {
+        // Ignore small contributions from imprecision
+        // Real thrusters can't throttle this deep anyways
+        return 0.0f;
+    }
+    return clamp(rotInfluence + translInfluence, 0.0f, 1.0f);
```
jonesmz, line 83: Is this any different than std::clamp? If not, just use std::clamp

src/adera/Machines/RCSController.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+void SysMachineRCSController::update_controls(ActiveScene& rScene)
+{
+    auto view = m_scene.get_registry().template view<MachineRCSController, ACompTransform>();
```
jonesmz, line 88: I know this is a thing...

But why have I gone almost my.whole career only using the template keyword to disambiguate a function call and OSP uses it everywhere?

src/adera/Machines/RCSController.h
```diff
@@ -0,0 +1,69 @@
...
+{
+public:
+    static inline std::string smc_name = "RCSController";
+
+    SysMachineRCSController(osp::active::ActiveScene &rScene);
+
+    void update_controls(osp::active::ActiveScene &rScene);
+
+    osp::active::Machine& instantiate(osp::active::ActiveEnt ent) override;
+
```
jonesmz, line 22: Copyright

jonesmz: Confused why github put this so far down the file. I meant it for the top of the file.

src/adera/Machines/RCSController.h
```diff
@@ -0,0 +1,69 @@
...
+    osp::active::Machine& get(osp::active::ActiveEnt ent) override;
+
+private:
+    float thruster_influence(Magnum::Vector3 posOset, Magnum::Vector3 direction,
+        Magnum::Vector3 cmdTransl, Magnum::Vector3 cmdRot);
+
+    osp::active::UpdateOrderHandle_t m_updateControls;
+};
+
+class MachineRCSController : public osp::active::Machine
```
jonesmz, line 32: More documentation for classes. E.g. doxygen block

jonesmz (approved):

src/adera/Machines/RCSController.cpp
z-adams: The rotation is represented by a vector describing its axis according to the right hand rule. Standard stuff, but I agree a comment or something could make it clearer what's going on.

src/adera/Machines/RCSController.cpp
z-adams: Nope, it's just Magnum's built-in version. I can switch it though.

src/adera/Machines/RCSController.cpp
jonesmz: Well. For my work projects I basically never use the using keyword to import things out of other namespaces into the current file.

The only exception is when I need to pull something from another namespace to make argument dependent lookup work (because there is no other way to do it. Thanks c++!!)

So a lot is "more than zero".

But both you and cap seem to like it, so when in Rome I guess.

src/adera/Machines/RCSController.h
```diff
@@ -0,0 +1,69 @@
...
+    SysMachineRCSController(osp::active::ActiveScene &rScene);
+
+    void update_controls(osp::active::ActiveScene &rScene);
+
+    osp::active::Machine& instantiate(osp::active::ActiveEnt ent) override;
+
+    osp::active::Machine& get(osp::active::ActiveEnt ent) override;
+
+private:
+    float thruster_influence(Magnum::Vector3 posOset, Magnum::Vector3 direction,
```
Capital-Asterisk, line 26: It would help to document this. I'm pretty sure it can be made static too.

src/adera/Machines/RCSController.cpp
Capital-Asterisk: I'm not sure why there's a template keyword here. It's not ambiguous and it compiles just fine without it.

Capital-Asterisk (approved):
It's all good so far. I tried running this and it behaves weirdly (ie. Rockets ignore rotation), which I assume is intentional as this is part 1. Only a few issues:
* It would be nice if there were a few more comments.
* RCS part does not have a collider
* SysMachineRCSController still uses m_scene. I'd encourage that static functions should be used, passing the scene through function arguments.

bin/OSPData/adera/ph_rcs.sturdy.gltf
z-adams: Oh man... how would one go about doing this? Sounds like an interesting exercise

src/adera/Machines/RCSController.cpp
z-adams: I only add it when it doesn't compile... I can try removing it again though.

bin/OSPData/adera/ph_rcs.sturdy.gltf
jonesmz: Start here: https://www.youtube.com/watch?v=PJwd4JLYJJY

it would probably be useful for commercial projects that absolutely must have initial application startup as fast as fucking possible

src/adera/Machines/RCSController.cpp
jonesmz: You're using VS2019 right? Double check you're up to date?

z-adams: I removed it and it compiles fine, but the problem is not my local compilation; MSVC will compile pretty egregiously noncompliant C++, the issue is that when I push the code the build fails on Clang or gcc or something. I'll let you know what happens when the build runs next time.

PR #48, opened 2021-01-10
Minor tweaks to cmake

jonesmz:


z-adams:
Ok I fixed the Magnum compilation errors, but now it failed to link Newton for some reason... no idea what's going on

z-adams:
Let's see if this builds now, I just rebased and turned shared libraries back on for Newton. I'm not sure why, but for some reason it refuses to work without shared libs.

PR #49, opened 2021-01-13
Physical RCS part 2

z-adams:
Alright this one is more significant. There are a few issues with the actual physics stuff (noted below) but the overall structure will otherwise be about like this so it's safe to give it a review.

The commit log for the main "implement physical RCS" commit summarizes the changes, but I'll walk through the order in which they propagate through everything to help any code reading:

`ACompMass` is a new component added to ActiveScene which simply stores a `float m_mass`. A part's `massdry` field is read in by AssetImporter and stored in a new `PrototypePart::m_mass` member (this is subject to change later, I can't remember). Upon instantiation in SysVehicle, this prototype mass value is used to instantiate an `ACompMass` and attach it to the entity representing that part.

When SysNewton creates a new rigidbody, a new function `compute_body_CoM()` is called which recursively adds up point-mass contributions from any `ACompMass`s in the tree, and returns a vector and the total mass(?). This center of mass is sent off to Newton, and also stored in the rigidbody component for external access.

When a system wants to do its fancy rocket engine update or whatever, it gets the center of mass offset from the RB origin, and uses this point for calculations involving torques, etc. DebugObject also now uses it to point the camera at the center of mass of the active vessel.


src/osp/Active/SysNewton.cpp
```diff
@@ -289,6 +290,8 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
                                 entBody.m_intertia.x(),
                                 entBody.m_intertia.x(),
                                 entBody.m_intertia.z());
+        NewtonBodySetCentreOfMass(entBody.m_body, centerOfMass.xyz().data());
+        entBody.m_centerOfMassOffset = centerOfMass.xyz();
```
z-adams, line 294: I just realized, I'm not sure where `entBody.m_mass` currently comes from. Is it set anywhere? Otherwise I'll set it with my own calculated value.

src/adera/Machines/Rocket.cpp
```diff
@@ -139,35 +111,28 @@ void SysMachineRocket::update_physics()
...
-
-            SysPhysics_t::body_apply_force(*compRb, thrustVec);
-        }
-
-
-
-        // this is suppose to be gimbal, but for now it applies torque
+            Vector3 thrustDir = relTransform.transformVector(Vector3{0.0f, 0.0f, 1.0f});
+            Vector3 thrust = thrustMag*percent->m_value * thrustDir;
+            Vector3 location = relTransform.translation();
```
z-adams, line 122: I realized that I don't use the center of mass offset here anywhere, and I don't think `find_rigidbody_ancestor()` gets it. Everything behaves physically find but I'll have to test it to make sure. I think it needs to be added.

src/adera/Machines/RCSController.cpp
```diff
@@ -127,16 +128,22 @@ void SysMachineRCSController::update_controls(ActiveScene& rScene)
...
             AttitudeControl *attCtrl = std::get_if<AttitudeControl>(gimbal);
 
             // TODO: RCS translation is not currently implemented, only rotation
             Vector3 commandTransl = Vector3{0.0f};
             Vector3 commandRot = attCtrl->m_attitude;
-            // HACK: hard-coded ship CoM, need to compute programmatically
-            Vector3 tmpOset{0.0f, 0.0f, -3.12f};
-            Vector3 thrusterPos = transform.m_transform.translation() - tmpOset;
+            Vector3 thrusterPos =
+                transform.m_transform.translation() - compRb->m_centerOfMassOffset;
```
z-adams, line 146: Similarly, I'm suspicious of this `transform` value from the Hierarchy. I think it's valid for now since the hierarchy structure of the vehicles is flat, but I don't think it holds in general. I think `find_rigidbody_ancestor()` is necessary to traverse the chain like in MachineRocket.

src/osp/Active/SysNewton.cpp
```diff
@@ -352,32 +355,84 @@ ACompNwtWorld* SysNewton::try_get_physics_world(ActiveScene &rScene)
...
+    while (currHier->m_level != gc_heir_physics_level);
 
-    auto *body = rScene.get_registry().try_get<ACompNwtBody>(prevEnt);
+    auto *body = rScene.get_registry().try_get<ACompRigidBody_t>(prevEnt);
 
     return {prevEnt, body};
 
 }
 
+std::pair<ACompNwtBody*, ACompTransform*> SysNewton::try_get_or_find_rigidbody_parent(
```
z-adams, line 391: I know this function name is kind of ridiculous but when you consider its function, it sort of fits. I'm open to any more concise ideas.

src/osp/Active/SysNewton.cpp
z-adams: It doesn't show up here because it's not directly adjacent to my code changes, but it's 1 line above this window. I suspect that it's never changed from its default initialization of `1.0f` and I should change it.

bin/OSPData/adera/ph_engine.sturdy.gltf
```diff
@@ -68,7 +68,7 @@
                         "type" : "FreeEnergyGenerator"
                     }
                 ],
-                "massdry" : 300,
+                "massdry" : 1000,
```
jonesmz, line 71: Annoyingly github doesn't allow me to comment on a whole file at a time.

is bin/OSPData/adera/ph_rcs.sturdy.bin  being deleted?

bin/OSPData/adera/ph_rcs.sturdy.gltf
```diff
@@ -8,49 +8,33 @@
...
-                0,
-                1,
-                0,
-                -1.6292068494294654e-07
-            ],
             "translation" : [
                 0,
                 0,
-                0.1351655125617981
+                -0.1351655125617981
```
jonesmz, line 21: These are awfully arbitrary.

Do we need named-constant support in the gltf files?

src/adera/Machines/RCSController.cpp
```diff
@@ -35,9 +35,10 @@ using Magnum::Vector3;
 
 
 MachineRCSController::MachineRCSController()
-    : Machine(true),
-    m_wiCommandOrient(this, "Orient"),
-    m_woThrottle(this, "Throttle")
+    : Machine(true)
+    , m_wiCommandOrient(this, "Orient")
```
jonesmz, line 39: Put these member variable initializations in the class declaration, not the constructor.

That'll leave you with a 2 line constructor, which I would argue should live in the header file as an "inline".

bin/OSPData/adera/ph_engine.sturdy.gltf
z-adams: No, why do you ask?

bin/OSPData/adera/ph_rcs.sturdy.gltf
z-adams: These aren't modified manually, they're exported from Blender and are bound to have ugly numbers. I do often try to model things with nice numbers but ultimately there are bound to be tons of arbitrary values so don't worry about it in these.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: I think this name is okay. It kind of makes sense, but passing an entity by reference seems a bit sketchy.

As a side note, multiple machine systems calling this function on their own for the same entity sounds more ridiculous. Perhaps a new component and system function can be added. An ACompRigidBodyAncestor managed by SysNewton that refers to an entity that has a rigid body, which can also track relative transform and stuff.

Capital-Asterisk (approved):

src/osp/Active/SysNewton.h
```diff
@@ -105,28 +107,55 @@ class SysNewton : public IDynamicSystem
...
+     * @param childEntity     [in] An entity whose rigidbody ancestor is sought
+     * @param rParentRigidBody [in/out] A reference to a machine member which
+     *                          either contains the entity owning the rigidbody
+     *                          ancestor component, or if not, said entity will
+     *                          be found and assigned as the value of the member
+     * 
+     * @return Pair of {pointer to found ACompNwtBody, pointer to ACompTransform
+     *         of the ACompNwtBody entity}. If either component can't be found,
+     *         returns {nullptr, nullptr}
+     */
```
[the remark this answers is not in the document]
z-adams: Yeah I struggled to document this function because its behavior is kind of complicated. Look at its actual definition, you'll recognize it from MachineRocket. It takes an in/out `ActiveEnt&` which receives the machine's handle on the rigid body ancestor. If it's null, the function finds it and fills it in, otherwise it just retrieves the rb component.

bin/OSPData/adera/ph_engine.sturdy.gltf
jonesmz: Because github says that 98% of the file is being deleted?

Why are .bin files checked into git? What do they represent?

bin/OSPData/adera/ph_rcs.sturdy.gltf
jonesmz: Ah, ok.

src/adera/Machines/RCSController.cpp
```diff
@@ -117,27 +108,36 @@ float SysMachineRCSController::thruster_influence(Vector3 posOset, Vector3 direc
...
 
         using wiretype::AttitudeControl;
         using Magnum::Vector3;
 
+        auto physComps = SysPhysics_t::try_get_or_find_rigidbody_parent(
+            rScene, ent, machine.m_rigidBody);
+        if (!physComps.first || !physComps.second) { continue; }
+
+        ACompRigidBody_t *compRb = physComps.first;
+        ACompTransform *compTf = physComps.second;
```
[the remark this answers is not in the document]
jonesmz: very much!

src/adera/Machines/RCSController.cpp
```diff
@@ -117,27 +108,36 @@ float SysMachineRCSController::thruster_influence(Vector3 posOset, Vector3 direc
...
+
         if (WireData *gimbal = machine.m_wiCommandOrient.connected_value())
         {
             AttitudeControl *attCtrl = std::get_if<AttitudeControl>(gimbal);
 
+            // TODO: cache this
+            Magnum::Matrix4 relTransform{};
+            SysPhysics_t::find_rigidbody_ancestor(rScene, ent, &relTransform);
```
jonesmz, line 133: I'd rather the function just return the Matrix4, instead of using a reference-parameter.

src/adera/Machines/RCSController.h
```diff
@@ -103,21 +103,32 @@ class MachineRCSController : public osp::active::Machine
...
     , m_woThrottle{this, std::move(move.m_woThrottle)}
+    , m_rigidBody(std::move(move.m_rigidBody))
 {}
 
 inline MachineRCSController& MachineRCSController::operator=(MachineRCSController&& move) noexcept
 {
     Machine::operator=(std::move(move));
     m_wiCommandOrient = {this, std::move(move.m_wiCommandOrient)};
     m_woThrottle = {this, std::move(move.m_woThrottle)};
+    m_rigidBody = std::move(move.m_rigidBody);
```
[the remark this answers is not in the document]
jonesmz: yes.

src/adera/Machines/Rocket.cpp
```diff
@@ -95,38 +94,12 @@ void SysMachineRocket::update_physics()
...
-                std::cout << "no rigid body!\n";
-                continue;
-            }
-
-            machine.m_rigidBody = bodyEnt;
-            compRb = pBody;
-            compTf = m_scene.get_registry().try_get<ACompTransform>(bodyEnt);
-        }
+        ACompRigidBody_t *compRb = physComps.first;
+        ACompTransform *compTf = physComps.second;
```
jonesmz, line 102: structured bindings

jonesmz: also prefix with lowercase p if these are pointers.

src/adera/Machines/Rocket.cpp
```diff
@@ -139,35 +112,30 @@ void SysMachineRocket::update_physics()
         {
             Percent *percent = std::get_if<Percent>(throttle);
```
jonesmz, line 113: p prefix

jonesmz: WireData is a variant?

src/osp/Active/SysNewton.cpp
```diff
@@ -352,32 +362,84 @@ ACompNwtWorld* SysNewton::try_get_physics_world(ActiveScene &rScene)
 }
 
 std::pair<ActiveEnt, ACompNwtBody*> SysNewton::find_rigidbody_ancestor(
-        ActiveScene& rScene, ActiveEnt ent)
+        ActiveScene& rScene, ActiveEnt ent, Matrix4* pTransform)
 {
     ActiveEnt prevEnt, currEnt = ent;
-    ACompHierarchy *currHeir = nullptr;
+    ACompHierarchy *currHier = nullptr;
```
jonesmz, line 368: p prefix

src/osp/Active/SysNewton.cpp
```diff
@@ -352,32 +362,84 @@ ACompNwtWorld* SysNewton::try_get_physics_world(ActiveScene &rScene)
...
 
     return {prevEnt, body};
 
 }
 
+std::pair<ACompNwtBody*, ACompTransform*> SysNewton::try_get_or_find_rigidbody_parent(
+    ActiveScene& rScene, ActiveEnt childEntity, ActiveEnt& rParentRigidBody)
+{
+    ACompNwtBody *compRb;
+    ACompTransform *compTf;
```
jonesmz, line 402: No reason to declare these variables here.

Declare them where they are defined, in both places.

jonesmz: Then return them inside the scope they're declared, instead of a single return at the end.

src/osp/Active/SysNewton.cpp
```diff
@@ -393,6 +455,45 @@ void SysNewton::body_apply_torque(ACompRigidBody_t &body, Vector3 torque) noexce
...
 }
 
+Vector4 SysNewton::compute_body_CoM(ActiveScene& rScene,
+    ActiveEnt root, Matrix4 currentTransform)
+{
+    ActiveEnt nextChild = root;
+    Vector3 localCoM{0.0f};
+    float localMass = 0.0f;
+
+    while (nextChild != entt::null)
```
jonesmz, line 465: This should be a for-loop

src/osp/Active/SysNewton.cpp
```diff
@@ -393,6 +455,45 @@ void SysNewton::body_apply_torque(ACompRigidBody_t &body, Vector3 torque) noexce
...
+
+        if (massComp)
+        {
+            float childMass = rScene.reg_get<ACompMass>(nextChild).m_mass;
+
+            Vector3 offset = childMatrix.translation();
+
+            localCoM += childMass * offset;
+            localMass += childMass;
+
```
jonesmz, line 481: empty space

src/osp/Active/SysNewton.cpp
```diff
@@ -393,6 +455,45 @@ void SysNewton::body_apply_torque(ACompRigidBody_t &body, Vector3 torque) noexce
...
+        }
+
+        Vector4 subCoM = compute_body_CoM(rScene, childHeir.m_childFirst, childMatrix);
+        localCoM += subCoM.w() * subCoM.xyz();
+        localMass += subCoM.w();
+
+        nextChild = childHeir.m_siblingNext;
+    }
+    if (localMass > 0.0f)
```
jonesmz, line 490: how could you get neg mass?

src/osp/Resource/PrototypePart.h
```diff
@@ -129,6 +129,10 @@ class PrototypePart
     constexpr std::vector<PrototypeMachine> const& get_machines() const
     { return m_machines; }
 
+    constexpr float& get_mass()
```
jonesmz, line 132: noexcept

src/test_application/universes/vehicles.cpp
```diff
@@ -218,7 +218,7 @@ osp::universe::Satellite testapp::debug_add_part_vehicle(
 
     Quaternion idRot;
     Vector3 scl{1};
-    Magnum::Rad qtrTurn(90.0_degf);
+    Magnum::Rad qtrTurn(-90.0_degf);
```
jonesmz, line 221: std::negate

jonesmz (approved):

bin/OSPData/adera/ph_engine.sturdy.gltf
Capital-Asterisk: These are bin files containing mostly mesh data. Perhaps its time to use LFS or similar solution?

z-adams: I suspect the 98% is meaning that 98% is remaining, as counterintuitive as it is. It says -256 bytes, and it's a 5kB file.

src/adera/Machines/RCSController.cpp
z-adams: The optional parameter allows the linear algebra to be skipped if the transformation isn't requested. I'll try to think up a cleaner solution though.

src/osp/Active/SysNewton.cpp
z-adams: You can't, I'm just checking for nonzero mass, and have an aversion to direct float comparisons

bin/OSPData/adera/ph_engine.sturdy.gltf
jonesmz: IMO git LFS is half-baked

For things under a megabyte, its not worth it.

src/adera/Machines/RCSController.cpp
jonesmz: Two different functions.

src/osp/Active/SysNewton.cpp
jonesmz: You should have an aversion to direct float comparisons 

Document in the code?

src/osp/Active/SysNewton.cpp
z-adams: This can't really be done easily due to the way hierarchy nodes are laid out, as far as I can tell

z-adams:
PR update, should fix most of the concerns. Generally, I removed the dual-purpose rigidbody_ancestor function and split it into one that finds the RBA, and one that finds the transformation. I added a component that stores the ancestor entity and relative transform, which is initialized by the `try_get_or_find_rigidbody_ancestor()` function and then accessed by the systems. I feel like it could still be improved somehow so that the ancestor components can be lazily added without requiring systems to call the `try_get_or_find` function rather than just using `reg.get<ACompRigidbodyAncestor>()` but this is what I've come up with for now. More detailed list of changes, from the most recent commit message:

Added `find_transform_rel_rigidbody_ancestor()`, which is similar to
`find_rigidbody_ancestor()` but returns the transformation that takes
the ancestor to the child rather than the ancestor itself.

Added ACompRigidbodyAncestor; it's filled automatically by
`try_get_or_find_rigidbody_ancestor()` and stores both the ancestor
entity itself and the relative transform between the two.
SysMachineRocket and SysRCSController access it via
`try_get_or_find_rigidbody_ancestor()`.

Added structured bindings everywhere

Fixed issue where switching between vessels that didn't have
MachineUserControl caused a crash

Vehicle modification code invalidates all ACompRigidbodyAncestors upon
splitting a vehicle into multiple islands

Disabled machines are not properly skipped by MachineRocket and
RCSController

Added a collider to the RCS block mesh

src/adera/Machines/RCSController.cpp
```diff
@@ -117,27 +109,36 @@ float SysMachineRCSController::thruster_influence(Vector3 posOset, Vector3 direc
...
+        auto const& rCompTf = rScene.reg_get<ACompTransform>(pRbAncestor->m_ancestor);
+
         if (WireData *gimbal = machine.m_wiCommandOrient.connected_value())
         {
             AttitudeControl *attCtrl = std::get_if<AttitudeControl>(gimbal);
 
+            Matrix4 transform = pRbAncestor->m_relTransform;
+
             // TODO: RCS translation is not currently implemented, only rotation
             Vector3 commandTransl = Vector3{0.0f};
```
jonesmz, line 138: Not really important, but personally i prefer to do this kind of thing as

Vector3 commandTransl{...}

over

Vector3 commandTransl = Vector3{...}

src/adera/Machines/RCSController.h
```diff
@@ -103,21 +103,32 @@ class MachineRCSController : public osp::active::Machine
     std::vector<osp::active::WireOutput*> existing_outputs() override;
 
 private:
-    osp::active::WireInput m_wiCommandOrient;
-    osp::active::WireOutput m_woThrottle;
+    osp::active::WireInput m_wiCommandOrient{this, "Orient"};
+    osp::active::WireOutput m_woThrottle{this, "Throttle"};
+
+    osp::active::ActiveEnt m_rigidBody{entt::null};
 };
```
jonesmz, line 110: recommend doing // class name of class here at the end of a class to make it easier for reviewers to recognize what class the code is for.

src/adera/Machines/Rocket.cpp
```diff
@@ -80,53 +79,25 @@ SysMachineRocket::SysMachineRocket(ActiveScene &scene) :
...
 
-void SysMachineRocket::update_physics()
+void SysMachineRocket::update_physics(ActiveScene& rScene)
 {
-    auto view = m_scene.get_registry().view<MachineRocket>();
+    auto view = m_scene.get_registry().view<MachineRocket, ACompTransform>();
 
     for (ActiveEnt ent : view)
     {
         auto &machine = view.get<MachineRocket>(ent);
```
jonesmz, line 88: out of curiousity, does EnTT not have a mechanism to do something like:

for(auto const& [things, from, the, view] : m_scene.get_registry().materialized_view<MachineRocket, ACompTransform>())
{

}

Directly?

Curious why we're forced to do the boilerplate of the view.get<>() call after we iterate the entity objects?

If it's not built into EnTT, we could probably make an iterator type fairly easily that would accomplish the task.

src/adera/Machines/Rocket.cpp
```diff
@@ -137,37 +108,29 @@ void SysMachineRocket::update_physics()
...
-
-            float thrust = 10.0f; // temporary
-
-            //std::cout << percent->m_value << "\n";
-
-            Vector3 thrustVec = compTf->m_transform.backward()
-                                    * (percent->m_value * thrust);
-
-            SysPhysics_t::body_apply_force(*compRb, thrustVec);
+            Percent *pPercent = std::get_if<Percent>(throttle);
```
jonesmz, line 111: Ah, ok, so 

```
// Supported data types
using WireData = std::variant<wiretype::Attitude,
                              wiretype::AttitudeControl,
                              wiretype::Percent,
                              wiretype::Deploy>;
```

Can't say I'm a huge fan of using std::variant here.

IMHO the wires should either be completely templated on the payload all the way through, so that the compiler is enforcing that the wire can't be connected to the wrong kind of thing. For example, you're gonna have a hard time plugging an ethernet cable into an HDMI port.

Or, if not templated throughput, then it should be completely type-erased, as in, void* + size_t, so that incorrectly connected wires get the correct "garbage in = garbage out" behavior that you would expect. At the receiving end of the wire data, we would need to do a size check to make sure we aren't evaluating data that is sized incorrectly, but beyond that, just do the "wrong thing".

src/adera/Machines/Rocket.cpp
```diff
@@ -137,37 +108,29 @@ void SysMachineRocket::update_physics()
...
+
+            Matrix4 relTransform = pRbAncestor->m_relTransform;
+
+            /* Compute thrust force
+             * Thrust force is defined to be along +Z by convention.
+             * Obtains thrust vector in rigidbody space
+             */
+            Vector3 thrustDir = relTransform.transformVector(Vector3{0.0f, 0.0f, 1.0f});
+            // Take thrust in rigidbody space and apply to RB in world space
+            Vector3 thrust = thrustMag*pPercent->m_value * thrustDir;
```
jonesmz, line 123: You're using pPercent with out making sure it's not nullptr.

src/osp/Active/SysNewton.cpp
jonesmz: Just do

for(ActiveEnt nextChild = root;
    nextChild != entt::null;
    )

to scope the nextChild variable.

src/osp/Active/SysNewton.h
```diff
@@ -169,6 +210,32 @@ class SysNewton : public IDynamicSystem
...
+    * 
+    * @param rScene           [in] ActiveScene containing relevant scene data
+    * @param root             [in] Entity at the root of the hierarchy subtree
+    * @param currentTransform [in] Transformation of root with respect to the
+    *                              initial call's root
+    * 
+    * @return A 4-vector containing xyz=CoM, w=total mass
+    */
+    static Magnum::Vector4 compute_body_CoM(ActiveScene& rScene,
+        ActiveEnt root, Matrix4 currentTransform = Magnum::Matrix4());
```
jonesmz, line 228: if you're not going to namespace Maxtrix4, then drop the Magnum::

src/osp/Active/SysVehicle.cpp
```diff
@@ -212,8 +212,8 @@ StatusActivated SysVehicle::activate_sat(ActiveScene &scene,
 
     // temporary: make the whole thing a single rigid body
     auto& vehicleBody = scene.reg_emplace<ACompRigidBody_t>(vehicleEnt);
-    ACompCollisionShape& vehicleShape = scene.reg_emplace<ACompCollisionShape>(vehicleEnt);
-    vehicleShape.m_shape = ECollisionShape::COMBINED;
+    auto& vehicleColShape = scene.reg_emplace<ACompCollisionShape>(vehicleEnt);
```
jonesmz, line 215: scene.reg_emplace<ACompCollisionShape>(vehicleEnt) = ECollisionShape::COMBINED;

src/osp/Active/SysVehicle.cpp
```diff
@@ -356,17 +358,26 @@ void SysVehicle::update_vehicle_modification(ActiveScene& rScene)
...
             // Separation requested
 
             // mark collider as dirty
             auto &rVehicleBody = m_scene.reg_get<ACompRigidBody_t>(vehicleEnt);
             rVehicleBody.m_colliderDirty = true;
 
+            // Invalidate all ACompRigidbodyAncestors
+            auto invalidateAncestors = [&rScene](ActiveEnt e)
+            {
+
```
jonesmz, line 374: Unneeded empty space

src/test_application/DebugObject.cpp
```diff
@@ -176,14 +180,15 @@ void DebugCameraController::update_physics_post()
...
+        targetValid = rReg.valid(m_orbiting);
 
         if (targetValid)
         {
             // enable the first MachineUserControl
             ActiveEnt firstPart = *(view.get(m_orbiting).m_parts.begin());
 
-            m_scene.reg_get<MachineUserControl>(firstPart).enable();
+            auto* pMUserCtrl = rReg.try_get<MachineUserControl>(firstPart);
+            if (pMUserCtrl != nullptr) { pMUserCtrl->enable(); }
```
jonesmz, line 191: if(auto* pMUserCtrl = rReg.try_get<MachineUserControl>(firstPart);
   pMUserCtrl != nullptr)
 { pMUserCtrl->enable(); }
 
 I guess it could go either direction. I imagine ya'll won't like this version.

src/test_application/DebugObject.cpp
```diff
@@ -207,6 +212,16 @@ void DebugCameraController::update_physics_post()
...
     Matrix4 const& xformTgt = m_scene.reg_get<ACompTransform>(m_orbiting).m_transform;
 
+    using osp::active::SysPhysics_t;
+    // Compute Center of Mass of target, if it's a rigid body
+    auto [rbEnt, pCompRb] = SysPhysics_t::find_rigidbody_ancestor(m_scene, m_orbiting);
+    Vector3 comOset{0.0f};
+    if (pCompRb != nullptr)
+    {
+        comOset = xformTgt.transformVector(pCompRb->m_centerOfMassOffset);
+    }
```
jonesmz, line 222:
```
Vector3 com0set =     pCompRb
                               ? xformTgt.transformVector(pCompRb->m_centerOfMassOffset);
                               : Vector3{0.0f};
```

src/adera/Machines/Rocket.cpp
jonesmz: You could change

```
if (WireData *throttle = machine.m_wiThrottle.connected_value())
```

to

```
if (Percent *pPercent = std::get_if<Percent>(machine.m_wiThrottle.connected_value()))
```

jonesmz: Also, this should be an if-initilizer expression. E.g.
```
if (WireData *throttle = machine.m_wiThrottle.connected_value(); throttle != nullptr)
```

jonesmz: also, throttle needs a p prefix :-)

src/adera/Machines/RCSController.cpp
z-adams: I do too usually, but in this case I do it this way because `Vector3{0.0f}` is sort of a placeholder, when there are proper translational controls It'll be `Vector3 commandTransl = something` so I kept the left side that way as a subtle reminder. I think it makes it a bit clearer.

src/adera/Machines/Rocket.cpp
z-adams: I've wondered about that too, that would be pretty slick. I'll make an issue for it and we can do it later.

src/adera/Machines/Rocket.cpp
z-adams: Paging @Capital-Asterisk, I didn't design the system so I'm not really in a place to reason about the long-term consequences of doing those things.

src/test_application/DebugObject.cpp
z-adams: I guess it's just what you're used to, but most of the time I find multi-statement `if` conditions equally or less readable than their normal counterparts. Particularly when squeezing them together causes the condition to run over multiple lines, it just becomes more difficult to tell which parts are the condition and which are the body. I'll do it in some places so continue suggesting it, but I'm going to leave it as-is in this particular instance.

src/osp/Active/SysNewton.cpp
z-adams: Check the new version and see if you think it's an improvement

src/test_application/DebugObject.cpp
z-adams: I'm going to leave this one as is since that way basically takes the same number of lines but is just less common and obvious compared to the way it is now. If it easily fit on one line I'd use a ternary conditional but it's kind of long and messy

jonesmz: Do consider the benefit of being able to mark the variable as const, and initialize it once and only once, instead of multiple places how you are now.

z-adams: Hmm fair point, I'll consider it

src/adera/Machines/Rocket.cpp
Capital-Asterisk: Yep, the wiring system definitely needs some beefing up. The entire thing needs to be nuked and rebuilt. I've written an issue on it, but I haven't submitted it yet.

PR #50, opened 2021-01-13
Rework Satellite Type system

Capital-Asterisk:
Satellites have a type to determine if it's a planet, vehicle, star, or whatever. This is mostly just a hint at what components the Satellite contains. Types can be registered at runtime by initializing `ITypeSatellite`-derived class (see main.cpp `register_universe_types()` function ). Satellites are assigned types by adding a `UCompType` to them, which contains a pointer to an `ITypeSatellite`. Types are separate instances so that they can be added and removed polymorphically; but as it turns out, this is over-complicated and really isn't needed.

If a function needs to create a planet in the universe, it has to know what a planet is. If a function needs to create a vehicle in the universe, it has to know what a vehicle is. There is no point in being able to create a 'something' polymorphically.

This PR will gut much of Universe.h. It removes ITypeSatellite entirely, replacing it with an integer enum class `TypeSatIndex`. The index of a type depends on which types were registered first. The first type ever registered is 0, next type is 1, third is 2, etc...

Many functions had been made static as a result. Obtaining instances of `ITypeSatellite`s through `sat_type_find` is no longer needed. The Sat*.cpp source files are now very empty, but there is a strong possibility that new functions will be added to them. These changes will simply many things needed for the next upcoming PR related to SysAreaAssociate.

### DIFF
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -101,7 +101,7 @@ void SysAreaAssociate::update_scan(ActiveScene& rScene)
 
         // Satellite is near! Attempt to load it
 
-        sat_activate(sat, view.get(sat));
+        sat_activate(sat);
     }
 
 }
@@ -167,8 +167,8 @@ void SysAreaAssociate::sat_transform_update(ActiveEnt ent)
     satPosTraj.m_dirty = true;
 }
 
-void SysAreaAssociate::activator_add(universe::ITypeSatellite const* type,
-                                         IActivator &activator)
+void SysAreaAssociate::activator_add(TypeSatIndex type,
+                                     IActivator &activator)
 {
     m_activators[type] = &activator;
 }
@@ -199,8 +199,7 @@ void SysAreaAssociate::floating_origin_translate(Vector3 translation)
 }
 
 
-int SysAreaAssociate::sat_activate(universe::Satellite sat,
-                                   universe::UCompActivatable &satAct)
+int SysAreaAssociate::sat_activate(universe::Satellite sat)
 {
 
     if (m_activatedSats.find(sat) != m_activatedSats.end())
--- a/src/osp/Satellites/SatActiveArea.cpp
+++ b/src/osp/Satellites/SatActiveArea.cpp
@@ -43,9 +43,3 @@ using Magnum::Matrix4;
 using namespace osp::universe;
 
 const std::string SatActiveArea::smc_name = "ActiveArea";
-
-SatActiveArea::SatActiveArea(Universe& universe) :
-        CommonTypeSat<SatActiveArea, UCompActiveArea>(universe)
-{
-
-}
--- a/src/osp/Satellites/SatActiveArea.cpp
+++ b/src/osp/Satellites/SatActiveArea.cpp
@@ -22,24 +22,17 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-#include <iostream>
-
-#include <Magnum/GL/DefaultFramebuffer.h>
 
 #include "SatActiveArea.h"
 
-#include "../Active/ActiveScene.h"
-#include "../Active/SysNewton.h"
-#include "../Active/SysMachine.h"
+using osp::universe::UCompActiveArea;
+using osp::universe::SatActiveArea;
 
-#include "../Universe.h"
+UCompActiveArea& SatActiveArea::add_active_area(Universe &rUni, Satellite sat)
+{
+    bool typeSetSuccess = rUni.sat_type_try_set(
+                sat, rUni.sat_type_find_index(SatActiveArea::smc_name));
+    assert(typeSetSuccess);
 
-
-// for the 0xrrggbb_rgbf and angle literals
-using namespace Magnum::Math::Literals;
-
-using Magnum::Matrix4;
-
-using namespace osp::universe;
-
-const std::string SatActiveArea::smc_name = "ActiveArea";
+    return rUni.get_reg().emplace<UCompActiveArea>(sat);
+}
--- a/src/osp/Satellites/SatActiveArea.h
+++ b/src/osp/Satellites/SatActiveArea.h
@@ -28,20 +28,11 @@
 
 #include <map>
 
-//#include <Magnum/Math/Color.h>
-//#include <Magnum/PixelFormat.h>
-//#include <Magnum/SceneGraph/Camera.h>
-
 #include "../types.h"
-//#include "../scene.h"
+#include "../Active/activetypes.h"
 
 #include "../Resource/Package.h"
 #include "../Resource/PrototypePart.h"
-
-//#include "../Active/ActiveScene.h"
-//#include "../OSPApplication.h"
-//#include "../UserInputHandler.h"
-//#include "../Active/FtrNewtonBody.h"
 
 namespace osp::universe
 {
@@ -50,56 +41,40 @@ namespace osp::universe
 class OSPMagnum;
 class SatActiveArea;
 
+struct UCompActivatable { };
+
+struct UCompActivationMutable
+{
+    Satellite m_area{entt::null};
+    active::ActiveEnt m_ent{entt::null};
+};
+
+struct UCompActivationRadius
+{
+    float m_radius;
+};
 
 struct UCompActiveArea
 {
-    //active::ActiveEnt m_camera;
 
-    unsigned m_sceneIndex;
+    // for later use
+    //active::MapActiveScene_t::iterator m_scene;
 
     // true when the ActiveArea is moving
     bool m_inMotion;
 };
 
 
-class SatActiveArea : public CommonTypeSat<SatActiveArea, UCompActiveArea>
+class SatActiveArea : CommonTypeSat<SatActiveArea, UCompActiveArea>
 {
+    using Base_t = CommonTypeSat<SatActiveArea, UCompActiveArea>;
 
 public:
 
     static const std::string smc_name;
 
-    SatActiveArea(Universe& universe);
-    ~SatActiveArea() = default;
-
-    /**
-     * Setup magnum scene and sets m_loadedActive to true.
-     * @return only 0 for now
-     */
-    //int activate(OSPApplication& app);
-
-    /**
-     * Do actual drawing of scene. Call only on context thread.
-     */
-    void draw_gl();
-    
-    /**
-     * Load nearby satellites, Maybe request a floating origin translation,
-     * then calls update_physics of ActiveScene
-     */
-    void update_physics(float deltaTime);
-
-    std::string get_name() { return smc_name; };
-
-private:
-
-    //bool m_loadedActive;
-
-    //ActiveEnt m_camera;
-
-    //std::shared_ptr<ActiveScene> m_scene;
-
-    //UserInputHandler& m_userInput;
+    using Base_t::add_get_ucomp_all;
+    using Base_t::add_get_ucomp;
 
 };
 
--- a/src/osp/Satellites/SatActiveArea.h
+++ b/src/osp/Satellites/SatActiveArea.h
@@ -65,17 +65,20 @@ struct UCompActiveArea
 };
 
 
-class SatActiveArea : CommonTypeSat<SatActiveArea, UCompActiveArea>
+class SatActiveArea
 {
-    using Base_t = CommonTypeSat<SatActiveArea, UCompActiveArea>;
-
 public:
 
-    static const std::string smc_name;
+    static constexpr std::string_view smc_name = "ActiveArea";
 
-    using Base_t::add_get_ucomp_all;
-    using Base_t::add_get_ucomp;
-
+    /**
+     * Set the type of a Satellite and add a UCompActiveArea to it
+     * @param rUni [out] Universe containing satellite
+     * @param sat  [in] Satellite add a UCompActiveArea to
+     * @return Reference to UCompActiveArea created
+     */
+    static UCompActiveArea& add_active_area(
+        osp::universe::Universe& rUni, osp::universe::Satellite sat);
 };
 
 }
--- a/src/osp/Universe.cpp
+++ b/src/osp/Universe.cpp
@@ -50,6 +50,16 @@ void Universe::sat_remove(Satellite sat)
     m_registry.destroy(sat);
 }
 
+bool Universe::sat_try_set_type(Satellite sat, TypeSatIndex type)
+{
+    auto &satType = m_registry.get<UCompType>(sat);
+    if (satType.m_type == TypeSatIndex::Invalid)
+    {
+        satType.m_type = type;
+        return true;
+    }
+    return false; // Type is already set
+}
 
 Vector3s Universe::sat_calc_pos(Satellite referenceFrame, Satellite target) const
 {
@@ -86,3 +96,15 @@ TypeSatIndex Universe::sat_type_find_index(std::string_view name)
 
     return TypeSatIndex::Invalid;
 }
+
+bool Universe::sat_type_try_set(Satellite sat, TypeSatIndex type)
+{
+    auto &satType = m_registry.get<UCompType>(sat);
+    if (satType.m_type != TypeSatIndex::Invalid)
+    {
+        return false;
+    }
+    satType.m_type = type;
+    return true;
+}
+
--- a/src/osp/Universe.cpp
+++ b/src/osp/Universe.cpp
@@ -75,4 +75,14 @@ Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target
     return Vector3(sat_calc_pos(referenceFrame, target)) / 1024.0f;
 }
 
+TypeSatIndex Universe::sat_type_find_index(std::string_view name)
+{
+    auto foundIt = m_typeSatIndices.find(name);
 
+    if (foundIt != m_typeSatIndices.end())
+    {
+        return foundIt->second;
+    }
+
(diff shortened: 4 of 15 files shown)


src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -199,8 +199,7 @@ void SysAreaAssociate::floating_origin_translate(Vector3 translation)
 }
 
 
-int SysAreaAssociate::sat_activate(universe::Satellite sat,
-                                   universe::UCompActivatable &satAct)
+int SysAreaAssociate::sat_activate(universe::Satellite sat)
```
jonesmz, line 202: This looks like it's returning an int with only 3 values, and the rules for understanding what the meaning of the int kind of "assumed".

Better would be to define an enum that clearly indicates what each return value is intended to indicate.

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -199,8 +199,7 @@ void SysAreaAssociate::floating_origin_translate(Vector3 translation)
...
 {
 
     if (m_activatedSats.find(sat) != m_activatedSats.end())
```
jonesmz, line 205: Got a lot of left over code in this function? Maybe remove commented out code and instead put more inline documentaitons.

src/osp/Active/SysAreaAssociate.cpp
jonesmz, line 202: I think you said that Satellite, in this case, is just an identifier.

Maybe we should be naming the types that are just handles / IDs into the registry with names that indicate they are such.

E.g. SatelliteHandle_t, SatelliteId_t

src/osp/Active/SysAreaAssociate.cpp
jonesmz: 8 lines below this, you have "satType".

What is "satType" supposed to represent? a type_id ?

src/osp/Satellites/SatActiveArea.cpp
```diff
@@ -43,9 +43,3 @@ using Magnum::Matrix4;
 using namespace osp::universe;
 
 const std::string SatActiveArea::smc_name = "ActiveArea";
```
jonesmz, line 45: This can live in the .h file as a "static inline std::string const smc_name = "ActiveArea";

jonesmz: Though do consider if it can instead be

static constexpr std::string_view smc_name = "ActiveArea"

src/osp/Satellites/SatActiveArea.h
```diff
@@ -28,78 +28,53 @@
...
+    // for later use
+    //active::MapActiveScene_t::iterator m_scene;
 
     // true when the ActiveArea is moving
     bool m_inMotion;
 };
 
 
-class SatActiveArea : public CommonTypeSat<SatActiveArea, UCompActiveArea>
+class SatActiveArea : CommonTypeSat<SatActiveArea, UCompActiveArea>
```
jonesmz, line 68: Private inheritance?

src/osp/Universe.cpp
```diff
@@ -75,4 +75,14 @@ Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target
     return Vector3(sat_calc_pos(referenceFrame, target)) / 1024.0f;
 }
 
+TypeSatIndex Universe::sat_type_find_index(std::string_view name)
+{
+    auto foundIt = m_typeSatIndices.find(name);
 
+    if (foundIt != m_typeSatIndices.end())
```
jonesmz, line 82: recommend c++17 if-initializer

src/osp/Universe.h
```diff
@@ -29,24 +29,23 @@
...
-
-
-using MapSatType = std::map<std::string, std::unique_ptr<ITypeSatellite>>;
+enum class Satellite : entt::id_type {};
 
-//typedef uint64_t Coordinate[3];
-//typedef Math::Vector3<int64_t> Coordinate;
+enum class TypeSatIndex : uint16_t
+{
+    Invalid = std::numeric_limits<uint16_t>::max()
```
jonesmz, line 47: no more than 16 thousand satellites?

src/osp/Universe.h
```diff
@@ -71,7 +70,12 @@ using MapSatType = std::map<std::string, std::unique_ptr<ITypeSatellite>>;
  */
 class Universe
 {
+    using MapTypeSats_t = std::map<std::string_view, TypeSatIndex, std::less<>>;
```
jonesmz, line 73: The string_view is pointing to a location that's stable in memory, right?

src/osp/Universe.h
```diff
@@ -122,27 +126,31 @@ class Universe
...
-     * Tries to locate an element in the map of registered Satellites by name.
-     * @param name [in] Name of ITypeSatellite
-     * @return Map iterator directly from std::map::find()
+     * @return Names of registered satellites. Access using a TypeSatIndex
      */
-    MapSatType::iterator sat_type_find(const std::string &name)
-    {
-        return m_satTypes.find(name);
-    }
+    constexpr std::vector<std::string_view> const& sat_type_get_names()
```
jonesmz, line 135: noexcept

src/osp/Universe.h
```diff
@@ -122,27 +126,31 @@ class Universe
...
+    /**
+     * Find index of a registered satellite by type. The type should contain a
+     * static member string named smc_name
+     * @tparam SATTYPE_T Satellite type containing smc_name
+     * @return Index of type, TypeSatIndex::Invalid if not found.
      */
     template<typename SATTYPE_T>
-    SATTYPE_T& sat_type_find();
+    TypeSatIndex sat_type_find_index()
+    { return sat_type_find_index(SATTYPE_T::smc_name); }
```
jonesmz, line 153: I'm wondering if it doesn't make sense to use the EnTT type_id system (Or at least the version of it that I'm reading about in 3.6) to do the name lookup.

It has a name() member function.

src/osp/Universe.h
```diff
@@ -152,44 +160,35 @@ class Universe
     template<typename TRAJECTORY_T, typename ... ARGS_T>
     TRAJECTORY_T& trajectory_create(ARGS_T&& ... args);
 
-    constexpr entt::basic_registry<Satellite>& get_reg() { return m_registry; }
-    constexpr const entt::basic_registry<Satellite>& get_reg() const
+    constexpr Registry_t& get_reg() { return m_registry; }
```
jonesmz, line 163: noexcept

src/osp/Universe.h
```diff
@@ -224,112 +222,62 @@ struct UCompTransformTraj
...
-
-private:
-
-    Universe& m_universe;
-};
-
-template<typename TYPESAT_T, typename ... UCOMP_T>
-std::tuple<UCOMP_T& ...> CommonTypeSat<TYPESAT_T, UCOMP_T ...>::add_get_ucomp_all(Satellite sat)
+template<typename UCOMP_T>
+constexpr decltype(auto) emplace_filter_tags(Universe& rUni, Satellite sat)
```
jonesmz, line 243: Could you document the goal behind returning an empty tuple if the requested type is empty?

Where is this used that that represents a big saving?

src/osp/Universe.h
```diff
@@ -224,112 +222,62 @@ struct UCompTransformTraj
...
+        // Return reference tuple for normal struct
+        UCOMP_T &comp = rUni.get_reg().emplace<UCOMP_T>(sat);
+        return std::forward_as_tuple(comp);
     }
-
-    type.m_type = this;
-    return std::get<0>(std::forward_as_tuple((m_universe.get_reg().emplace<UCOMP_T>(sat)) ...));
 }
 
 template<typename TYPESAT_T, typename ... UCOMP_T>
```
jonesmz, line 259: I'm having a hard time with this naming convention for UCOMP_T

I feel like it should be plural, but what's the "U" prefix represent? It's used in multiple places.

src/osp/Universe.h
```diff
@@ -224,112 +222,62 @@ struct UCompTransformTraj
...
-void CommonTypeSat<TYPESAT_T, UCOMP_T ...>::remove(Satellite sat)
+class CommonTypeSat
 {
-    auto& type = m_universe.get_reg().get<UCompType>(sat);
-
-    if (type.m_type != this)
+public:
+    static decltype(auto) add_get_ucomp_all(Universe& rUni, Satellite sat)
```
jonesmz, line 263: noexcept?

src/osp/Universe.h
```diff
@@ -224,112 +222,62 @@ struct UCompTransformTraj
...
+        assert(type.m_type == TypeSatIndex::Invalid);
+
+        type.m_type = rUni.sat_type_find_index(TYPESAT_T::smc_name);
+        return std::tuple_cat(emplace_filter_tags<UCOMP_T>(rUni, sat) ...);
     }
 
-    type.m_type = nullptr;
-    m_universe.get_reg().remove<UCOMP_T ...>(sat);
-}
+     static decltype(auto) add_get_ucomp(Universe& rUni, Satellite sat)
```
jonesmz, line 273: constexpr

src/test_application/main.cpp
```diff
@@ -357,12 +357,16 @@ void debug_print_hier()
...
+    using osp::universe::Universe;
     using osp::universe::UCompTransformTraj;
     using osp::universe::UCompType;
 
-    osp::universe::Universe const &universe = g_osp.get_universe();
+    Universe const &rUni = g_osp.get_universe();
 
-    auto const view = universe.get_reg().view<const UCompTransformTraj,
+    std::vector<std::string_view> const& typeSatNames
+            = g_osp.get_universe().sat_type_get_names();
```
jonesmz, line 367: You get the universe, but then don't use rUni?

jonesmz (approved):

src/osp/Active/SysAreaAssociate.cpp
Capital-Asterisk: No worries, a revamp of SysAreaAssociate is the topic of my next PR.

src/osp/Active/SysAreaAssociate.cpp
Capital-Asterisk: This function will be gone on my next PR. `satType` is a UCompType struct. It determines the type of a Satellite by storing an TypeSatIndex.

src/osp/Active/SysAreaAssociate.cpp
Capital-Asterisk: An ECS entity is kind of its own special thing. It's just an id, but it's treated like a container for components. Maybe SatelliteEnt, or UniverseEnt in the same way ActiveEnt is named?

src/osp/Satellites/SatActiveArea.h
Capital-Asterisk: Should this be explicit?

src/osp/Universe.cpp
Capital-Asterisk: like this?
```cpp
if (auto foundIt = m_typeSatIndices.find(name); foundIt != m_typeSatIndices.end())
```
I think it looks pretty weird. It feels like this hurts readability to save a line. I guess there might be some benefit if there was a long series of similar if statements, but this isn't the case here.

src/osp/Active/SysAreaAssociate.cpp
jonesmz: That's an easy enough naming convention to teach someone, so i'd be fine with it.

src/osp/Universe.h
Capital-Asterisk: Only 16 thousand different types of satellites. All I can think of is Planet, Vehicle, ActiveArea, Star, BlackHole, Asteroid, and I doubt anyone will be able to think about a thousand different more.

src/osp/Satellites/SatActiveArea.h
jonesmz: Generally I prefer to explicitly say public / private for the inheritance because it changes the question people ask from

"Did you know this makes it private inheritance?"
to
"Could you explain why you want private inheritance?"

I don't object to private inheritance, just wasn't sure it's what you meant.

src/osp/Universe.h
Capital-Asterisk: Yes, all of these point to constants

src/osp/Universe.cpp
jonesmz: Sure, either way.

Personally I prefer the consistency, and the reduction of variable scope, but it doesn't make a big difference.

src/osp/Universe.h
jonesmz: Note; We can manually specialize the type_id for known types, and I believe we can make the ids use the scheme that EnTT defaults to even while specializing it.

That consolidates the type names to a single place, right?

Capital-Asterisk: If I remember correctly, it uses a few non-standard compiler features, and can return different values. This is something that should be consistent across all clients in a multiplayer scenario.

src/osp/Universe.h
jonesmz: Additionally, you can safely narrow the scope of your if-constexpr, i believe, to only the return line, to slightly cut down on duplicated lines.

Compiler will properly handle the reference variable being unused.

src/osp/Universe.h
jonesmz: Then reduce it to uint8_t from uint16_t.

I doubt there's a meaningful difference between more than 100 different types of celestial bodies, much less 255.

src/osp/Universe.h
jonesmz: That's why I'm saying to specialize the type_id for each struct / class that we're currently providing an smc_name.

By doing the specialization, we eliminate compiler compatibility problems, and we also concentrate the strings to a single place without any loss of functionality, *and* we make the places where the strings are used a bit sexier.

src/osp/Universe.h
Capital-Asterisk: Compilation fails when getting a reference to an empty struct, like UCompActiveArea. This is used by `add_get_ucomp_all`, which is suppose to emplace multiple components through variadic templates, and return a tuple of references all of them. This function together with `std::tuple_cat` can filter out the empty structs. It took a bit of trial and error to figure this out, but I learned quite a bit more metaprogramming on the way.

jonesmz: I guess I'm just a bit sceptical that with a template function like this that the existence of the tag struct is actually kept at the assembly level.

But it's good that you learned more metaprogramming :-)

src/osp/Universe.h
Capital-Asterisk: "UComp" is for "Universe Component" structs for use in ECS. It parallels "AComp" for "Active Component."

src/osp/Universe.h
```diff
@@ -224,112 +222,62 @@ struct UCompTransformTraj
...
+        return std::forward_as_tuple(comp);
     }
-
-    type.m_type = this;
-    return std::get<0>(std::forward_as_tuple((m_universe.get_reg().emplace<UCOMP_T>(sat)) ...));
 }
 
 template<typename TYPESAT_T, typename ... UCOMP_T>
-void CommonTypeSat<TYPESAT_T, UCOMP_T ...>::remove(Satellite sat)
+class CommonTypeSat
```
z-adams, line 260: Some documentation for this class and its functions would be nice, it seems to be used everywhere but its function is not immediately obvious

z-adams (approved):
Looks fine to me, although some documenting comments would be nice. I don't really understand all the universe/activearea/areaassociate stuff very well so I'll take your word for everything here.

src/osp/Universe.h
jonesmz: > Compilation fails when getting a reference to an empty struct, like UCompActiveArea. This is used by add_get_ucomp_all,

Wait. What?

That's not true at all.

```
struct empty {};
void foo()
{
    empty bob;
    empty const& alice = bob;
}
```
That'll work just fine.

jonesmz:
```
template<typename UCOMP_T>
constexpr std::tuple<UCOMP_T&> emplace_filter_tags(Universe& rUni, Satellite sat)
{
    // Return reference tuple for normal struct
    return { rUni.get_reg().emplace<UCOMP_T>(sat) };
}
```

That should work just fine, regardless of the size of the types.

Capital-Asterisk: Oh, this is an optimization by EnTT. `emplace` will return void if the type is an empty struct.

https://github.com/skypjack/entt/wiki/Crash-Course:-entity-component-system#empty-type-optimization

src/osp/Universe.h
Capital-Asterisk: It use to be a "CRTP to implement common functions of TypeSatellites," but now it's only used to emplace multiple components through the add_get_ucomp. The functions only default-construct, and adding a big initializer list would just make it into a more confusing way to call emplace multiple times. In this case, I think this class should go too :hocho:

src/osp/Universe.h
jonesmz: Ahhhhhhhhh, that makes sense.

src/osp/Universe.h
jonesmz: Good to know.

But in this case, UCOMP_T represents several UCOMP_T template parameters, which is why i recommended UCOMPS_T

Capital-Asterisk:
The new changes gets rid of the confusing CommonTypeSat class. Its main purpose was to emplace multiple components into the universe using the `add_get_ucomp`. This is just confusing and can only default construct. It would be better to just call `emplace` explicitly.

New functions were added to SatActiveArea, SatVehicle, and SatPlanet to add their respective components (with useful arguments too).

jonesmz (approved):

PR #51, opened 2021-01-14
Create input configuration by file

EhWhoAmI:
This makes it possible to modify the input using a config file, using `settings.toml`.

Issues with it:
 - Fails if you do not include all the inputs
 - Still needs a few keys, such as function keys
 - Does not function without `settings.toml`
 - `settings.toml` is not copied by default after build


src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+//Map for all the keys
+//The tuple is in this order: device, number, and hold/pressed
+using namespace osp;
+
+using Key_t = OSPMagnum::KeyEvent::Key;
+using Mouse_t = OSPMagnum::MouseEvent::Button;
+using VarOp_t = ButtonVarConfig::VarOperator;
+using VarTrig_t = ButtonVarConfig::VarTrigger;
+
+typedef std::tuple<int, int> button_tuple;
```
Capital-Asterisk, line 216: ButtonTuple_t, prefer using instead of typedef, and also use UserInputHandler::DeviceId (which should also be DeviceId_t) instead of int for the first one.

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+	auto start = 0U;
+	auto end = s.find(delim);
+
+
+	if (s == "None") {
+		return handlers;
+	}
+	if (end == std::string::npos) {
+		try {
+			button_tuple button
```
Capital-Asterisk, line 317: A good time to use structured binding
```cpp
auto const &[device, button] = buttonMap.at(s);
```

Some comments for this area would be good too.
```
// If no "+" is found
```

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+//The tuple is in this order: device, number, and hold/pressed
+using namespace osp;
+
+using Key_t = OSPMagnum::KeyEvent::Key;
+using Mouse_t = OSPMagnum::MouseEvent::Button;
+using VarOp_t = ButtonVarConfig::VarOperator;
+using VarTrig_t = ButtonVarConfig::VarTrigger;
+
+typedef std::tuple<int, int> button_tuple;
+const std::map<std::string, button_tuple> buttonMap = {
```
Capital-Asterisk, line 217: static constant is prefixed with sc_ `sc_buttonMap`.

Also a good idea to use a 'transparent comparator' by adding std::less<>, so this can be accessed using string view:
```cpp
std::map<std::string, button_tuple, std::less<> >
```

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+	//Mouse
+	{"RMouse", {sc_mouse, (int)OSPMagnum::MouseEvent::Button::Right }},
+	{"LMouse", {sc_mouse, (int)OSPMagnum::MouseEvent::Button::Left }},
+	{"MMouse", {sc_mouse, (int)OSPMagnum::MouseEvent::Button::Middle }}
+};
+
+std::vector<ButtonVarConfig> parse_control(std::string_view str) {
+	//Split string
+	std::string s(str);
```
Capital-Asterisk, line 303: Why use a string view when it's just copied into a string? String views work best if all you're doing is substr, which seems to be the case here.

bin/settings.toml
```diff
@@ -0,0 +1,104 @@
...
+
+[vehicle_yaw_lf]
+primary = "A"
+secondary = "None"
+invert = true
+
+[vehicle_yaw_rt]
+primary = "D"
+secondary = "None"
+invert = true
```
jonesmz, line 44: what is "invert" ?

bin/settings.toml
```diff
@@ -0,0 +1,104 @@
...
+
+[ui_rmb]
+primary = "RMouse"
+secondary = "None"
+invert = true
+
+[debug_planet_update]
+primary = "LCtrl+1"
+secondary = "None"
+invert = true
```
jonesmz, line 104: please always leave one newline at the end of a file.

src/osp/UserInputHandler.cpp
```diff
@@ -245,8 +245,15 @@ void UserInputHandler::config_register_control(std::string const& name,
         std::initializer_list<ButtonVarConfig> vars)
 {
     std::vector<ButtonVarConfig> varVector(vars);
-    m_controlConfigs[name] = ButtonConfig{std::move(varVector), holdable,
-                                          false, 0};
+    config_register_control(name, holdable, varVector);
```
jonesmz, line 248: std::move your vector

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+			vec.insert(vec.end(), secondaryKeys.begin(), secondaryKeys.end());
+
+			bool invert = toml::find(v, "invert").as_boolean();
+
+			//Add user input
+			rUserInput.config_register_control(k, invert, vec);
+		}
+	}
+	catch (std::out_of_range oor) {
```
jonesmz, line 201: newline before brace

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+using Key_t = OSPMagnum::KeyEvent::Key;
+using Mouse_t = OSPMagnum::MouseEvent::Button;
+using VarOp_t = ButtonVarConfig::VarOperator;
+using VarTrig_t = ButtonVarConfig::VarTrigger;
+
+typedef std::tuple<int, int> button_tuple;
+const std::map<std::string, button_tuple> buttonMap = {
+	//Keyboard
+	{"LCtrl", {sc_keyboard, (int)OSPMagnum::KeyEvent::Key::LeftCtrl}},
```
jonesmz, line 219: i feel like the whole point of the EnTT system is that it gets used to store data like this?

@Capital-Asterisk ?

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+
+	auto start = 0U;
+	auto end = s.find(delim);
+
+
+	if (s == "None") {
+		return handlers;
+	}
+	if (end == std::string::npos) {
```
jonesmz, line 315: you're comparing std::string_view::npos for std::string:npos...

src/test_application/OSPMagnum.cpp
jonesmz, line 303: Uhhh, don't copy your string view into a string...

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+std::vector<ButtonVarConfig> parse_control(std::string_view str) {
+	//Split string
+	std::string s(str);
+	std::string delim = "+";
+	std::vector<ButtonVarConfig> handlers;
+
+
+	auto start = 0U;
+	auto end = s.find(delim);
```
Capital-Asterisk, line 309: I'm a bit confused over these two: avoid auto. I thought `end` here was an iterator. Why is it named "end" when it's the location of a '+'? `start` can be moved down to the else block where it is actually used.

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+	using namespace osp;
+
+	using Key_t = OSPMagnum::KeyEvent::Key;
+	using Mouse_t = OSPMagnum::MouseEvent::Button;
+	using VarOp_t = ButtonVarConfig::VarOperator;
+	using VarTrig_t = ButtonVarConfig::VarTrigger;
+
+	UserInputHandler& rUserInput = rOspApp.get_input_handler();
+	try {
```
Capital-Asterisk, line 184: This is the first time I've seen exceptions in this project. I'd need an expert opinion on this.

src/test_application/OSPMagnum.cpp
```diff
@@ -159,71 +161,195 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+		for (const auto& [k, v] : data.as_table()) {
+			std::cout << k << std::endl;
+			std::string primary = toml::find(v, "primary").as_string();
```
Capital-Asterisk, line 187: Learning Rust and jonesmz's comments makes every unnecessary string allocation look evil to me. There has to be a way to not copy the string here.

src/test_application/OSPMagnum.cpp
Capital-Asterisk: This maps string identifiers to a device and button, as far as I know this is a proper use of a map.

bin/settings.toml
EhWhoAmI: It's supposed to be holdable, I made a mistake while reading the code

jonesmz:
@EhWhoAmI You'll need to resolve the conflicts before we'll be able to merge it.

EhWhoAmI:
@jonesmz, yup, fixed those

Capital-Asterisk (approved):

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+                = buttonMap.at(sub);
+            handlers.push_back(osp::ButtonVarConfig(device, button, VarTrig_t::PRESSED, false, VarOp_t::OR));
+        }
+        catch (std::out_of_range oor) {
+            std::cout << oor.what() << sub << std::endl;
+        }
+    }
+
+    return handlers;
+}
```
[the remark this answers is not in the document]
jonesmz: Various old style Unix tools get annoyed and make noise about it.

Don't anger the old gods basically

src/osp/UserInputHandler.cpp
```diff
@@ -245,8 +245,15 @@ void UserInputHandler::config_register_control(std::string const& name,
...
-    m_controlConfigs[name] = ButtonConfig{std::move(varVector), holdable,
-                                          false, 0};
+    config_register_control(name, holdable, std::move(varVector));
+}
+
+void UserInputHandler::config_register_control(std::string const& name,
+    bool holdable,
+    std::vector<ButtonVarConfig> vars)
+{
+    m_controlConfigs[name] = ButtonConfig{ std::move(vars), holdable,
```
jonesmz, line 255: use std::map::insert_or_assign for this.

src/osp/UserInputHandler.cpp
```diff
@@ -257,7 +264,7 @@ ButtonControlHandle UserInputHandler::config_get(std::string const& name)
 
     if (cfgIt == m_controlConfigs.end())
     {
-        // Config not found!
+        // Config not found, crashes after this, so needs to be fixed
```
jonesmz, line 267: If it's going to crash if it's not found, then just explicitly crash the program, don't rely on it crashing on it's own.

do something like `throw std::exception("some explaination");`

src/osp/UserInputHandler.cpp
jonesmz: and you should also make two instances of the function.

one that takes the `name` parameter by const&, and the other by &&. That way we can get the speed gains of move semantics when the key is an rvalue reference.

Note that i say to do this way instead of taking the parameter by value, specifically because the key may not actually be moved from if the item already exists in the map, so we don't want to pay the cost of making a copy of the string if we don't need to.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+            controls.insert(controls.end(), secondaryKeys.begin(), secondaryKeys.end());
+
+            bool holdable = toml::find(v, "holdable").as_boolean();
+
+            //Add user input
+            rUserInput.config_register_control(k, holdable, controls);
+
+        }
+    }
+    catch (std::out_of_range oor) {
```
jonesmz, line 187: don't catch std::out_of_range.

Catch std::exception.

Also, never catch an exception by value. If you catch an exception by value, you're causing object slicing.

Always catch by const&.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+//The tuple is in this order: device, number, and hold/pressed
+using Key_t = OSPMagnum::KeyEvent::Key;
+using Mouse_t = OSPMagnum::MouseEvent::Button;
+using VarOp_t = osp::ButtonVarConfig::VarOperator;
+using VarTrig_t = osp::ButtonVarConfig::VarTrigger;
+
+typedef std::tuple<int, int> button_tuple;
+const std::map<std::string_view, button_tuple, std::less<>> buttonMap = {
+    //Keyboard
+    {"LCtrl", {osp::sc_keyboard, (int)Key_t::LeftCtrl}},
```
jonesmz, line 203: I'm sad that we have to have a manually specified list of strings like this.

but I can't think of any way to avoid it.

Can you?

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+    {"RCtrl", {osp::sc_keyboard, (int)Key_t::RightCtrl }},
+    {"LShift", {osp::sc_keyboard, (int)Key_t::LeftShift }},
+    {"RShift", {osp::sc_keyboard, (int)Key_t::RightShift }},
+    {"LAlt", {osp::sc_keyboard, (int)Key_t::LeftAlt }},
+    {"RAlt", {osp::sc_keyboard, (int)Key_t::RightAlt }},
+    {"Up", {osp::sc_keyboard, (int)Key_t::Up }},
+    {"Down", {osp::sc_keyboard, (int)Key_t::Down }},
+    {"Left", {osp::sc_keyboard, (int)Key_t::Left }},
```
jonesmz, line 211: I don't think that these should be getting cast to int.

Why cant you make the button tuple take the key type directly?

src/test_application/OSPMagnum.cpp
z-adams: Nope, I think it's best to just suck it up in this case. It's ugly but doing it programmatically means circumventing Magnum's abstraction layer which is a bad idea imo.

src/test_application/OSPMagnum.cpp
jonesmz: Is it because of the mouse keys?

Makes me wonder about two possibilities.

1) Should the keyboard and mouse have their own maps, instead of a shared map?
2) Instead of explicitly saying osp::sc_keyboard vs osp::sc_mouse, you could instead use an std::variant, and then determine if it's a keyboard or mouse mapping based on the item held by the variant.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+    //Mouse
+    {"RMouse", {osp::sc_mouse, (int)OSPMagnum::MouseEvent::Button::Right }},
+    {"LMouse", {osp::sc_mouse, (int)OSPMagnum::MouseEvent::Button::Left }},
+    {"MMouse", {osp::sc_mouse, (int)OSPMagnum::MouseEvent::Button::Middle }}
+};
+
+std::vector<osp::ButtonVarConfig> parse_control(std::string_view str) {
+    //Split by +
+    constexpr std::string_view delim = "+";
```
jonesmz, line 287: static constexpr.

jonesmz: That being said, what's the main advantage of making this it's own variable, if it's only used in one place?

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+    if (str == "None") {
+        return handlers;
+    }
+
+    auto start = 0U;
+    auto end = str.find(delim);
+
+    //If no '+' is found
+    if (end == std::string_view::npos) {
+        try {
```
jonesmz, line 300: Consistant bracing please.

Personally I prefer braces to always go on the next line.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+    //If none, then no actions
+    if (str == "None") {
+        return handlers;
+    }
+
+    auto start = 0U;
+    auto end = str.find(delim);
+
+    //If no '+' is found
+    if (end == std::string_view::npos) {
```
jonesmz, line 299: use c++17's if-initializer sytax.

src/test_application/OSPMagnum.cpp
z-adams: Next-line braces is how the rest of the repo is formatted so let's stick with that

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+        try {
+            auto const& [device, button]
+                = buttonMap.at(str);
+
+            handlers.push_back(osp::ButtonVarConfig(device, button, VarTrig_t::PRESSED, false, VarOp_t::OR));
```
jonesmz, line 304: Can't this be a call to emplace_back, instead of push_back?

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+    }
+
+    auto start = 0U;
+    auto end = str.find(delim);
+
+    //If no '+' is found
+    if (end == std::string_view::npos) {
+        try {
+            auto const& [device, button]
+                = buttonMap.at(str);
```
jonesmz, line 302: Why are you assuming that the item with key "str" will be found?

You should use .find(), and check that you aren't getting an iterator pointing to std::map::end before dereferrencing.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,178 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+            handlers.push_back(osp::ButtonVarConfig(device, button, VarTrig_t::PRESSED, false, VarOp_t::OR));
+        }
+        catch (std::out_of_range oor) {
+            std::cout << oor.what() << str << std::endl;
+        }
+    }
+    else {
+        while (end != std::string_view::npos)
```
jonesmz, line 311: I think it's weird that you've got this while loop in the else-condition.

It seems to me, based on trying to understand the code, that you could have only the while loop, without needing to have the if/else branch.

src/test_application/OSPMagnum.h
```diff
@@ -91,3 +93,5 @@ class OSPMagnum : public Magnum::Platform::Application
 void config_controls(OSPMagnum& rOspApp);
 
 }
+
+std::vector<osp::ButtonVarConfig> parse_control(std::string_view str);
```
jonesmz, line 97: I recommend putting some doxygen style documentation in the header files for functions declarations.

jonesmz: Additionally, it looks like this functions isn't used by this file.

Would it make sense for it to have it's own header file instead?

src/test_application/OSPMagnum.cpp
jonesmz: Noted.

jonesmz:
I recommend one of:

- Only one commit.

Or

- Make a series of commits

1) Add the toml lib and nothing else.
2) Add the toml parsing code, but don't replace the existing config code
3) Add the toml file
4) Replace the existing config code with the parsing of the toml file

EhWhoAmI:
I'll do one commit, because I have already squashed the commits XD

jonesmz:
> I'll do one commit, because I have already squashed the commits XD

Sure, that's fine.

src/test_application/OSPMagnum.cpp
EhWhoAmI: You mean like this: `if (auto end = str.find(delim); end == std::string_view::npos) {`?

src/test_application/OSPMagnum.cpp
EhWhoAmI: `int` is also how `osp::ButtonVarConfig` stores the button, so I was just following what was already done.

src/test_application/OSPMagnum.cpp
EhWhoAmI: I did a programmatic method before this, but apparently not all devices map their button keys to ASCII, so this was the best  method. I didn't want to do this either.

src/test_application/OSPMagnum.cpp
z-adams: Yep

jonesmz: don't forget liberal use of `const`

src/test_application/OSPMagnum.cpp
jonesmz: Sure.

Maybe @Capital-Asterisk or @z-adams can chime in on it.

I just don't see why casting to int is the right approach. Maybe there's a good reason.

src/test_application/OSPMagnum.cpp
EhWhoAmI: It's used in 2 places, and it's clearer this way.

src/osp/UserInputHandler.cpp
```diff
@@ -252,8 +252,14 @@ void UserInputHandler::config_register_control(std::string const& name,
     bool holdable,
     std::vector<ButtonVarConfig> vars)
 {
-    m_controlConfigs[name] = ButtonConfig{ std::move(vars), holdable,
-                                          false, 0 };
+    m_controlConfigs.insert_or_assign(name, ButtonConfig{ std::move(vars), holdable, false, 0 });
+}
+
+void UserInputHandler::config_register_control(std::string const&& name,
```
jonesmz, line 258: There is no such thing as const&&.

jonesmz: Basically I was wanting you to make two functions.
```
void UserInputHandler::config_register_control(std::string const& name, bool holdable, std::vector<ButtonVarConfig> vars)
{
    m_controlConfigs.insert_or_assign(name, ButtonConfig{ std::move(vars), holdable, false, 0 });
}
void UserInputHandler::config_register_control(std::string && name, bool holdable, std::vector<ButtonVarConfig> vars)
{
    m_controlConfigs.insert_or_assign(std::move(name), ButtonConfig{ std::move(vars), holdable, false, 0 });
}
```

The reason for this is that insert_or_assign does not move from the first parameter if the item in question already exists, so we don't want to take the parameter by value.
Instead we want to take the parameter by const-& OR &&, so that the caller of the function gets to decide whether to use move or not.

src/osp/UserInputHandler.cpp
```diff
@@ -264,8 +270,9 @@ ButtonControlHandle UserInputHandler::config_get(std::string const& name)
 
     if (cfgIt == m_controlConfigs.end())
     {
-        // Config not found, crashes after this, so needs to be fixed
-        std::cout << "config not found!\n";
+        // Config not found, no way to have an empty key so far, so throw an exception
+        std::cout << "No config for " << name << std::endl;
+        throw std::runtime_error("Error: no config with " + name);
         return ButtonControlHandle(nullptr, 0);
```
jonesmz, line 276: You don't need a return statement after a throw statement. The return statement will never be reached.

src/osp/UserInputHandler.h
```diff
@@ -298,6 +298,10 @@ class UserInputHandler
     void config_register_control(std::string const& name, bool holdable,
             std::initializer_list<ButtonVarConfig> vars);
```
jonesmz, line 299: You do not need a version of this function that takes an initializer list.

The version that takes the std::vector is sufficent.

src/osp/UserInputHandler.h
```diff
@@ -298,6 +298,10 @@ class UserInputHandler
...
 
+    void config_register_control(std::string const&& name,
+        bool holdable,
```
jonesmz, line 302: But remember, you need const& and &&, for the name parameter, as two seperate functions.

src/osp/UserInputHandler.h
```diff
@@ -342,7 +346,7 @@ class UserInputHandler
     */
     void scroll_delta(Vector2i offset);
 
-private:
+//private:
```
jonesmz, line 349: Don't comment things out like this. Either delete the line entirely, or fix the reason why you needed to unhide these variables.

src/test_application/OSPMagnum.cpp
```diff
@@ -181,11 +182,11 @@ void testapp::config_controls(OSPMagnum& rOspApp)
...
             //Add user input
             rUserInput.config_register_control(k, holdable, controls);
-
         }
     }
-    catch (std::out_of_range oor) {
-        std::cout << oor.what() << std::endl;
+    catch (const std::exception& exception) 
+    {
+        std::cout << exception.what() << std::endl;
```
jonesmz, line 189: I'm not sure I agree with just catching the exception and doing nothing. The program is probably in a bad state if an exc eption gets to this level.

Just let the exception go to the top level and crash the program

src/osp/UserInputHandler.cpp
```diff
@@ -245,8 +245,21 @@ void UserInputHandler::config_register_control(std::string const& name,
         std::initializer_list<ButtonVarConfig> vars)
```
jonesmz, line 245: As I said in the .h file, there's no need to have a version of this function that takes an initializer list. It always gets converted to std::vector, so just keep that version(s) of the function.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,165 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
     // Configure Controls
+    //Load toml
+    auto data = toml::parse("settings.toml");
+    try 
+    {
+        osp::UserInputHandler& rUserInput = rOspApp.get_input_handler();
+        for (const auto& [k, v] : data.as_table())
+        {
+            std::cout << "Parsing " << k << std::endl;
+            std::string primary = toml::find(v, "primary").as_string();
```
jonesmz, line 173: If you're not going to modify, or move, the `std::string primary` then bind the temporary returned by toml::find to a const& variable, like this:
```
std::string const& primary = toml::find();.
```
The reason to get into the habit of doing this is that you don't typically know (sometimes you do, most times you dont) whether the function you are calling is returning a std::string by *value*, or is instead returning a const& to a std::string that lives somewhere else.

So if you aren't intending to modify / move the string, you don't want to make a copy of the string if the return type is const&, you just want the const&.

If the function returns by value, then the temporary value that is returned can either be stored as a real variable (what you're doing now) or it can be stored as a "temporary with lifetime extension", and then your const& points to that temporary, which is what I'm recommending you do.

jonesmz: In the generic sense, there are a few catagories of how to accept return values. But I'll give you the very simplified version.

1) You don't modify or move from the variable. It's "const" as far as the current code cares. Accpet the variable as const&.
2) You *do* modify or move the variable. Accept the variable by value, without const&.

src/osp/UserInputHandler.h
EhWhoAmI: Are they not?

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,165 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+
+            std::string secondary = toml::find(v, "secondary").as_string();
+            std::vector<osp::ButtonVarConfig> secondaryKeys = parse_control(secondary);
+
+            controls.insert(controls.end(), secondaryKeys.begin(), secondaryKeys.end());
+
+            bool holdable = toml::find(v, "holdable").as_boolean();
+
+            //Add user input
+            rUserInput.config_register_control(k, holdable, controls);
```
jonesmz, line 184: This is the last place where you're referencing "controls". 
As such you'll need to make sure you use std::move() on it to tell the compiler you're done with it.
Sadly, it can't detect that itself.
See here for a bit more info: https://stackoverflow.com/questions/45763943/compiler-deduction-of-rvalue-references-for-variables-going-out-of-scope

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,165 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+        for (const auto& [k, v] : data.as_table())
+        {
+            std::cout << "Parsing " << k << std::endl;
+            std::string primary = toml::find(v, "primary").as_string();
+            std::vector<osp::ButtonVarConfig> controls = parse_control(primary);
+
+            std::string secondary = toml::find(v, "secondary").as_string();
+            std::vector<osp::ButtonVarConfig> secondaryKeys = parse_control(secondary);
+
+            controls.insert(controls.end(), secondaryKeys.begin(), secondaryKeys.end());
```
jonesmz, line 179: Here you're failing to use std::move on your osp::ButtonVarConfig objects.

Use
```
controls.insert(controls.end(), std::make_move_iterator(secondaryKeys.begin()), std::make_move_iterator(secondaryKeys.end());
```
to tell the controls variable that it should move from the secondaryKeys array, instead of copy.

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,165 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+    static constexpr std::string_view delim = "+";
+    std::cout << str << std::endl;
+
+    auto start = 0U;
+    auto end = str.find(delim);
+    while (end != std::string::npos)
+    {
+        std::string_view sub = str.substr(start, end - start);
+        auto const it
+            = buttonMap.find(sub);
```
jonesmz, line 304: no need for a line-break here.

Also, instead of `auto const it` use `auto const& it`

src/test_application/OSPMagnum.cpp
```diff
@@ -160,76 +162,165 @@ osp::active::ActiveScene& OSPMagnum::scene_create(std::string && name)
...
+            std::cout << sub << " " << device << " " << button << std::endl;
+
+        }
+        start = end + delim.length();
+        end = str.find(delim, start);
+    }
+
+    std::string_view sub = str.substr(start, end);
+    auto const it
+        = buttonMap.find(sub);
```
jonesmz, line 318: no need for a line-break here.

Also, instead of auto const it use auto const& it

jonesmz:
At this point, I think this PR is nearly ready to merge.

I have only a very small number of complaints, which I think you can rectify fairly easily.

As such, when you've fixed these last items, I'll approve the PR, and then we'll only need @z-adams to give a final approval.

src/osp/UserInputHandler.h
EhWhoAmI: That was for debugging something, I just forgot to uncomment it.

src/osp/UserInputHandler.h
jonesmz: Currently you have one function that takes const&&_ which isn't really a thing

jonesmz (approved):

PR #57, opened 2021-01-19
Procedural inertia

z-adams:
Adds machinery to procedurally compute moments of inertia. This is more difficult than procedurally computing center of mass, because a point-model of the vessel is insufficient to model it properly; the volume of parts needs to be considered for realistic physics. A decent way of doing this quickly (and in general, if some potential pitfalls are accounted for) is to approximate the mass distribution of parts by spreading their mass evenly over their collider volume, which is what this PR adds.

To compliment these changes, I've also added sturdy-defined engine thrusts. I know there is a more complicated system developed in the past, but this will suffice for now.

Changelog:

* Added realistic masses, engine thrusts to part sturdies [part files]
* Updated PrototypeMachine, which stores a map of config values (e.g. thrust, for an engine)
* AssetImporter reads in arbitrary machine config data; properly discriminates between collider types. Collider shapes don't yet work for actual collisions, but are used for inertia calculations
* SysNewton received a bunch of new functions for calculating inertia. `compute_body_inertia()` calculates the inertia for a whole rigidbody by calling `compute_part_inertia()` to compute the inertia of each part.
* New CommonPhysics files contain helper functions used to calculate the inertia tensors of collision shapes, and transform the results to compute composite inertias
* SysMachine::instantiate now accepts a `PrototypeMachine` and `BlueprintMachine` argument, which are used during instantiation to configure the machine settings/parameters [See SysMachine]
* SysVehicle, SysUserControl, SysMachineRocket updated to comply with the above, although only SysMachineRocket leverages it (the peak engine thrust is now a constructor parameter) [See  `SysMachineRocket::instantiate`]
* Added convenient `ActiveScene::reg_try_get()` equivalent to `reg_get()` so that we don't have to always call `rScene.get_registry().try_get<>()`

Since some of the physics stuff went into CommonPhysics and the new `osp::phys` namespace, there are a bunch of random changes scattered about; sorry about that.

The main changes take place in PrototypePart.cpp, AssetImporter.cpp, and SysNewton.h/.cpp; the rest of the changes are simply updating `instantiate()` functions, and changing Rocket.h/.cpp to now accept and store a thrust parameter.

### DIFF
--- a/src/osp/Active/ActiveScene.h
+++ b/src/osp/Active/ActiveScene.h
@@ -133,6 +133,14 @@ public:
 
 
     /**
+     * Shorthand for get_registry().try_get<T>()
+     * @tparam T Component to get
+     * @return Pointer to component
+     */
+    template<class T>
+    constexpr T* reg_try_get(ActiveEnt ent) { return m_registry.try_get<T>(ent); };
+
+    /**
      * Shorthand for get_registry().emplace<T>()
      * @tparam T Component to emplace
      */
@@ -366,28 +374,32 @@ void ActiveScene::hierarchy_traverse(ActiveEnt root, FUNC_T callable)
     std::stack<ActiveEnt> parentNextSibling;
     ActiveEnt currentEnt = root;
 
+    unsigned int rootLevel = reg_get<ACompHierarchy>(root).m_level;
+
     while (true)
     {
         ACompHierarchy &hier = reg_get<ACompHierarchy>(currentEnt);
 
         if (callable(currentEnt) == EHierarchyTraverseStatus::Stop) { return; }
 
-        if (hier.m_childCount)
+        if (hier.m_childCount > 0)
         {
             // entity has some children
             currentEnt = hier.m_childFirst;
 
-
-            // save next sibling for later if it exists
-            if (hier.m_siblingNext != entt::null)
+            // Save next sibling for later if it exists
+            // Don't check siblings of the root node
+            if ((hier.m_siblingNext != entt::null) && (hier.m_level > rootLevel))
             {
                 parentNextSibling.push(hier.m_siblingNext);
             }
-        } else if (hier.m_siblingNext != entt::null)
+        }
+        else if ((hier.m_siblingNext != entt::null) && (hier.m_level > rootLevel))
         {
             // no children, move to next sibling
             currentEnt = hier.m_siblingNext;
-        } else if (parentNextSibling.size())
+        }
+        else if (parentNextSibling.size() > 0)
         {
             // last sibling, and not done yet
             // is last sibling, move to parent's (or ancestor's) next sibling
--- a/src/osp/Active/SysNewton.cpp
+++ b/src/osp/Active/SysNewton.cpp
@@ -37,6 +37,7 @@ using osp::active::ACompNwtBody;
 using osp::active::ACompNwtWorld;
 using osp::active::ACompTransform;
 
+using namespace osp;
 using Magnum::Vector4;
 using Magnum::Matrix4;
 
@@ -243,7 +244,6 @@ void SysNewton::find_colliders_recurse(ActiveScene& rScene, ActiveEnt ent,
 void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
                             NewtonWorld const* nwtWorld)
 {
-
     auto const& entHier = rScene.reg_get<ACompHierarchy>(entity);
     auto& entTransform = rScene.reg_get<ACompTransform>(entity);
 
@@ -258,14 +258,13 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
 
     switch (entShape->m_shape)
     {
-    case ECollisionShape::COMBINED:
+    case phys::ECollisionShape::COMBINED:
     {
         // Combine collision shapes from descendants
 
         NewtonCollision* rCompound
                 = NewtonCreateCompoundCollision(nwtWorld, 0);
 
-        //
         NewtonCompoundCollisionBeginAddRemove(rCompound);
         find_colliders_recurse(rScene, entHier.m_childFirst, Matrix4(),
                                nwtWorld, rCompound);
@@ -274,11 +273,10 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
         // Compute mass, center of mass
         Vector4 centerOfMass = compute_body_CoM(rScene, entHier.m_childFirst, Matrix4{});
         entBody.m_centerOfMassOffset = centerOfMass.xyz();
-        /* TODO: mass is left as the default (1.0f) for now; when we have the
-         * logic necesary to represent different thrusts per engine we can
-         * enable this; for now, real masses are too large.
-         */
-        //entBody.m_mass = centerOfMass.w();
+
+        // Compute moments of inertia
+        Matrix3 inertia = compute_body_inertia(rScene, entHier.m_childFirst,
+            centerOfMass.xyz(), Matrix4{});
 
         // TODO: deal with mass and stuff
 
@@ -295,15 +293,20 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
         NewtonDestroyCollision(rCompound);
 
         // Set inertia and mass
+        entBody.m_mass = centerOfMass.w();
+        entBody.m_inertia.x() = inertia[0][0];  // Ixx
+        entBody.m_inertia.y() = inertia[1][1];  // Iyy
+        entBody.m_inertia.z() = inertia[2][2];  // Izz
+
         NewtonBodySetMassMatrix(entBody.m_body, entBody.m_mass,
-                                entBody.m_intertia.x(),
-                                entBody.m_intertia.x(),
-                                entBody.m_intertia.z());
+                                entBody.m_inertia.x(),
+                                entBody.m_inertia.y(),
+                                entBody.m_inertia.z());
         NewtonBodySetCentreOfMass(entBody.m_body, centerOfMass.xyz().data());
 
         break;
     }
-    case ECollisionShape::TERRAIN:
+    case phys::ECollisionShape::TERRAIN:
     {
         // Get NewtonTreeCollision generated from elsewhere
         // such as SysPlanetA::debug_create_chunk_collider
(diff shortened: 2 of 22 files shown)


src/osp/Active/ActiveScene.h
```diff
@@ -366,28 +374,32 @@ void ActiveScene::hierarchy_traverse(ActiveEnt root, FUNC_T callable)
...
         {
             // entity has some children
             currentEnt = hier.m_childFirst;
 
-
-            // save next sibling for later if it exists
-            if (hier.m_siblingNext != entt::null)
+            // Save next sibling for later if it exists
+            // Don't check siblings of the root node
+            if (hier.m_siblingNext != entt::null && hier.m_level > rootLevel)
```
jonesmz, line 392: Parens

src/osp/Active/CommonPhysics.cpp
```diff
@@ -0,0 +1,148 @@
...
+    case ECollisionShape::CYLINDER:
+        // Default radius: 1, default height: 2
+        return pi * scale.x() * scale.x() * 2.0f*scale.z();
+    case ECollisionShape::CAPSULE:
+    case ECollisionShape::CONVEX_HULL:
+    case ECollisionShape::TERRAIN:
+    case ECollisionShape::COMBINED:
+    default:
+        std::cout << "Error: unsupported shape for volume calc\n";
+        return 0.0f;
```
jonesmz, line 55: Assert

src/osp/Active/CommonPhysics.cpp
```diff
@@ -0,0 +1,148 @@
...
+    // Apply rotation via similarity transformation
+    I = rotation.transposed() * I * rotation;
+
+    // Translate via tensor generalized parallel axis theorem
+    using Magnum::Math::dot;
+    Vector3 r = translation;
+    Matrix3 outerProductR = {r * r.x(), r * r.y(), r * r.z()};
+    Matrix3 E3 = Matrix3{};
+
+    return I + mass * (dot(r, r) * E3 - outerProductR);
```
jonesmz, line 70: Weird parents?

src/osp/Active/CommonPhysics.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+Vector3 collider_inertia_tensor(ECollisionShape shape, Vector3 scale, float mass)
+{
+    switch (shape)
+    {
+    case ECollisionShape::CYLINDER:
+    {
+        // Default cylinder dimensions: radius 1, height 2
+        float height = 2.0f * scale.z();
+        // If sclY != sclX I will be mad
```
jonesmz, line 81: Const

src/osp/Active/CommonPhysics.cpp
```diff
@@ -0,0 +1,148 @@
...
+    float xx = y2 + z2;
+    float yy = x2 + z2;
+    float zz = x2 + y2;
+
+    // Pre-multiply constant to avoid non-constexpr operator*(float, Vector3)
+    float c = (1.0f / 12.0f) * mass;
+    return Vector3{c * xx, c * yy, c * zz};
+}
+
+constexpr Vector3 ellipsoid_inertia_tensor(const Vector3 semiaxes, float mass)
```
jonesmz, line 133: Constexpr functions in CPP file?

src/osp/Active/CommonPhysics.h
```diff
@@ -0,0 +1,135 @@
...
+{
+    NONE,
+    COMBINED,
+    SPHERE,
+    BOX,
+    CAPSULE,
+    CYLINDER,
+    //MESH,
+    CONVEX_HULL,
+    TERRAIN
```
jonesmz, line 43: Give these concrete values to help with debugging in log files

src/osp/Active/CommonPhysics.h
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Compute the inertia tensor for a cylinder
+ *
+ * Computes the moment of inertia about the principal axes of a cylinder
+ * with specified mass, height, and radius, whose axis of symmetry lies
+ * along the z-axis
+ *
+ * @return The moment of inertia about the 3 principal axes (x, y, z)
+ */
+constexpr Vector3 cylinder_inertia_tensor(float radius, float height, float mass);
```
jonesmz, line 107: Magnum doesn't have any of these?

Might be worth a PR to them.

src/osp/Active/SysNewton.cpp
```diff
@@ -274,11 +274,10 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
...
         entBody.m_centerOfMassOffset = centerOfMass.xyz();
-        /* TODO: mass is left as the default (1.0f) for now; when we have the
-         * logic necesary to represent different thrusts per engine we can
-         * enable this; for now, real masses are too large.
-         */
-        //entBody.m_mass = centerOfMass.w();
+
+        // Compute moments of inertia
+        Matrix3 inertia = compute_body_inertia(rScene, entHier.m_childFirst,
+            centerOfMass.xyz(), Matrix4{});
```
jonesmz, line 280: Default constructed Matrix4 instead of 0.0f?

src/osp/Resource/AssetImporter.cpp
```diff
@@ -121,9 +121,43 @@ void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
             continue;
         }
 
-        // TODO: more stuff
         PrototypeMachine machine;
         machine.m_type = std::move(type);
+
+        for (std::string key : value.Keys())
```
jonesmz, line 127: Iterating the string by value? You sure?

src/osp/Resource/AssetImporter.cpp
```diff
@@ -121,9 +121,43 @@ void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
...
+            {
+                int val = v.Get<int>();
+                machine.m_config.emplace(key, val);
+                std::cout << val << " (int)\n";
+                break;
+            }
+            case tinygltf::Type::STRING_TYPE:
+            {
+                std::string val = v.Get<std::string>();
+                machine.m_config.emplace(key, val);
```
jonesmz, line 151: std::move all over this function

src/osp/Resource/AssetImporter.cpp
```diff
@@ -383,9 +417,23 @@ void AssetImporter::proto_add_obj_recurse(TinyGltfImporter& gltfImporter,
...
+
+        std::string const& shapeName = extras.Get("shape").Get<std::string>();
+        phys::ECollisionShape shape;
+        if (shapeName == "cylinder")
+        {
+            shape = phys::ECollisionShape::CYLINDER;
+        }
+        else
+        {
+            shape = phys::ECollisionShape::BOX;
```
jonesmz, line 432: Trinary if. And make const.

jonesmz (approved):

src/osp/Active/CommonPhysics.cpp
z-adams: Hmm I may have committed something I didn't mean to, I'll check it out tomorrow

src/osp/Active/SysNewton.cpp
z-adams: Default constructed Matrix4 should be an identity matrix, whereas {0.0f} will likely be all zeroes

src/osp/Active/SysNewton.h
```diff
@@ -236,6 +219,50 @@ class SysNewton : public IDynamicSystem
...
+     * Traverses the immediate children of the specified entity and sums the
+     * volumes of any detected collision volumes. Cannot account for overlapping
+     * collider volumes.
+     * 
+     * @param rScene [in] ActiveScene containing relevant scene data
+     * @param part   [in] The part
+     * 
+     * @return The part's collider volume
+     */
+    static float compute_part_volume(ActiveScene& rScene, ActiveEnt part);
```
Capital-Asterisk, line 234: I'm a bit confused by what "part" refers to. It's clearly not a vehicle part; should this be rigid body or just body?

src/osp/Active/CommonPhysics.cpp
z-adams: The parentheses? They're as intended

src/osp/Active/CommonPhysics.h
z-adams: Magnum doesn't really concern itself with physics stuff, do you mean Newton? It may, but it probably won't for all the shapes we need, and I just prefer to do it myself.

src/osp/Resource/AssetImporter.cpp
z-adams: No I'm not :) I wrote this long before I knew what `std::string_view` was.

Capital-Asterisk (approved):
Putting part configs in the .sturdy.gltf sounded like a good idea at first, but considering things like difficulty settings, language localization, and upgrades, it would be better to put part configs in separate files.

It sounds like we're approaching some kind of patch-based system for machine configs. The configs that BlueprintMachine store will patch over the configs of the PrototypeMachine. It might make sense to make some functions that easily combines these configs.

src/osp/Active/SysNewton.h
z-adams: Any child of a rigidbody, which has collider children. The rigidbody function is separate, although it may be possible to combine them

src/osp/Active/SysNewton.cpp
```diff
@@ -295,15 +294,20 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
...
+        entBody.m_inertia.x() = inertia[0][0];  // Ixx
+        entBody.m_inertia.y() = inertia[1][1];  // Iyy
+        entBody.m_inertia.z() = inertia[2][2];  // Izz
+
         NewtonBodySetMassMatrix(entBody.m_body, entBody.m_mass,
-                                entBody.m_intertia.x(),
-                                entBody.m_intertia.x(),
-                                entBody.m_intertia.z());
+                                entBody.m_inertia.x(),
+                                entBody.m_inertia.x(),
```
Capital-Asterisk, line 304: oops

src/osp/Active/CommonPhysics.cpp
jonesmz: Ah, so you're relying on the operator precedence rules. Ok.

src/osp/Active/CommonPhysics.h
jonesmz: So is this a Magnum::Vector3 ?

Or is this a Newton::Vector3 ?

Maybe not all of it is something Magnum would care about, but i suspect at least some of these operations would be available, or something that Magnum would upstream from us.

src/osp/Active/SysNewton.cpp
jonesmz: Hrmm, alright.

src/osp/Resource/AssetImporter.cpp
jonesmz: I'd say just do

for(auto const& key : value.Keys())
{}

in this situation.

No need to type erase the string into a string_view.

src/osp/Active/CommonPhysics.cpp
z-adams: I will, although maybe we should discuss exceptions at some point, as they seem a bit more informational than just doing like `assert(false)` or something

src/osp/Active/CommonPhysics.h
z-adams: Magnum is generally physics-agnostic since it's just a graphics layer, it just has basic math types like vectors and matrices because they're relevant to graphics. We can try but I don't think it's relevant to its purpose.

src/osp/Resource/AssetImporter.cpp
z-adams: Where exactly? I changed that specific instance of moving the string, but I don't really see any other obvious places in this function

PR #58, opened 2021-01-21
Add ShipResources

z-adams:
"Adds a number of tools for storing and handling consumable on-ship
resources such as fuel, battery charge, etc. ShipResourceType provides a
definition for a specific type of resource and how it's stored
numerically (resource quantities will be represented by 64-bit unsigned
integers; see comments in source for detail). MachineContainer will
represent a container to store such resources."
---
None of this is implemented anywhere in the codebase yet, but I figured I'd split it into its own PR since the next one is going to be kind of large, and get some preliminary feedback. We discussed the underlying data representation a while back. This code will change a bit with the coming PR so if something looks out of place it will either be changed later or I can change it in the next PR.

### DIFF
--- a/src/adera/ShipResources.cpp
+++ b/src/adera/ShipResources.cpp
@@ -1,0 +1,119 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2021 Open Space Program Project
+ *
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+#include "ShipResources.h"
+#include "osp/Active/ActiveScene.h"
+
+using namespace osp;
+using namespace osp::active;
+using namespace adera::active::machines;
+
+/* MachineContainer */
+
+void MachineContainer::propagate_output(WireOutput* output)
+{
+
+}
+
+WireInput* MachineContainer::request_input(WireInPort port)
+{
+    return nullptr;
+}
+
+WireOutput* MachineContainer::request_output(WireOutPort port)
+{
+    return nullptr;
+}
+
+std::vector<WireInput*> MachineContainer::existing_inputs()
+{
+    return m_inputs;
+}
+
+std::vector<WireOutput*> MachineContainer::existing_outputs()
+{
+    return m_outputs;
+}
+
+/* SysMachineContainer */
+
+SysMachineContainer::SysMachineContainer(ActiveScene& rScene)
+    : SysMachine<SysMachineContainer, MachineContainer>(rScene)
+    , m_updateContainers(rScene.get_update_order(), "mach_container", "", "mach_rocket",
+        [this](ActiveScene& rScene) { this->update_containers(rScene); })
+{ }
+
+void SysMachineContainer::update_containers(ActiveScene& rScene)
+{
+    auto view = rScene.get_registry().view<MachineContainer>();
+
+    for (ActiveEnt ent : view)
+    {
+        auto& container = view.get<MachineContainer>(ent);
+
+        // Do something useful here... or not
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/adera/ShipResources.cpp
```diff
@@ -0,0 +1,119 @@
...
+    //std::string resName = std::get<std::string>(settings.m_config["resourcename"]);
+    std::string_view resName = "";
+    Path resPath = decompose_path(resName);
+    Package& pkg = m_scene.get_application().debug_find_package(resPath.prefix);
+
+    ShipResource resource;
+    resource.m_type = pkg.get<ShipResourceType>(resPath.identifier);
+    resource.m_quantity = resource_capacity(*resource.m_type, capacity);
+
+    return m_scene.reg_emplace<MachineContainer>(ent, capacity, resource);
```
z-adams, line 113: This function will change completely in the next PR when everything is hooked up; for now I just have it this way so that the file compiles.

src/adera/ShipResources.h
```diff
@@ -0,0 +1,158 @@
...
+    : public osp::active::SysMachine<SysMachineContainer, MachineContainer>
+{
+public:
+    SysMachineContainer(osp::active::ActiveScene& rScene);
+
+    static void update_containers(osp::active::ActiveScene& rScene);
+
+    static float resource_volume(ShipResourceType type, uint64_t quantity);
+    static float resource_mass(ShipResourceType type, uint64_t quantity);
+    static uint64_t resource_capacity(ShipResourceType type, float volume);
```
z-adams, line 96: In the next PR these become member functions of `struct ShipResource`, which is much cleaner (e.g. `fuel.compute_volume()` rather than `compute_volume(fuel.m_type, fuel.m_quantity)`), but I'm leaving this stuff as-is for now because it will make my life a lot easier when I have to cherry-pick my commits for the next PR.

jonesmz (approved):

jonesmz:
Short and sweet, nothing obvious that I saw.

src/adera/ShipResources.h
```diff
@@ -0,0 +1,158 @@
...
+ * making 1/(2^quanta) the smallest representable unit of the resource. The
+ * remaining 64-quanta bits of precision are left over for tank capacity.
+ */
+struct ShipResourceType
+{
+    // A short, unique, identifying name readable by both human and machine
+    std::string m_identifier;
+
+    // The full, screen-display name of the resource
+    std::string m_displayName;
```
Capital-Asterisk, line 63: I use to like doing display names like this, but it would make language localization harder to do. Some time in the future, there should be a locale config(s) that stores all display names.

src/adera/ShipResources.h
```diff
@@ -0,0 +1,158 @@
...
+
+#include "osp/Resource/Resource.h"
+#include "osp/Active/SysMachine.h"
+
+/* Sample resource:
+
+identifier: lox
+name: Liquid Oxygen
+quanta: 16
+mass: 1141.0 (kg)
```
Capital-Asterisk, line 36: Would be useful to add that 1141.0kg is the mass of 2^16 game units of LOX.

Capital-Asterisk (approved):
The biggest problem here is that I'm cringing at my own wiring code.

Also considering the idea of generic containers machines. Containers that can store some volume of anything. A variant of MachineContainer that stores multiple ShipResource, where the sum of their volumes can't exceed some limit?

Just a random thought, everything else is good.

src/adera/ShipResources.h
```diff
@@ -0,0 +1,175 @@
...
+
+inline MachineContainer::MachineContainer(MachineContainer&& move) noexcept
+    : Machine(std::move(move))
+    , m_capacity(std::move(move.m_capacity))
+    , m_contents(std::move(move.m_contents))
+{}
+
+inline MachineContainer& MachineContainer::operator=(MachineContainer&& move) noexcept
+{
+    m_enable = move.m_enable;
```
jonesmz, line 168: Conceptually, is it still enabled after being moved from?

I recommend std:;exchange for situations where the moved-from object shouldn't be in an inconstant state.

jonesmz (approved):

src/adera/ShipResources.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+    return nullptr;
+}
+
+std::vector<WireInput*> MachineContainer::existing_inputs()
+{
+    return m_inputs;
+}
+
+std::vector<WireOutput*> MachineContainer::existing_outputs()
```
jonesmz, line 54: These should be returning by const&.

Otherwise, you're forcing the caller to always make a copy, when maybe they just want to reference what you got.

src/adera/ShipResources.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        auto& container = view.get<MachineContainer>(ent);
+
+        // Do something useful here... or not
+    }
+}
+
+float SysMachineContainer::resource_volume(ShipResourceType type, uint64_t quantity)
+{
+    uint64_t units = quantity / (1ull << type.m_quanta);
```
jonesmz, line 81: Your intention here is to multiply 1 by 2, m_quanta times?

If I recall correctly, std::pow is floating point only (stupid).

I recommend making a utility function that abstracts the bit-shifting. Makes it much easier for future readers to easily understand what you're trying to accomplish.

Or, considering you're always casting to float anyway, maybe just use std::pow and do everything as float.

src/adera/ShipResources.h
jonesmz: Yea, localization is going to be important, and it is not something i've ever had to deal with before.

My understanding is that typically the recommended way to do localization is to tie each string to a stable ID number of some kind, and then do the string lookup based on that ID number in a localization file.

I don't think we can (directly) use EnTT for generating these ID numbers, because the ID numbers need to remain stable across builds. So We're really looking at needing to have some header file that just has a big series of inline constexpr uint64_t's starting at 0 and counting up. 

Every time we add a new string, it goes at the end.

Any time we remove a string, we don't re-shuffle ID numbers, just remove the one that was removed.

Then, before "release" (not meaningful for open source, but you get the point) we can run a script to compact all the IDs in that header, and also the localization files.

jonesmz (approved):

src/adera/ShipResources.cpp
z-adams: std::pow will work. I'm currently wondering how these functions worked at all since 1/(large int) is zero, so I'll need to take a closer look with my next PR. Maybe I fix them in the later implementation.

Issue #56, opened 2021-01-18
Improve Wiring system


Capital-Asterisk:
The wiring system is a core feature that connects a vehicle's Machines together, similar to a circuit simulator.

See: https://github.com/TheOpenSpaceProgram/OpenSpaceProgram/blob/master/docs/Technical.md#wiring

### Problem:

The current implementation is incomplete; and so far, it is not friendly to the cache, generally hard to use, and doesn't properly utilize the ECS. Data stored by wire values also use std::variant, which may not be the best solution.

Machines are large classes that store `WireInput`s and `WireOutput`s classes as members. `WireOutput`s use Magnum's linked list to connect to multiple `WireInput`s. These store pointers to the Machine components. Storing pointers or references to ECS components is not a good idea, as the buffers they're in can reallocate.

The reallocation issue is dealt with by having a complicated move constructors, where `WireInput`s and `WireOutput`s must be moved explicitly to update the pointer to the new location. These move constructors are also rather expensive, as each `WireOutput` moved has to update every input connected to it.

There is also the possibility of requiring multiple outputs connected to one input. For example, multiple rocket abort triggers (explosion detectors on parts) connected to a single abort input (trigger launch escape system).

### Possible solution:
Store the wire value (previously in `WireOutput`) in a new `ACompWireNode` component. `WireInput`s and `WireOutput`s can just be an entity (id) that an `ACompWireNode` is part of. `ACompWireNode` will also store a list of the entities of connected inputs and their propagate functions.

#### Data type of wire value:
See https://github.com/TheOpenSpaceProgram/osp-magnum/pull/49#discussion_r559089519

Capital-Asterisk:
Also see sdl2 (stupid digital logic simulator 2) https://gist.github.com/Capital-Asterisk/68c487804fed88bad3a502bbf413e2a5

Since nodes are just temporary states that only need to exist during updates, they don't need to be components. The use of std::variant can be avoided too. I'm thinking of storing single vectors of different node types for the entire scene, like ACompNodePercents for all wiretype::Percent nodes. Machines can refer elements in these vectors by reserving indices.

PR #61, opened 2021-01-24
Per-object machines, minimally functional fuel consumption

z-adams:
This one makes up for the last PR, now you see why I split it into two :)

At a high level, the changes primarily add the ability for machines to be owned by non-part objects. The new `PrototypeObject::m_machineIndices` array allows a PrototypeObject to keep track of the machines it owns. When a PrototypeObject is created in `AssetImporter::proto_add_obj_recurse()`, its machines are loaded by `AssetImporter::load_machines()`. This function takes a reference to the ProtoObject's index array, and the parent PrototypePart's machine array. PrototypeMachines are added to the PrototypePart's master list, and the indices of the machines populate the PrototypeObject's index array. When a part is instantiated in `SysVehicle::activate_sat()`, the `SysVehicle::part_instantiate()` function creates an array of objects which track the relationships between the new object `ActiveEnt`s and the PrototypePart's machine array, since the PrototypePart and PrototypeObject information is discarded once the entities have been loaded into the scene. A new function `SysVehicle::part_instantiate_machines()` takes this array, and uses it to add the appropriate machines to the various entities in the scene. This deferred action is necessary because machines may depend on the entire part's subhierarchy existing; the example given in the code is how `MachineRocket` needs the plume effect node to exist when it's instantiated.

The rest of the stuff is just a quick way of adding functional fuel consumption to `MachineRocket`. It's handled by an ugly and hardcoded solution right now (hardcoded tank ID, storing tank entity ID as `MachineRocket` member). This will change in the next PR, which generalized everything with a new "Pipe" wire type. If it's unacceptable to do the hardcoding (it currently renders the "moon" scene vessels immobile) I can include the next few commits in this PR as well, but it's already a large amount of code and making the PR 70% larger is going to suck.

The current functionality is basically that the 10m^3 fuselage tank is used to store a generic "fuel" which is drained by the rocket engines. When the fuel is depleted the engines cease to provide thrust, although the plumes keep working. The next PR will smooth out this proof of concept into a generalized system.

### DIFF
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -33,6 +33,7 @@
 #include "adera/Shaders/Phong.h"
 #include "adera/Shaders/PlumeShader.h"
 #include "osp/Resource/blueprints.h"
+#include "osp/PhysicsConstants.h"
 #include "adera/SysExhaustPlume.h"
 #include "adera/Plume.h"
 #include <Magnum/Trade/MeshData.h>
@@ -60,6 +61,15 @@ WireOutput* MachineRocket::request_output(WireOutPort port)
 
 std::vector<WireInput*> MachineRocket::existing_inputs()
 {
+    /*std::vector<WireInput*> inputs;
+    inputs.reserve(3 + m_resourceLines.size());
+
+    inputs.insert(inputs.begin(), {&m_wiGimbal, &m_wiIgnition, &m_wiThrottle});
+    for (auto& resource : m_resourceLines)
+    {
+        inputs.push_back(&resource.m_lineIn);
+    }
+    return inputs;*/
     return {&m_wiGimbal, &m_wiIgnition, &m_wiThrottle};
 }
 
@@ -92,7 +102,46 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
             continue;
         }
 
-        //std::cout << "updating a rocket\n";
+        // Check for nonzero throttle, continue otherwise
+        WireData *pThrottle = machine.m_wiThrottle.connected_value();
+        wiretype::Percent* pPercent = nullptr;
+        if (pThrottle != nullptr)
+        {
+            using wiretype::Percent;
+            pPercent = std::get_if<Percent>(pThrottle);
+            if ((pPercent == nullptr) || !(pPercent->m_value > 0.0f))
+            {
+                continue;
+            }
+        }
+        else
+        {
+            continue;
+        }
+
+        // Check for adequate resource inputs
+        bool fail = false;
+        for (auto& resource : machine.m_resourceLines)
+        {
+            auto const* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
+            if (!src)
+            {
+                std::cout << "Error: no source found\n";
+                fail = true;
+                break;
+            }
+            if (!src->check_contents().m_quantity > 0)
+            {
+                fail = true;
+                break;
+            }
+        }
+        if (fail)
+        {
+            continue;
+        }
+
+        // Perform physics calculation
 
         // Get rigidbody ancestor and its transformation component
         auto const* pRbAncestor =
@@ -105,34 +154,38 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
 
         }
 
-        using wiretype::Percent;
+        Matrix4 relTransform = pRbAncestor->m_relTransform;
 
-        if (WireData *pThrottle = machine.m_wiThrottle.connected_value();
-            pThrottle != nullptr)
+        /* Compute thrust force
+            * Thrust force is defined to be along +Z by convention.
+            * Obtains thrust vector in rigidbody space
+            */
+        Vector3 thrustDir = relTransform.transformVector(Vector3{0.0f, 0.0f, 1.0f});
+        float thrustMag = machine.m_params.m_maxThrust * pPercent->m_value;
+        // Take thrust in rigidbody space and apply to RB in world space
+        Vector3 thrust = thrustMag * thrustDir;
+        Vector3 worldThrust = rCompTf.m_transform.transformVector(thrust);
+        SysPhysics_t::body_apply_force(rCompRb, worldThrust);
+
+        // Obtain point where thrust is applied relative to RB CoM
+        Vector3 location = relTransform.translation();
+        // Compute worldspace torque from engine location, thrust vector
+        Vector3 torque = Magnum::Math::cross(location, thrust);
+        Vector3 worldTorque = rCompTf.m_transform.transformVector(torque);
+        SysPhysics_t::body_apply_torque(rCompRb, worldTorque);
+ 
+        // Perform resource consumption calculation
+        float massFlowRateTot = thrustMag /
+            (phys::constants::g_0 * machine.m_params.m_specImpulse);
+        for (auto const& resource : machine.m_resourceLines)
         {
-            Percent *pPercent = std::get_if<Percent>(pThrottle);
-            if (pPercent == nullptr) { continue; }
-
-            float thrustMag = machine.m_thrust;
-
-            Matrix4 relTransform = pRbAncestor->m_relTransform;
-
-            /* Compute thrust force
-             * Thrust force is defined to be along +Z by convention.
-             * Obtains thrust vector in rigidbody space
-             */
-            Vector3 thrustDir = relTransform.transformVector(Vector3{0.0f, 0.0f, 1.0f});
-            // Take thrust in rigidbody space and apply to RB in world space
-            Vector3 thrust = thrustMag*pPercent->m_value * thrustDir;
-            Vector3 worldThrust = rCompTf.m_transform.transformVector(thrust);
-            SysPhysics_t::body_apply_force(rCompRb, worldThrust);
-
-            // Obtain point where thrust is applied relative to RB CoM
-            Vector3 location = relTransform.translation();
-            // Compute worldspace torque from engine location, thrust vector
-            Vector3 torque = Magnum::Math::cross(location, thrust);
-            Vector3 worldTorque = rCompTf.m_transform.transformVector(torque);
-            SysPhysics_t::body_apply_torque(rCompRb, worldTorque);
+            float massFlowRate = massFlowRateTot * resource.m_massRateFraction;
+            float massFlow = massFlowRate * m_scene.get_time_delta_fixed();
+            uint64_t required = resource.m_type->resource_quantity(massFlow);
+            auto* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
+            uint64_t consumed = src->request_contents(required);
+            std::cout << "consumed " << consumed << " units of fuel, "
+                << src->check_contents().m_quantity << " remaining\n";
         }
     }
 }
@@ -182,10 +235,40 @@ Machine& SysMachineRocket::instantiate(ActiveEnt ent, PrototypeMachine config,
     BlueprintMachine settings)
 {
     // Read engine config
-    float thrust = std::get<double>(config.m_config["thrust"]);
-    
+    MachineRocket::Parameters params;
+    params.m_maxThrust = std::get<double>(config.m_config["thrust"]);
+    params.m_specImpulse = std::get<double>(config.m_config["Isp"]);
+
+    MachineRocket::fuel_list_t inputs;
+    Package& pkg = m_scene.get_application().debug_find_package("lzdb");
+    DependRes<ShipResourceType> fuel = pkg.get<ShipResourceType>("fuel");
+    if (!fuel.empty())
+    {
+        // TMP: finds the first non-empty fuel tank in the ship
+        // Only works as long as the fuselage is added before the engine
+
+        ActiveEnt foundTank = entt::null;
+        auto find_fuel = [this, &foundTank](ActiveEnt e)
+        {
+            if (auto* fuel = this->m_scene.reg_try_get<MachineContainer>(e);
+                fuel != nullptr && fuel->check_contents().m_quantity > 0)
+            {
+                foundTank = e;
+                return EHierarchyTraverseStatus::Stop;
+            }
+            return EHierarchyTraverseStatus::Continue;
+        };
+        auto const& parent = m_scene.reg_get<ACompHierarchy>(ent);
+        m_scene.hierarchy_traverse(parent.m_parent, find_fuel);
+        // endTMP
+
+        ActiveEnt fuelSource = foundTank;
+        MachineRocket::ResourceInput input = {std::move(fuel), 1.0f, fuelSource};
+        inputs.push_back(std::move(input));
+    }
+
     attach_plume_effect(ent);
-    return m_scene.reg_emplace<MachineRocket>(ent, thrust);
+    return m_scene.reg_emplace<MachineRocket>(ent, params, inputs);
 }
 
 Machine& SysMachineRocket::get(ActiveEnt ent)
--- a/src/adera/Machines/Rocket.h
+++ b/src/adera/Machines/Rocket.h
@@ -23,10 +23,11 @@
  * SOFTWARE.
  */
 #pragma once
-
+#include <utility>
 #include <osp/Active/SysMachine.h>
 #include <osp/Active/physics.h>
 #include <osp/Resource/blueprints.h>
+#include "adera/ShipResources.h"
 
 namespace adera::active::machines
 {
@@ -80,9 +81,22 @@ private:
 class MachineRocket : public osp::active::Machine
 {
     friend SysMachineRocket;
+    struct ResourceInput
+    {
+        osp::DependRes<ShipResourceType> m_type;
+        float m_massRateFraction;
+        osp::active::ActiveEnt m_sourceEnt;
+    };
+    using fuel_list_t = std::vector<ResourceInput>;
+
+    struct Parameters
+    {
+        float m_maxThrust;
+        float m_specImpulse;
+    };
 
 public:
-    MachineRocket(float thrust);
+    MachineRocket(Parameters params, fuel_list_t& resources);
     MachineRocket(MachineRocket &&move) noexcept;
     MachineRocket& operator=(MachineRocket&& move) noexcept;
 
@@ -103,21 +117,25 @@ private:
     osp::active::WireInput m_wiThrottle { this, "Throttle" };
 
     osp::active::ActiveEnt m_rigidBody  { entt::null };
-    float m_thrust{0.0f};
+    fuel_list_t m_resourceLines;
+
+    Parameters m_params;
 }; // MachineRocket
 
-inline MachineRocket::MachineRocket(float thrust)
- : Machine(true)
- , m_thrust(thrust)
+inline MachineRocket::MachineRocket(Parameters params, fuel_list_t& resources)
+    : Machine(true)
+    , m_params(params)
+    , m_resourceLines(std::move(resources))
 { }
 
 inline MachineRocket::MachineRocket(MachineRocket&& move) noexcept
- : Machine(std::move(move))
- , m_wiGimbal(this, std::move(move.m_wiGimbal))
- , m_wiIgnition(this, std::move(move.m_wiIgnition))
- , m_wiThrottle(this, std::move(move.m_wiThrottle))
- , m_rigidBody(std::move(move.m_rigidBody))
- , m_thrust(std::exchange(move.m_thrust, 0.0f))
+   : Machine(std::move(move))
+   , m_wiGimbal(this, std::move(move.m_wiGimbal))
+   , m_wiIgnition(this, std::move(move.m_wiIgnition))
+   , m_wiThrottle(this, std::move(move.m_wiThrottle))
+   , m_rigidBody(std::move(move.m_rigidBody))
+   , m_params(std::move(move.m_params))
+   , m_resourceLines(std::move(move.m_resourceLines))
 { }
 
 inline MachineRocket& MachineRocket::operator=(MachineRocket&& move) noexcept
@@ -127,7 +145,8 @@ inline MachineRocket& MachineRocket::operator=(MachineRocket&& move) noexcept
     m_wiIgnition = { this, std::move(move.m_wiIgnition) };
     m_wiThrottle = { this, std::move(move.m_wiThrottle) };
     m_rigidBody  = std::move(move.m_rigidBody);
-    m_thrust = std::exchange(move.m_thrust, 0.0f);
+    m_params = std::move(move.m_params);
+    m_resourceLines = std::move(move.m_resourceLines);
     return *this;
 }
 
--- a/src/adera/ShipResources.cpp
+++ b/src/adera/ShipResources.cpp
@@ -24,10 +24,39 @@
  */
 #include "ShipResources.h"
 #include "osp/Active/ActiveScene.h"
+#include "osp/Resource/PrototypePart.h"
 
 using namespace osp;
 using namespace osp::active;
 using namespace adera::active::machines;
+
+/* ShipResourceType */
+
+double ShipResourceType::resource_volume(uint64_t quantity) const
+{
+    double units = static_cast<double>(quantity) / std::pow(2.0, m_quanta);
+    return units * m_volume;
+}
+
+double ShipResourceType::resource_mass(uint64_t quantity) const
+{
+    double units = static_cast<double>(quantity) / std::pow(2.0, m_quanta);
+    return units * m_mass;
+}
+
+uint64_t ShipResourceType::resource_capacity(double volume) const
+{
+    double units = volume / m_volume;
+    double quantaPerUnit = std::pow(2.0, m_quanta);
+    return static_cast<uint64_t>(units * quantaPerUnit);
+}
+
+uint64_t ShipResourceType::resource_quantity(double mass) const
+{
+    double units = mass / m_mass;
+    double quantaPerUnit = std::pow(2.0, m_quanta);
+    return static_cast<uint64_t>(units * quantaPerUnit);
+}
 
 /* MachineContainer */
 
@@ -56,6 +85,17 @@ std::vector<WireOutput*> MachineContainer::existing_outputs()
     return m_outputs;
 }
 
+uint64_t MachineContainer::request_contents(uint64_t quantity)
+{
+    if (quantity > m_contents.m_quantity)
+    {
+        return std::exchange(m_contents.m_quantity, 0);
+    }
+
+    m_contents.m_quantity -= quantity;
+    return quantity;
+}
+
 /* SysMachineContainer */
 
 SysMachineContainer::SysMachineContainer(ActiveScene& rScene)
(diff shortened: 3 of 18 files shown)


src/osp/Resource/AssetImporter.cpp
```diff
@@ -474,6 +479,14 @@ void AssetImporter::proto_add_obj_recurse(TinyGltfImporter& gltfImporter,
         }
     }
 
+    // Check for and read machines
+    tinygltf::Node const& node =
+        *static_cast<tinygltf::Node const*>(childData->importerState());
+    if (node.extras.Has("machines"))
+    {
+        load_machines(node.extras, part.get_machines(), obj.m_machineIndices);
+    }
```
z-adams, line 488: When each object is loaded, its machine configs are loaded here

src/osp/Resource/AssetImporter.cpp
```diff
@@ -158,8 +145,26 @@ void osp::AssetImporter::load_part(TinyGltfImporter& gltfImporter,
             }
         }
 
-        part.get_machines().emplace_back(std::move(machine));
+        machineIndexArray.push_back(machineArray.size());
+        machineArray.emplace_back(std::move(machine));
```
z-adams, line 149: PrototypeMachines are added to the PrototypePart's machine array, and the indices are stored in the PrototypeObject's machineIndexArray

src/osp/Active/SysVehicle.cpp
```diff
@@ -331,8 +368,15 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
         }
-    }
 
+        // Save the list of machines this object owns
+        MachineDef machineDef
+        {
+            currentEnt,
+            currentPrototype.m_machineIndices
+        };
+        machineMapping.push_back(std::move(machineDef));
```
z-adams, line 378: When the objects are instantiated, the PrototypeObject's machines are saved in this array for later instantiation

src/osp/Active/SysVehicle.cpp
```diff
@@ -124,36 +124,30 @@ StatusActivated SysVehicle::activate_sat(ActiveScene &scene,
...
+        * Since machines may depend on the existence of other objects in the
+        * part's hierarchy, machine instantiation must be deferred until the
+        * entire hierarchy exists. For instance, a rocket engine's root
+        * "MachineRocket" machine may depend on the existence of a child node
+        * to receive an "ACompExhaustPlume" component that handles graphical
+        * effects related to the rocket machine, but that child node would not
+        * exist in time for the MachineRocket to be instantiated alongside
+        * the root part entity.
+        */
+        part_instantiate_machines(partEntity, machineMapping, proto, partBp);
```
z-adams, line 150: After the part and objects have all been instantiated, the machines are instantiated

src/osp/Active/SysVehicle.h
```diff
@@ -111,16 +112,70 @@ class SysVehicle : public IDynamicSystem, public IActivator
...
+     * @param rootParent [in] Entity to put part into
+     * @param machineDefinitions [out] List of part machines for later creation
      * @return Pointer to object created
      */
-    ActiveEnt part_instantiate(PrototypePart& part, ActiveEnt rootParent);
-
-    // Handle deleted parts and separations
-    void update_vehicle_modification(ActiveScene& rScene);
+    ActiveEnt part_instantiate(PrototypePart& part, BlueprintPart& blueprint,
+        ActiveEnt rootParent, std::vector<MachineDef>& machineDefinitions);
```
z-adams, line 178: I meant to have this return a tuple and use structured bindings instead of the `machineDefinitions` output parameter; I'll do it tomorrow

src/adera/Machines/Rocket.cpp
```diff
@@ -182,10 +236,22 @@ Machine& SysMachineRocket::instantiate(ActiveEnt ent, PrototypeMachine config,
...
+    MachineRocket::Parameters params;
+    params.m_maxThrust = std::get<double>(config.m_config["thrust"]);
+    params.m_specImpulse = std::get<double>(config.m_config["Isp"]);
+
+    MachineRocket::fuel_list_t inputs;
+    Package& pkg = m_scene.get_application().debug_find_package("lzdb");
+    DependRes<ShipResourceType> fuel = pkg.get<ShipResourceType>("fuel");
+    if (!fuel.empty())
+    {
+        ActiveEnt fuelSource = static_cast<ActiveEnt>(13);
```
z-adams, line 248: So this is pretty egregious, but generalizing this stuff is going to take this PR from +500 lines to +700, so perhaps this is acceptable for the time being

src/osp/Resource/PrototypePart.h
```diff
@@ -87,14 +95,8 @@ struct PrototypeObject
...
     // Put more OSP-specific data in here
-};
-
-using config_node_t = std::variant<double, int, std::string>;
 
-struct PrototypeMachine
-{
-    std::string m_type;
-    std::map<std::string, config_node_t> m_config;
+    std::vector<unsigned> m_machineIndices;
```
z-adams, line 99: This is the only functional change to this file; not sure why the other stuff got rearranged (probably during the cherry pick merge). I'll revert those changes before merging as they're just adding noise to an already large commit

src/adera/Machines/Rocket.h
```diff
@@ -80,9 +81,22 @@ class SysMachineRocket :
 class MachineRocket : public osp::active::Machine
 {
     friend SysMachineRocket;
+    struct ResourceInput
+    {
+        osp::DependRes<ShipResourceType> m_type;
+        float m_massRateFraction;
+        osp::active::ActiveEnt m_sourceEnt;
+    };
+    using fuel_list_t = std::vector<ResourceInput>;
```
z-adams, line 90: This arrangement kind of sucks, and will be changed in the next PR or two when the connections are automated/generalized

Capital-Asterisk (commented):
The biggest problem here is that I'm cringing at my own wiring code, but 10x more.

I also find it quite problematic how SysMachineRocket does the fuel management.

In issue #56 I talk about some 'node' system. Instead of storing states in WireOutput, it should be stored in a separate node component. This can do more than just logic.

All connected fuel containers and consumers can be connected to a common node using this theoretical wiring system, where it acts more like a pressure system instead. When a consumer needs fuel, it writes how much fuel it needs to the node, like negative pressure. This notifies connected fuel tanks which, will try to fill the void and set the node's pressure back to 0.

All connected tanks are known, and requirements from multiple consumers are combined, making it really simple for a system to determine how much fuel to take from each container.

Also consider pumping fuel between tanks, and fuel pipes that connect two different nodes which is making my head hurt right now.

1 / (1/R1 + 1/R2)

src/adera/Machines/Rocket.cpp
z-adams: I'm just manually assigning the feeding tank to the rocket.

Yes, it should be searching, but since any tank anywhere on the ship can be attached to any rocket anywhere on the ship, it needs a bit of an involved solution. I guess I could give it a config value for the time being and read that, it would just push the hard-codedness to the part config files. I will fix this in the next PR or two, but if you want to change it for now let me know how you think we should handle it. I can include my subsequent commits in this branch too, it'll just make the PR twice as bloated.

src/adera/Machines/Rocket.cpp
```diff
@@ -133,6 +173,20 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
             Vector3 torque = Magnum::Math::cross(location, thrust);
             Vector3 worldTorque = rCompTf.m_transform.transformVector(torque);
             SysPhysics_t::body_apply_torque(rCompRb, worldTorque);
+            
+ 
+            // Perform resource consumption calculation
+            float massFlowRateTot = thrustMag / (9.81f * machine.m_params.m_specImpulse);
```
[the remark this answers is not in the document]
z-adams: For something like fuel mass flow rate, it's enough. I would vote to put the scientific constants inside CommonPhysics.h/cpp; we could also put generic stuff like this fuel flow rate/rocket equation math there as well.

src/adera/Machines/Rocket.cpp
```diff
@@ -92,7 +101,40 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
...
+                continue;
+            }
+        }
+        else { continue; }
+
+        // Check for adequate resource inputs
+        bool fail = false;
+        for (auto& resource : machine.m_resourceLines)
+        {
+            MachineContainer* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
```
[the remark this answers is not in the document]
z-adams: I think ultimately some of the systems are going to get pretty complicated. Rocket engines in particular *are* kind of complicated because you have pretty much every possible system either driving them or being driven by them. The rocket engine depends on user command inputs, fuel availability, more complicated physics simulation stuff like temperature and fuel flow rate, and drives other physics (ship rigidbody physics, thermal stuff), fuel depletion, and graphical effects. I vote we keep expanding it until it becomes obvious that something should be separated. When we have a more functional game I'm down to try to optimize the architecture more, but frankly my goal at the moment is to start the "bloat it with features" stage of development and let that drive engine development and architecture discussions when we notice something getting ugly.

> "To properly plan a project, you have to do the project"
>  -- Capital-Asterisk [UTC-7]é#@ñ±, 11/17/2020

src/adera/Machines/Rocket.cpp
```diff
@@ -133,6 +173,20 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
...
+ 
+            // Perform resource consumption calculation
+            float massFlowRateTot = thrustMag / (9.81f * machine.m_params.m_specImpulse);
+            for (auto const& resource : machine.m_resourceLines)
+            {
+                float massFlowRate = massFlowRateTot * resource.m_massRateFraction;
+                float massFlow = massFlowRate * m_scene.get_time_delta_fixed();
+                uint64_t required = resource.m_type->resource_quantity(massFlow);
+                MachineContainer* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
+                uint64_t consumed = src->request_contents(required);
```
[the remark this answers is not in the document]
z-adams: Good point; it should calculate the mass flow required before withdrawing anything from the tanks. This should probably be combined with the earlier step where I check for fuel availability. I should probably test all this with a bipropellant rocket to make sure it works in general. We can discuss having a node intermediate between fuel sources/sinks elsewhere, I do like that idea.

As for `m_massRateFraction`, MachineRocket stores a vector `m_resourceLines` which lists all the types of fuel demanded by the engine, and the by-mass mixture ratio of each one. massFlowRate is the total fuel mass flow rate from earlier, multiplied by the mass mixture ratio to get the mass flow rate of the current resource being iterated over.

src/adera/Machines/Rocket.cpp
```diff
@@ -60,6 +60,15 @@ WireOutput* MachineRocket::request_output(WireOutPort port)
...
 {
+    /*std::vector<WireInput*> inputs;
+    inputs.reserve(3 + m_resourceLines.size());
+
+    inputs.insert(inputs.begin(), {&m_wiGimbal, &m_wiIgnition, &m_wiThrottle});
+    for (auto& resource : m_resourceLines)
+    {
+        inputs.push_back(&resource.m_lineIn);
+    }
+    return inputs;*/
```
jonesmz, line 71: prefer
```
#if 0
#endif 
```
for commenting out code.

jonesmz: or #if false

src/adera/Machines/Rocket.cpp
```diff
@@ -92,7 +101,40 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
...
+        if (throttle)
+        {
+            using wiretype::Percent;
+            percent = std::get_if<Percent>(throttle);
+            if (!(percent->m_value > 0.0f))
+            {
+                continue;
+            }
+        }
+        else { continue; }
```
jonesmz, line 116: i recommend against doing one-line else statements like this, unless its matched to a 1-line if statement.

Makes the flow of the program a bit hard to anticipate.

src/adera/Machines/Rocket.cpp
```diff
@@ -92,7 +101,40 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
...
 
-        //std::cout << "updating a rocket\n";
+        // Check for nonzero throttle, continue otherwise
+        WireData *throttle = machine.m_wiThrottle.connected_value();
+        wiretype::Percent* percent = nullptr;
+        if (throttle)
+        {
+            using wiretype::Percent;
+            percent = std::get_if<Percent>(throttle);
+            if (!(percent->m_value > 0.0f))
```
jonesmz, line 111: if std::get_if returns a nullptr, you just crashed the program.

src/adera/Machines/Rocket.cpp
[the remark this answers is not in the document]
jonesmz: Why not PhysicsConstants.h ?

src/adera/Machines/Rocket.h
```diff
@@ -103,21 +117,25 @@ class MachineRocket : public osp::active::Machine
...
-    float m_thrust{0.0f};
+    fuel_list_t m_resourceLines;
+
+    Parameters m_params;
 }; // MachineRocket
 
-inline MachineRocket::MachineRocket(float thrust)
- : Machine(true)
- , m_thrust(thrust)
+inline MachineRocket::MachineRocket(Parameters params, fuel_list_t& resources)
```
jonesmz, line 125: don't take by l-value reference if you are going to unconditionally std::move. Just take the parameter by value in that case.

src/adera/ShipResources.cpp
```diff
@@ -24,11 +24,40 @@
...
 #include "osp/Active/ActiveScene.h"
+#include "osp/Resource/PrototypePart.h"
 
 using namespace osp;
 using namespace osp::active;
 using namespace adera::active::machines;
 
+/* ShipResourceType */
+
+double ShipResourceType::resource_volume(uint64_t quantity) const
```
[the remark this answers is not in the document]
jonesmz: i agree that if you're always using a multiple of two, you could create a constexpr bitshift function for that.

src/adera/ShipResources.cpp
```diff
@@ -24,11 +24,40 @@
...
+double ShipResourceType::resource_mass(uint64_t quantity) const
+{
+    double units = static_cast<double>(quantity) / std::pow(2.0, m_quanta);
+    return units * m_mass;
+}
+
+uint64_t ShipResourceType::resource_capacity(double volume) const
+{
+    double units = volume / m_volume;
+    double quantaPerUnit = std::pow(2.0, m_quanta);
```
jonesmz, line 50: if you're always turning m_quanta into 2^m_quanta, why even have m_quanta?

src/adera/ShipResources.cpp
```diff
@@ -56,6 +85,19 @@ std::vector<WireOutput*> MachineContainer::existing_outputs()
     return m_outputs;
 }
 
+uint64_t MachineContainer::request_contents(uint64_t quantity)
+{
+    if (quantity > m_contents.m_quantity)
+    {
+        uint64_t remainder = m_contents.m_quantity;
+        m_contents.m_quantity = 0;
+        return remainder;
```
jonesmz, line 94: return std::exchange(m_contents.m_quantity, 0);

src/osp/Active/SysVehicle.cpp
```diff
@@ -124,36 +124,28 @@ StatusActivated SysVehicle::activate_sat(ActiveScene &scene,
 
         PrototypePart &proto = *partDepends;
 
-        ActiveEnt partEntity = this->part_instantiate(proto, vehicleEnt);
+        auto [partEntity, machineMapping] = this->part_instantiate(proto, partBp,
```
jonesmz, line 127: auto const& ?

src/osp/Active/SysVehicle.cpp
```diff
@@ -227,20 +219,68 @@ int SysVehicle::deactivate_sat(ActiveScene &scene, SysAreaAssociate &area,
     return 0;
 }
 
+void SysVehicle::add_machines_to_object(ActiveEnt partEnt, ActiveEnt objEnt,
+    std::vector<PrototypeMachine> const& protoMachines,
+    std::vector<BlueprintMachine> const& blueprintMachines,
+    std::vector<unsigned> const& machineIndices)
+{
+    if (machineIndices.empty()) { return; }
```
jonesmz, line 227: recommend adding a newline after this.

src/osp/Active/SysVehicle.cpp
```diff
@@ -227,20 +219,68 @@ int SysVehicle::deactivate_sat(ActiveScene &scene, SysAreaAssociate &area,
...
-                                       ActiveEnt rootParent)
+        if (!(m_scene.system_machine_it_valid(sysMachine)))
+        {
+            std::cout << "Machine type: " << protoMachine.m_type << " Not found\n";
+            continue;
+        }
+
+        // Instantiate the machine for the object
+        Machine& machine = sysMachine->second->instantiate(objEnt,
+            protoMachine, bpMachine);
```
jonesmz, line 247: why bother holding a reference to Machine& machine if you don't use it?

src/osp/Active/SysVehicle.cpp
```diff
@@ -227,20 +219,68 @@ int SysVehicle::deactivate_sat(ActiveScene &scene, SysAreaAssociate &area,
...
+
+void SysVehicle::part_instantiate_machines(ActiveEnt partEnt,
+    std::vector<MachineDef> machineMapping,
+    PrototypePart& part, BlueprintPart& partBP)
 {
+    std::vector<PrototypeMachine> const& protoMachines = part.get_machines();
+    std::vector<BlueprintMachine> const& bpMachines = partBP.m_machines;
+    for (auto const& obj : machineMapping)
+    {
+        add_machines_to_object(partEnt, obj.m_machineOwner,
```
jonesmz, line 262: unless you're going to std::move the m_machineOwner and m_machineIndicies objects into the add_machines_to_object, then you should take the std::vector in the part_instantiate_machines function parameter lists by const&.

src/osp/Active/SysVehicle.cpp
```diff
@@ -227,20 +219,68 @@ int SysVehicle::deactivate_sat(ActiveScene &scene, SysAreaAssociate &area,
...
+std::pair<ActiveEnt, std::vector<SysVehicle::MachineDef>> SysVehicle::part_instantiate(
+    PrototypePart& part, BlueprintPart& blueprint, ActiveEnt rootParent)
+{
     std::vector<PrototypeObject> const& prototypes = part.get_objects();
     std::vector<ActiveEnt> newEntities(prototypes.size());
 
+    /* A list of MachineDefs, which catalog all part machines and the ActiveEnts
+     * that are created to represent the objects that own them
+     */
+    std::vector<MachineDef> machineMapping;
```
jonesmz, line 276: machineMapping.reserve(prototypes.size());

src/osp/Active/SysVehicle.cpp
```diff
@@ -227,20 +219,68 @@ int SysVehicle::deactivate_sat(ActiveScene &scene, SysAreaAssociate &area,
...
+
     //std::cout << "size: " << newEntities.size() << "\n";
 
-    for (unsigned i = 0; i < prototypes.size(); i ++)
+    for (unsigned i = 0; i < prototypes.size(); i++)
```
jonesmz, line 280: unsigned is icky.

src/osp/Active/SysVehicle.cpp
jonesmz: Another reserve is needed inside the loop as well.

 std::vector<DependRes<Texture2D>> textureResources;
            for (unsigned i = 0; i < drawable.m_textures.size(); i++)
            {

src/osp/Active/SysVehicle.h
```diff
@@ -111,17 +112,77 @@ class SysVehicle : public IDynamicSystem, public IActivator
...
     void update_vehicle_modification(ActiveScene& rScene);
 
+private:
+
+    // Stores the association between a PrototypeObj entity and the indices of
+    // its owned machines in the PrototypePart array
+    struct MachineDef
+    {
+        ActiveEnt m_machineOwner;
+        std::vector<unsigned> const& m_machineIndices;
```
jonesmz, line 126: are you sure you want to have a member variable that is a const-reference to a vector? that's weird.

src/osp/Active/SysVehicle.h
```diff
@@ -111,17 +112,77 @@ class SysVehicle : public IDynamicSystem, public IActivator
...
+     * @param partEnt [in] The part entity which owns the ACompMachines component
+     * @param entity [in] The target entity to receive the machines
+     * @param protoMachines [in] The master array of prototype machines
+     * @param blueprintMachines [in] The master array of machine configs
+     * @param machineIndices [in] The indices of the machines to add to entity
+     */
+    void add_machines_to_object(ActiveEnt partEnt, ActiveEnt entity,
+        std::vector<PrototypeMachine> const& protoMachines,
+        std::vector<BlueprintMachine> const& blueprintMachines,
+        std::vector<unsigned> const& machineIndices);
```
jonesmz, line 150: if some third party library is using "unsigned" as an index into itself, typedef that to a named type. E.g. LibOtherIndex_t or similar.

src/osp/Active/SysVehicle.h
```diff
@@ -111,17 +112,77 @@ class SysVehicle : public IDynamicSystem, public IActivator
...
+     * Effectively just calls add_machines_to_object() for each object in the
+     * part.
+     * 
+     * @param partEnt [in] The root entity of the part
+     * @param machineMapping [in] The mapping from objects to their machines
+     * @param part [in] The prototype part being created
+     * @param partBP [in] The blueprint configs of the part being created
+     */
+    void part_instantiate_machines(ActiveEnt partEnt,
+        std::vector<MachineDef> machineMapping,
```
jonesmz, line 171: sure about passing by value?

src/osp/Resource/AssetImporter.cpp
jonesmz: Make sure to reserve space in the vector before calling emplace in a loop.

Note though that if this function is being called in a loop, don't reserve a whole bunch, instead figure out a way to get the total size the vect needs to be, and reserve once at the top of the stack.

src/osp/Resource/blueprints.cpp
```diff
@@ -56,12 +56,20 @@ void BlueprintVehicle::add_part(
 
     // now we have a part index.
 
-    // just create a new part blueprint will all the data
+    // Create and default initialize object blueprint machines
+    size_t numMachines = prototype->get_machines().size();
+    std::vector<BlueprintMachine> machineBPs(numMachines, BlueprintMachine{});
```
jonesmz, line 61: I believe that std::vector::resize would accomplish this goal slightly more efficently.

src/osp/Resource/blueprints.cpp
```diff
@@ -56,12 +56,20 @@ void BlueprintVehicle::add_part(
...
 
-    BlueprintPart blueprint{partIndex, translation, rotation, scale};
+    BlueprintPart blueprint
+    {
+        partIndex,
+        translation, rotation, scale,
```
jonesmz, line 66: weird newlines?

src/test_application/main.cpp
```diff
@@ -291,6 +310,37 @@ void debug_print_update_order()
...
+    ActiveScene& scene = g_ospMagnum->get_scenes().begin()->second;
+    auto view = scene.get_registry().view<osp::active::ACompMachines>();
+
+    for (ActiveEnt ent : view)
+    {
+        ACompHierarchy& hier = scene.reg_get<ACompHierarchy>(ent);
+        std::cout << "[" << int(ent) << "]: " << hier.m_name << "\n";
+
+        auto& machines = scene.reg_get<ACompMachines>(ent);
+        for (auto mach : machines.m_machines)
```
jonesmz, line 335: don't get too crazy with the use of auto, it makes it hard for readers to pick up, via context clues, what's going on.

src/test_application/main.cpp
```diff
@@ -291,6 +310,37 @@ void debug_print_update_order()
...
+        {
+            ActiveEnt machEnt = mach.m_partEnt;
+            std::string sysName = mach.m_system->first;
```
jonesmz, line 338: const-ref.

jonesmz (approved):

src/adera/Machines/Rocket.cpp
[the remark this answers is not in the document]
z-adams: We can, it just seems sort of unnecessary to split them given that there won't be very many of them; 5 or 6, maybe?

src/adera/Machines/Rocket.cpp
z-adams: Really? Why?

src/adera/ShipResources.cpp
z-adams: It's tied to the number of bits used to represent the base unit of resource. You could make an enum or something instead but you can't just pick any old quantaPerUnit.

I will concede though that the language is really confusing in all of this stuff. "Quanta" and "unit" are both super ambiguous words in this context; "quanta" might have to change so that it's easier to differentiate the "unit" 1 and the arbitrary unit chosen by the user.

src/adera/Machines/Rocket.cpp
jonesmz: Much easier for a lot of code editors to handle gracefully, in addition to not having weird behavior when it comes to nested comments.

src/osp/Active/SysVehicle.h
z-adams: It is weird. The thing is, the objects of this type only live in a local function scope. In `activate_vehicle()`, there's a `PrototypePart` which owns the `PrototypeObjects` that store the machineIndices arrays. These `MachineDef` objects are created by `part_instantiate()` and filled with the machineIndices arrays, and consumed by `part_machine_instantiate()` immediately afterwards. I'm open to suggestions though.

jonesmz: If its only used in one place just declare the class inside the function in question.

Otherwise put a comment in the class explaining the weird.

src/adera/ShipResources.cpp
Capital-Asterisk: Hmm, now that I think of it, m_quanta is there to increase the precision of the other parameters. If m_quanta is decremented, then the mass, volume, and density can be divided by two to compensate. Though it is pretty useful to describe resources as "1.0m^3 of LOX is 1141.0kg, the game specifically divides this into 65536 units."

m_quanta can just be stored as some unsigned int instead, and is enforced to be a power of two.

A better name might be long but more obvious one: "quantaPerOneUnit"

src/adera/Machines/Rocket.cpp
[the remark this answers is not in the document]
jonesmz: Small headers aren't a problem, I encourage using a stand alone header file if it keeps the rest of the code free of attractive nuisances like constant values that get replicated across files.

src/osp/Resource/AssetImporter.cpp
z-adams: For the `machineIndexArray`, that's easy (and I may actually just change that parameter to a return value). For the `machineArray`, it's not so easy. There's not really any way to know how many machines will be in the part's hierarchy unless we add a second recursion pass just to pre-allocate machine storage. I think it's probably best to just let `std::vector` do its thing and then maybe shrink it to fit afterwards, unless you think the allocations are really such a big deal that it's worth doing a pre-pass to figure out how much space it needs. Otherwise, I'll just leave a comment on line 149 for now that pre-allocation would be nice, and if we ever find ourselves needing to do multiple passes over the config for some reason, we can add this at that point.

Capital-Asterisk (approved):
I'd say it's best to wait until the new wiring system is implemented before making any major changes, so I'll approve this for now.

src/osp/Active/SysVehicle.h
z-adams: It's only used in one place, except that two member functions need to return and accept them, so comment it is :P

src/osp/Resource/blueprints.cpp
z-adams: Resize where? You mean like
```cpp
std::vector<BlueprintMacine> machineBPs;
machineBPs.resize(numMacines);
```
?

jonesmz: Yep.

Resize calls default constructor directly.

Your code is using the copy constructor from a default value.

A smart optimizer should figure it out, but not all optimizers are smart, so better to make it do fewer things. Less places to get wrong.

Then again, maybe it makes zero difference. Its always hard to really predict these things.

Issue #46, opened 2021-01-10
Make all Systems static


Capital-Asterisk:
Systems in ECS are suppose to be functions, but here they're polymorphic instances of classes (`osp::active::IDynamicSystem`). This wasn't a good design choice on my end in the first place. Having the system class store state leads to complications, such as unnecessary coupling (that can be fixed only with virtual functions).

The original intention for dynamic systems, was enabling systems to link together and call each others member functions. This isn't needed, systems should communicate with each other through only components. 

OOP decouples systems through virtual interfaces. Pseudo-code example:
```cpp
// Imaginary rocket that needs the physics engine to apply thrust
anyPhysicsEngine->apply_force(rigidBody, thrust);
// Physics engine functions are called right away

```

DOD decouples systems through common data. Pseudo-code example:
```cpp
// Imaginary rocket that needs the physics engine to apply thrust
genericRigidBodyData.force += thrust;
// Physics engine reads the data later on when its system updates
```

The `m_scene` member in most system classes isn't needed. `UpdateOrderHandle_t` can be stored in some container in ActiveScene with an identifier

 
For example: SysFFGravity, dealing with gravity force fields looks like this:
```cpp
class SysFFGravity : public SysForceField // SysForceField is empty class right now
{
public:

    static const std::string smc_name;

    SysFFGravity(ActiveScene &scene);
    ~SysFFGravity() = default;

    void update_force(ActiveScene& rScene);

private:

    ActiveScene &m_scene;

    UpdateOrderHandle_t m_updateForce;
};
```

When it can be refactored to this:
```cpp
class SysFFGravity : public SysForceField
{
public:

    static const std::string smc_name;

    static void update_force(ActiveScene& rScene);
};
```

Capital-Asterisk:
Done in #99

PR #62, opened 2021-01-26
Rework SysAreaAssociate

Capital-Asterisk:
This PR guts `SysAreaAssociate`. This system is responsible for ~~loading~~ activating nearby Satellites into the ActiveScene. Satellites, like planets and vehicles, are stored in the Universe as simple representations, with no meshes, no node hierarchies, and no rigid body physics. Only when the user gets close to them (entering the ActiveArea), do they need to be 'activated' into the scene.

In accordance to Issue #46, many functions were made static, and state is moved into a new component: `ACompAreaLink`. The `IActivator` interface used by `SysVehicle` and `SysPlanet` is also removed. This was previously used as a callback for `SysAreaAssociate` to activate and deactivate Satellites right away. (Just as a nearby Satellite is detected, it calls the activate function right away, adding new entities and components to the scene.)

Instead of callbacks, this new system stores [just entered], and [just exited] event vectors in `ACompAreaLink`, which also keeps track of which Satellites are currently inside the area. It is up to another system function to read these events where their component pools can be modified predictably. This makes the application more cache-friendly, and easier for future concurrency.

Other changes:
* Fix reallocation error in ActiveScene::hier_destroy

### DIFF
--- a/src/osp/Active/ActiveScene.cpp
+++ b/src/osp/Active/ActiveScene.cpp
@@ -121,12 +121,14 @@ void ActiveScene::hier_set_parent_child(ActiveEnt parent, ActiveEnt child)
 
 void ActiveScene::hier_destroy(ActiveEnt ent)
 {
-    auto &entHier = m_registry.get<ACompHierarchy>(ent);
+    auto *pEntHier = &m_registry.get<ACompHierarchy>(ent);
 
-    // recurse into children. this can be optimized more but i'm lazy
-    while (entHier.m_childCount)
+    // Destroy children first recursively, if there is any
+    while (pEntHier->m_childCount > 0)
     {
-        hier_destroy(entHier.m_childFirst);
+        hier_destroy(pEntHier->m_childFirst);
+        // previous hier_destroy might have caused a reallocation
+        pEntHier = m_registry.try_get<ACompHierarchy>(ent);
     }
 
     hier_cut(ent);
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -26,18 +26,22 @@
 
 #include "ActiveScene.h"
 
+using osp::active::SysAreaAssociate;
 
-using namespace osp::universe;
-using namespace osp::active;
-using namespace osp;
+using osp::active::ACompAreaLink;
+
+using osp::universe::Universe;
+using osp::universe::Satellite;
+
+using osp::universe::UCompActivatable;
+using osp::universe::UCompActivationRadius;
+using osp::universe::UCompActiveArea;
 
 const std::string SysAreaAssociate::smc_name = "AreaAssociate";
 
-SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene, Universe &uni)
- : m_scene(rScene)
- , m_universe(uni)
- , m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
-                [this] (ActiveScene& rScene) { this->update_scan(rScene); })
+SysAreaAssociate::SysAreaAssociate(ActiveScene &rScene)
+ : m_updateScan(rScene.get_update_order(), "areascan", "physics", "",
+                &SysAreaAssociate::update_scan)
 {
 
 }
@@ -46,17 +50,37 @@ void SysAreaAssociate::update_scan(ActiveScene& rScene)
 {
     // scan for nearby satellites, maybe move this somewhere else some day
 
-    if (!m_universe.get_reg().valid(m_areaSat))
+    ACompAreaLink *pArea = try_get_area_link(rScene);
+
+    if (pArea == nullptr)
     {
         return;
     }
 
-    //Universe& uni = m_universe;
-    //std::vector<Satellite>& satellites = uni.
-    auto view = m_universe.get_reg().view<UCompActivatable>();
+    Universe &rUni = pArea->m_rUniverse;
+    Satellite areaSat = pArea->m_areaSat;
 
-    for (Satellite sat : view)
+    auto &areaUComp = rUni.get_reg().get<UCompActiveArea>(areaSat);
+
+    auto viewActRadius = rUni.get_reg().view<UCompActivationRadius,
+                                             UCompActivatable>();
+
+    for (Satellite sat : viewActRadius)
     {
+        auto satRadius = viewActRadius.get<UCompActivationRadius>(sat);
+
+        float distSqr = rUni.sat_calc_pos_meters(areaSat, sat).dot();
+        float radius = areaUComp.m_areaRadius + satRadius.m_radius;
+
+        if ((radius * radius) > distSqr)
+        {
+            std::cout << "nearby!\n";
+        }
+
+    }
+
+    //for (Satellite sat : view)
+    //{
 //        if (SatelliteObject* obj = sat.get_object())
 //        {
 //            //std::cout << "obj: " << obj->get_id().m_name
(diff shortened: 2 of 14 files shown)


src/osp/Active/ActiveScene.cpp
```diff
@@ -121,12 +121,14 @@ void ActiveScene::hier_set_parent_child(ActiveEnt parent, ActiveEnt child)
 
 void ActiveScene::hier_destroy(ActiveEnt ent)
 {
-    auto &entHier = m_registry.get<ACompHierarchy>(ent);
+    auto *pEntHier = &m_registry.get<ACompHierarchy>(ent);
 
-    // recurse into children. this can be optimized more but i'm lazy
-    while (entHier.m_childCount)
+    // Destroy children first recursively, if there is any
+    while (pEntHier->m_childCount > 0)
```
jonesmz, line 127: check for null before dereferencing.

If you're positive that it'll never be null (which I don't think you are, or it wouldn't be "try_get()", then use an assert that's only enabled in debug builds.

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -26,157 +26,162 @@
...
+        float radius = areaUComp.m_areaRadius + satRadius.m_radius;
 
+        if ((radius * radius) > distanceSquared)
+        {
+            if (alreadyInside)
+            {
+                continue; // Satellite already activated
+            }
+            // Satellite is nearby, activate it!
+            auto entered = activations.m_inside.emplace(sat, entt::null).first;
```
jonesmz, line 107: might be better to use C++17 structured bindings here.

Is m_inside a map? i assume it is

auto const& [iter, bool] = activations.m_inside.emplace(sat, entt::null);
activations.m_enter.emplace_back(iter);

src/osp/Active/SysVehicle.cpp
```diff
@@ -317,40 +313,73 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+    }
+
+    Universe &rUni = pArea->get_universe();
+    TypeSatIndex type = rUni.sat_type_find_index<universe::SatVehicle>();
+    ActivationTracker& activations = pArea->get_tracker(type);
+
+    // Delete vehicles that have gone too far from the ActiveArea range
+    for (auto const &[sat, ent] : activations.m_leave)
+    {
+        rScene.hier_destroy(ent);
```
jonesmz, line 355: (void)sat; before rScene.hier_destroy(ent) to shut up the compiler about unused variable warnings.

src/osp/Active/SysVehicle.cpp
```diff
@@ -317,40 +313,73 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+    }
+
+    // Update transforms of already-activated vehicle satellites
+    auto view = rScene.get_registry().view<ACompVehicle, ACompTransform,
+                                           ACompActivatedSat>();
+    for (ActiveEnt vehicleEnt : view)
+    {
+        auto &vehicleTf = view.get<ACompTransform>(vehicleEnt);
```
jonesmz, line 363: That issue that the EnTT author commented on might be useful in this specific location? maybe?

src/osp/Active/SysVehicle.cpp
```diff
@@ -317,40 +313,73 @@ ActiveEnt SysVehicle::part_instantiate(PrototypePart& part,
...
+        auto &vehicleSat = view.get<ACompActivatedSat>(vehicleEnt);
+        SysAreaAssociate::sat_transform_set_relative(
+            rUni, pArea->m_areaSat, vehicleSat.m_sat, vehicleTf.m_transform);
+    }
+
+    // Activate nearby vehicle satellites that have just entered the ActiveArea
+    for (auto &entered : activations.m_enter)
+    {
+        universe::Satellite sat = entered->first;
+        ActiveEnt &rEnt = entered->second;
```
jonesmz, line 373: i'd say structured bindinds, but it looks like m_enter is a container of pointers based on this code?

jonesmz: I suppose you could write a wrapper class that had iterators that automatically dereferenced for you, to get the nice syntax.

though, i'm imagining that you don't want todo that.

src/osp/Universe.h
```diff
@@ -136,6 +136,12 @@ class Universe
     constexpr std::vector<std::string_view> const& sat_type_get_names() const
     noexcept { return m_typeSatNames; }
 
+    /**
+     * @return Number of reistered satellites
+     */
+    std::underlying_type<TypeSatIndex>::type sat_type_count() const noexcept
```
jonesmz, line 142: std::underlying_type_t<TypeSatIndex> ?

src/osp/Active/SysVehicle.cpp
jonesmz: Could look something like (there are lots of problems with the below example)
```cpp
template<typename RANGE_T>
struct AutoDerefRange
{
    using underlyingIt_t = decltype(m_range.begin());
    struct Iterator
    {
        Iterator(underlyingIt_t value) : m_value(value) {};
        underlyingIt_t m_value;
        decltype(auto) operator*() { return *(m_value.operator*());
        decltype(auto) operator&() { return &(m_value.operator*());
        operator++() { m_value++; }
    }
    AutoDerefRange(RANGE_T & range) : m_range(range) {}
    Iterator begin() { return m_range.begin(); }
    Iterator end() { return m_range.end(); }
}
```

src/osp/Active/SysAreaAssociate.h
```diff
@@ -34,125 +34,120 @@
...
+     , m_activated(rUniverse.sat_type_count())
+    { }
+
+    ActivationTracker& get_tracker(universe::TypeSatIndex type)
+    { return m_activated[std::size_t(type)]; }
+
+    universe::Universe& get_universe() noexcept
+    { return m_rUniverse.get(); }
+
+    universe::Satellite m_areaSat;
```
z-adams, line 63: I've never really looked at how the sys area associate stuff works. So is the area itself a satellite? It's associated more closely to the universe than to the scene?

jonesmz:
Something to consider for future work is having different distances for each of loading assets from disk, assembling models for those assets in ram, pre-loading the stuff to the GPU, and then actually rendering. As well as the reverse process for things moving farther away.

Loading stuff from disk should be done asyncronously on another thread, just started when the precondition is triggered.

src/osp/Active/SysVehicle.cpp
Capital-Asterisk: `activations.m_enter` is a `std::vector<unordered_map<Satellite, ActiveEnt>::iterator>`s. It tracks new elements added to `activations.m_inside`, allowing the values to be modified (an ActiveEnt) without needing to access the map again. It might be better as a `std::pair<Satellite, std::reference_wrapper<ActiveEnt>>`.

z-adams (approved):
I've scanned everything, I think I actually understand how it all works now and it's pretty neat. Feel free to merge when ready, I'll adapt my per-objects machine to these changes afterwards. It will be interesting to test all this when we couple the active scenes to an underlying solar system.

src/osp/Active/SysVehicle.cpp
jonesmz: My intention was to say that you could get sexier syntax here at the place you're using the std::vector if you wrote a wrapper class to help get sexier syntax.

jonesmz: As for references versus iterators, i'd point out that the same conditions that invalidate an iterator also typically invalidate the thing it's pointing to.

So if a reference makes your life easier....

jonesmz (approved):

PR #64, opened 2021-01-31
Fueltank physics

z-adams:
Fuel tanks are upgraded to store their masses and keep them up-to-date based on their current contents.

- Added `MachineContainer::compute_mass()` which returns the mass of the current contents of the container

- ACompMass is now automatically assigned to MachineContainers; SysMachineContainer keeps them up to date.

- Added ACompContainerShape, which represents the shape of MachineContainers for mass distribution calc

- Added `MachineRocket::current_power_output()` which reports the most recent power level produced by the engine; SysExhaustPlume now reads from this function rather than getting itself tangled up in throttle wire values, which wouldn't handle situations like the engine being off at full throttle due to lack of fuel.

Newton:
Now that containers have dynamically updated masses, rigidbodies can now be dynamically updated to account for shifting fuel mass distribution. When MachineRocket draws fuel from a container, it sets a flag which causes SysNewton to recalculate the mass distribution of the ship (which should ultimately change when we upgrade the wire/pipe system, so that these manual details aren't necessary).

- `struct DataRigidBody` received a new `m_rigidbodyDirty` member which behaves just like `m_colliderDirty`, but instead triggers the rigidbody inertia to be recalculated in `SysNewton::update_world()`.

- Add `SysNewton::update_rigidbody_inertia()` which encapsulates the logic for sending inertia data to Newton, previously located in `create_body()`

- `compute_mass_inertia()` now accepts a center of mass offset as parameter; previously the inertia was calculated from the part's origin, but now the CoM can be precomputed and passed for more physically accurate inertial calculations

- Renamed `compute_body_CoM()` to `compute_hier_CoM()` since its task is really to generally calculate the center of mass of all ACompMasses in a hierarchy subtree.

- Renamed `compute_body_inertia()` to `compute_rigidbody_inertia()`

### DIFF
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -102,14 +102,16 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
             continue;
         }
 
+        machine.m_powerOutput = 0.0f;  // Will be set later if engine is on
+
         // Check for nonzero throttle, continue otherwise
         WireData *pThrottle = machine.m_wiThrottle.connected_value();
-        wiretype::Percent* pPercent = nullptr;
+        wiretype::Percent* pThrotPercent = nullptr;
         if (pThrottle != nullptr)
         {
             using wiretype::Percent;
-            pPercent = std::get_if<Percent>(pThrottle);
-            if ((pPercent == nullptr) || !(pPercent->m_value > 0.0f))
+            pThrotPercent = std::get_if<Percent>(pThrottle);
+            if ((pThrotPercent == nullptr) || !(pThrotPercent->m_value > 0.0f))
             {
                 continue;
             }
@@ -157,11 +159,11 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         Matrix4 relTransform = pRbAncestor->m_relTransform;
 
         /* Compute thrust force
-            * Thrust force is defined to be along +Z by convention.
-            * Obtains thrust vector in rigidbody space
-            */
+         * Thrust force is defined to be along +Z by convention.
+         * Obtains thrust vector in rigidbody space
+         */
         Vector3 thrustDir = relTransform.transformVector(Vector3{0.0f, 0.0f, 1.0f});
-        float thrustMag = machine.m_params.m_maxThrust * pPercent->m_value;
+        float thrustMag = machine.m_params.m_maxThrust * pThrotPercent->m_value;
         // Take thrust in rigidbody space and apply to RB in world space
         Vector3 thrust = thrustMag * thrustDir;
         Vector3 worldThrust = rCompTf.m_transform.transformVector(thrust);
@@ -173,11 +175,13 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         Vector3 torque = Magnum::Math::cross(location, thrust);
         Vector3 worldTorque = rCompTf.m_transform.transformVector(torque);
         SysPhysics_t::body_apply_torque(rCompRb, worldTorque);
- 
+        
+        rCompRb.m_inertiaDirty = true;
+
         // Perform resource consumption calculation
         float massFlowRateTot = thrustMag /
             (phys::constants::g_0 * machine.m_params.m_specImpulse);
-        for (auto const& resource : machine.m_resourceLines)
+        for (MachineRocket::ResourceInput const& resource : machine.m_resourceLines)
         {
             float massFlowRate = massFlowRateTot * resource.m_massRateFraction;
             float massFlow = massFlowRate * m_scene.get_time_delta_fixed();
@@ -187,6 +191,10 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
             std::cout << "consumed " << consumed << " units of fuel, "
                 << src->check_contents().m_quantity << " remaining\n";
         }
+
+        // Set output power level (for plume effect)
+        // TODO: later, take into account low fuel pressure, bad mixture, etc.
+        machine.m_powerOutput = pThrotPercent->m_value;
     }
 }
 
--- a/src/adera/Machines/Rocket.h
+++ b/src/adera/Machines/Rocket.h
@@ -111,6 +111,20 @@ public:
     std::vector<osp::active::WireInput*> existing_inputs() override;
     std::vector<osp::active::WireOutput*> existing_outputs() override;
 
+    /**
+     * Return normalized power output level of the rocket this frame
+     *
+     * Returns a value [0,1] corresponding to the current output power of the
+     * engine. This value is equal to the throttle input level, unless the
+     * engine has run out of fuel, has a nonlinear throttle response, or some
+     * similar reason. Used primarily by SysExhaustPlume to determine what the
+     * exhaust plume effect should look like.
+     *
+     * @return normalized float [0,1] representing engine power output
+     */
+    float current_output_power() const
+    { return m_powerOutput; }
+
 private:
     osp::active::WireInput m_wiGimbal   { this, "Gimbal"   };
     osp::active::WireInput m_wiIgnition { this, "Ignition" };
@@ -118,8 +132,10 @@ private:
 
     osp::active::ActiveEnt m_rigidBody  { entt::null };
     fuel_list_t m_resourceLines;
-
     Parameters m_params;
+
+    // Rocket power output for the current frame
+    float m_powerOutput{0.0f};
 }; // MachineRocket
 
 inline MachineRocket::MachineRocket(Parameters params, fuel_list_t& resources)
@@ -136,6 +152,7 @@ inline MachineRocket::MachineRocket(MachineRocket&& move) noexcept
    , m_rigidBody(std::move(move.m_rigidBody))
    , m_params(std::move(move.m_params))
    , m_resourceLines(std::move(move.m_resourceLines))
+   , m_powerOutput(std::exchange(move.m_powerOutput, 0.0f))
 { }
 
 inline MachineRocket& MachineRocket::operator=(MachineRocket&& move) noexcept
@@ -147,6 +164,7 @@ inline MachineRocket& MachineRocket::operator=(MachineRocket&& move) noexcept
     m_rigidBody  = std::move(move.m_rigidBody);
     m_params = std::move(move.m_params);
     m_resourceLines = std::move(move.m_resourceLines);
+    m_powerOutput = std::exchange(move.m_powerOutput, 0.0f);
     return *this;
 }
 
--- a/src/adera/ShipResources.h
+++ b/src/adera/ShipResources.h
@@ -27,6 +27,7 @@
 
 #include "osp/Resource/Resource.h"
 #include "osp/Active/SysMachine.h"
+#include "osp/CommonPhysics.h"
 
 namespace adera::active::machines
 {
@@ -130,7 +131,6 @@ public:
     osp::active::Machine& get(osp::active::ActiveEnt ent) override;
 
 private:
-    
     osp::active::UpdateOrderHandle_t m_updateContainers;
 }; // class SysMachineContainer
 
@@ -168,6 +168,12 @@ public:
      * @return The amount of resource that was received
      */
     uint64_t request_contents(uint64_t quantity);
+
+    /**
+     * Compute the current mass of the container contents
+     * @return The current mass, in kg
+     */
+    float compute_mass() const noexcept;
 private:
     std::vector<osp::active::WireInput*> m_inputs;
     std::vector<osp::active::WireOutput*> m_outputs;
--- a/src/osp/Active/SysNewton.cpp
+++ b/src/osp/Active/SysNewton.cpp
@@ -26,6 +26,8 @@
 
 #include "ActiveScene.h"
 
+#include "adera/ShipResources.h"
+
 #include <Newton.h>
 
 using osp::active::SysNewton;
@@ -38,8 +40,6 @@ using osp::active::ACompNwtWorld;
 using osp::active::ACompTransform;
 
 using namespace osp;
-using Magnum::Vector4;
-using Magnum::Matrix4;
 
 // Callback called for every Rigid Body (even static ones) on NewtonUpdate
 void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
@@ -112,7 +112,7 @@ SysNewton::SysNewton(ActiveScene &scene)
     scene.get_registry().on_destroy<ACompNwtBody>()
                     .connect<&SysNewton::on_body_destruct>();
 
-    scene.get_registry().on_destroy<ACompCollisionShape>()
+    scene.get_registry().on_destroy<ACompCollider>()
                     .connect<&SysNewton::on_shape_destruct>();
 
     scene.get_registry().on_destroy<ACompNwtWorld>()
@@ -123,7 +123,7 @@ SysNewton::~SysNewton()
 {
     // Clean up newton dynamics stuff
     m_scene.get_registry().clear<ACompNwtBody>();
-    m_scene.get_registry().clear<ACompCollisionShape>();
+    m_scene.get_registry().clear<ACompCollider>();
 }
 
 void SysNewton::update_world(ActiveScene& rScene)
@@ -163,11 +163,19 @@ void SysNewton::update_world(ActiveScene& rScene)
             entBody.m_colliderDirty = false;
         }
 
+        // Initialize body if not done so yet;
         if (entBody.m_body == nullptr)
         {
-            // Initialize body if not done so yet;
             create_body(rScene, ent, nwtWorld);
         }
+
+        // Recompute inertia and center of mass if rigidbody has been modified
+        if (entBody.m_inertiaDirty)
+        {
+            compute_rigidbody_inertia(rScene, ent);
+            std::cout << "Updating RB: new CoM Z=" << entBody.m_centerOfMassOffset.z() << "\n";
+        }
+
     }
 
     // Update the world
@@ -201,8 +209,7 @@ void SysNewton::find_colliders_recurse(ActiveScene& rScene, ActiveEnt ent,
         auto const &childHeir = rScene.reg_get<ACompHierarchy>(nextChild);
         auto const &childTransform = rScene.reg_get<ACompTransform>(nextChild);
 
-        auto* childCollide = rScene.get_registry()
-                                .try_get<ACompCollisionShape>(nextChild);
+        auto* childCollide = rScene.reg_try_get<ACompCollider>(nextChild);
 
         Matrix4 childMatrix = transform * childTransform.m_transform;
 
@@ -248,11 +255,12 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
     auto& entTransform = rScene.reg_get<ACompTransform>(entity);
 
     auto& entBody  = rScene.reg_get<ACompNwtBody>(entity);
-    auto* entShape = rScene.get_registry().try_get<ACompCollisionShape>(entity);
+    auto* entCollider = rScene.reg_try_get<ACompCollider>(entity);
+    auto* entShape = rScene.reg_try_get<ACompShape>(entity);
 
-    if (entShape == nullptr)
+    if ((entCollider == nullptr) || (entShape == nullptr))
     {
-        // Entity must also have a collision shape to make a body
+        // Entity must have a collision shape to define the collision volume
         return;
     }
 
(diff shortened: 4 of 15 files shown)


src/osp/Active/SysNewton.cpp
```diff
@@ -37,6 +39,8 @@ using osp::active::ACompNwtBody;
 using osp::active::ACompNwtWorld;
 using osp::active::ACompTransform;
 
+using adera::active::machines::ACompContainerShape;
```
z-adams, line 42: Previously I had `ACompContainerShape` defined in ActiveScene, but that doesn't make sense so I moved it to ShipResources.h. The downside is that it requires coupling SysNewton explicitly to the container stuff. Maybe we need a generic "empty shape" primitive enum in the CommonPhysics.h or something?

src/osp/Active/SysNewton.cpp
```diff
@@ -571,29 +594,42 @@ Matrix3 SysNewton::compute_mass_inertia(ActiveScene& rScene, ActiveEnt part)
...
+                shape = childCollide->m_shape;
+                
+            }
+            else if (childContainer)
+            {
+                auto& containerMass = rScene.reg_get<ACompMass>(nextChild);
+
+                mass = containerMass.m_mass;
+                shape = childContainer->m_shape;
+            }
```
z-adams, line 621: The annoying thing about this arrangement is that different types of masses need to be computed differently; perhaps part densities should be precalculated and assigned to ACompMass entities like I've done with the containers? That would allow us to avoid coupling the details of the mass distribution to Newton, especially if we had some way to unify the shape representation.

src/osp/Active/SysNewton.cpp
Capital-Asterisk: This as a pretty serious issue. There should be a system that converts ACompContainerShape data into collision shapes and mass. Is ACompCollisionShape and ACompMass not able to represent every kind of mass distribution?

src/adera/Machines/Rocket.h
```diff
@@ -111,15 +111,30 @@ class MachineRocket : public osp::active::Machine
...
     osp::active::WireInput m_wiGimbal   { this, "Gimbal"   };
     osp::active::WireInput m_wiIgnition { this, "Ignition" };
     osp::active::WireInput m_wiThrottle { this, "Throttle" };
 
     osp::active::ActiveEnt m_rigidBody  { entt::null };
     fuel_list_t m_resourceLines;
-
     Parameters m_params;
+
+    float m_lastPowerOutput{0.0f};
```
Capital-Asterisk, line 137: This wouldn't really be a 'last' power output. Just use `m_powerOutput`; the values calculated in update_physics are the ones used that very frame.

src/osp/CommonPhysics.h
```diff
@@ -62,6 +62,7 @@ struct DataRigidBody
     Vector3 m_centerOfMassOffset{0, 0, 0};
 
     bool m_colliderDirty{false}; // set true if collider is modified
+    bool m_rigidbodyDirty{false}; // set true if rigidbody is modified
```
Capital-Asterisk, line 65: If this is only for inertia, then it should be named `m_inertiaDirty`

src/adera/Machines/Rocket.cpp
```diff
@@ -173,11 +175,13 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         Vector3 torque = Magnum::Math::cross(location, thrust);
         Vector3 worldTorque = rCompTf.m_transform.transformVector(torque);
         SysPhysics_t::body_apply_torque(rCompRb, worldTorque);
- 
+        
+        rCompRb.m_rigidbodyDirty = true;
```
Capital-Asterisk, line 179: The rocket changed the inertia?

src/adera/ShipResources.h
```diff
@@ -198,4 +204,9 @@ inline MachineContainer& MachineContainer::operator=(MachineContainer&& move) no
     return *this;
 }
 
+struct ACompContainerShape
+{
+    osp::phys::ECollisionShape m_shape{osp::phys::ECollisionShape::NONE};
```
Capital-Asterisk, line 209: It sounds like a good idea to separate collision shapes from inertia-only shapes, instead of having this specifically made for containers.

src/osp/Active/SysNewton.h
```diff
@@ -193,22 +192,34 @@ class SysNewton : public IDynamicSystem
...
 
+    /**
+     * Update the inertia properties of a rigid body
+     *
+     * Given an existing rigid body, computes and updates the mass matrix and
+     * center of mass. Entirely self contained, calls the other inertia
+     * functions in this class.
+     * @param entity [in] The rigid body to update
+     */
+    static void update_rigidbody_inertia(ActiveScene& rScene, ActiveEnt entity);
```
Capital-Asterisk, line 203: This looks like an update function when it isn't. Should this be "compute_" like the other functions?

src/adera/Machines/Rocket.cpp
z-adams: In the current configuration, the rocket is the system that tells the containers how much fuel is being withdrawn from the container. I know you didn't like this before, but until we have a better pipe/wire system I'm not sure if there's a reasonable way to resolve it. In any case, it makes sense to set the dirty flag wherever the fuel withdrawal is calculated.

src/osp/Active/SysNewton.cpp
z-adams: The root of the problem is that the mass of parts can be distributed over multiple collision volumes, whereas fuel tanks can't. The fuel tank is represented by a single empty, whose volume is used to calculate it capacity. To compute the mass of a fuel tank, you simply check its ACompMass directly. To compute the mass of a part collider, since there may be multiple colliders, you have to compute the total volume of all colliders in the part, divide the part's "massdry" by this volume to get the `partDensity`, and then use the density to calculate the individual mass of each child collider (another name that sounds odd out of context) in order to compute a realistic inertia for the composite part. This means that somewhere, there needs to be a distinction made between the sources of mass.

Alternatively, we could add logic that would distribute the massdry of a part over its collider volumes from the beginning; the tradeoff is that it would no longer be possible to immediately retrieve the massdry of the part from the ACompMass (although it could be stored in the part component or something), but at least that way the masses could all be treated homogeneously and there'd be no need to discriminate between different types of volumes (so we could get rid of the `ACompContainerShape` and have one unified "shape" enum).

src/osp/Active/SysNewton.h
z-adams: Yep

src/osp/Active/SysNewton.cpp
```diff
@@ -162,12 +164,19 @@ void SysNewton::update_world(ActiveScene& rScene)
             NewtonDestroyBody(std::exchange(entBody.m_body, nullptr));
             entBody.m_colliderDirty = false;
         }
-
+
```
jonesmz, line 167: whitespace changes :-)

src/osp/Active/SysNewton.cpp
```diff
@@ -248,9 +256,10 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
     auto& entTransform = rScene.reg_get<ACompTransform>(entity);
 
     auto& entBody  = rScene.reg_get<ACompNwtBody>(entity);
-    auto* entShape = rScene.get_registry().try_get<ACompCollisionShape>(entity);
+    auto* entCollider = rScene.reg_try_get<ACompCollider>(entity);
+    auto* entShape = rScene.reg_try_get<ACompShape>(entity);
 
-    if (entShape == nullptr)
+    if (entCollider == nullptr || entShape == nullptr)
```
jonesmz, line 262: parenthesis

src/osp/Active/SysNewton.cpp
```diff
@@ -311,20 +302,17 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt entity,
...
             {
-                NewtonBodySetCollision(entBody.m_body, entShape->m_collision);
+                NewtonBodySetCollision(entBody.m_body, entCollider->m_collision);
             }
             else
             {
                 entBody.m_body = NewtonCreateDynamicBody(nwtWorld,
-                                                         entShape->m_collision,
-                                                         Matrix4().data());
+                    entCollider->m_collision, Matrix4().data());
```
jonesmz, line 314: .data() from a default-constructed Matrix4 ? I feel like this warrents a comment elaborating.

src/osp/Active/SysNewton.cpp
```diff
@@ -486,18 +492,34 @@ void SysNewton::body_apply_torque(ACompRigidBody_t &body, Vector3 torque) noexce
...
 {
     Vector3 localCoM{0.0f};
     float localMass = 0.0f;
 
-    for (ActiveEnt nextChild = root; nextChild != entt::null;)
+    if (includeRootMass)
     {
-        auto const &childHeir = rScene.reg_get<ACompHierarchy>(nextChild);
+        auto const* rootMass = rScene.reg_try_get<ACompMass>(root);
+        if (rootMass != nullptr) { localMass += rootMass->m_mass; }
```
jonesmz, line 513: tying back to our discussion about exceptions, i'm wondering about the benefit of having an ActiveEnt that represents a root that can't be located?

That seems like an "exceptional" situation that we should make the game bork-out about, instead of just adding nothing to the mass.

z-adams: In this case, the thing being found is the mass component, so `rootMass` is type `ACompMass`. `try_get<T>(ent)` tries to find component `T` belonging to entity `ent`, and if `ent` doesn't have a `T` then nullptr is returned. In this case, `includeRootMass` simply indicates that you're interested in the root node's mass if it has one, and if it doesn't you don't care (it skips it just like this function skips any other massless entity).

src/osp/Active/SysNewton.cpp
```diff
@@ -486,18 +492,34 @@ void SysNewton::body_apply_torque(ACompRigidBody_t &body, Vector3 torque) noexce
...
+    }
+
+    for (ActiveEnt nextChild = rScene.reg_get<ACompHierarchy>(root).m_childFirst;
+        nextChild != entt::null;)
```
jonesmz, line 517: wasn't there a range-based for-loop function for a hierachy that we could use?

Or is this recursing down a tree structure?

Ah, yea it is.

Can you check to see if (maybe even just ask directly) there's any utility functions in EnTT that'll make this sexier? like some kind of special notation that we can use to tell EnTT that we want a Tree structure, and then a way to ask for pre-order /  post order / level-order / so on.

src/osp/Active/SysNewton.cpp
```diff
@@ -525,128 +548,45 @@ Vector4 SysNewton::compute_body_CoM(ActiveScene& rScene,
...
-    Vector3 centerOfMass, Matrix4 currentTransform)
-{
-    Matrix3 localI{0.0f};
-    float localMass = 0.0f;
-
-    for (ActiveEnt nextChild = root; nextChild != entt::null;)
+    // Include entity's own inertia, if it has a mass and volume from which to compute it
+    auto const* mass = rScene.reg_try_get<ACompMass>(entity);
+    auto const* shape = rScene.reg_try_get<ACompShape>(entity);
+    if (mass != nullptr && shape != nullptr)
```
jonesmz, line 575: parenthesis

src/osp/Active/SysNewton.cpp
```diff
@@ -525,128 +548,45 @@ Vector4 SysNewton::compute_body_CoM(ActiveScene& rScene,
...
-        auto const &childTransform = rScene.reg_get<ACompTransform>(nextChild);
+        Matrix4& transform = rScene.reg_get<ACompTransform>(entity).m_transform;
+        Vector3 principalAxes = phys::collider_inertia_tensor(
+            shape->m_shape, transform.scaling(), mass->m_mass);
 
-        Matrix4 childMatrix = currentTransform * childTransform.m_transform;
+        Matrix3 localInertiaTensor{};
+        localInertiaTensor[0][0] = principalAxes.x();
+        localInertiaTensor[1][1] = principalAxes.y();
+        localInertiaTensor[2][2] = principalAxes.z();
```
jonesmz, line 584: I'm slightly surprised that the Matrix3 class doesn't have a static function to do this for you.

jonesmz: This seems like a common enough usage to warrant creating a function for it.

src/osp/Active/SysNewton.cpp
```diff
@@ -525,128 +548,45 @@ Vector4 SysNewton::compute_body_CoM(ActiveScene& rScene,
...
-    for (ActiveEnt nextChild = root; nextChild != entt::null;)
+    // Include entity's own inertia, if it has a mass and volume from which to compute it
+    auto const* mass = rScene.reg_try_get<ACompMass>(entity);
+    auto const* shape = rScene.reg_try_get<ACompShape>(entity);
+    if (mass != nullptr && shape != nullptr)
     {
-        auto const &childHier = rScene.reg_get<ACompHierarchy>(nextChild);
-        auto const &childTransform = rScene.reg_get<ACompTransform>(nextChild);
+        Matrix4& transform = rScene.reg_get<ACompTransform>(entity).m_transform;
+        Vector3 principalAxes = phys::collider_inertia_tensor(
```
jonesmz, line 578: Don't forget to add liberal use of "const", even on non-reference variables

src/osp/Active/SysNewton.cpp
z-adams: Too bad we can't inject a constructor into it or something, glm has this functionality (to initialize a diagonal matrix from a vector)

src/osp/Active/SysNewton.h
```diff
@@ -193,77 +197,51 @@ class SysNewton : public IDynamicSystem
...
+    * possess an ACompMass component are used to compute a center of mass for 
+    * the entire subtree, treating it as a system of point masses. By default,
+    * the root entity's mass is not included; to include it, the optional
+    * includeRootMass argument can be set to 'true'.
     * 
     * @param rScene           [in] ActiveScene containing relevant scene data
     * @param root             [in] Entity at the root of the hierarchy subtree
-    * @param currentTransform [in] Transformation of root with respect to the
-    *                              initial call's root
+    * @param includeRootMass  [in] Set to true if the root entity's mass should
```
jonesmz, line 221: should this really be a runtime option? or a compile time option?

You could make this a non-type template parameter (bool), and then use if constexpr inside the function to omit the extra code at runtime for code that doesn't want the mass added.

src/osp/Active/SysNewton.h
```diff
@@ -193,77 +197,51 @@ class SysNewton : public IDynamicSystem
...
-     * @param centerOfMass [in] The center of mass of the rigid body
-     * 
-     * @return The inertia tensor of the rigid body about its center of mass
+     *
+     * @return The inertia tensor of the rigid body about its center of mass, and 
+     *         a 4-vector containing xyz=CoM, w=total mass
      */
-    static Matrix3 compute_body_inertia(ActiveScene& rScene, ActiveEnt root,
-        Vector3 centerOfMass, Matrix4 currentTransform);
+    static std::pair<Matrix3, Magnum::Vector4> compute_hier_inertia(ActiveScene& rScene,
```
jonesmz, line 243: Inconsistant namespacing? Should Magnum:: be added to Matrix3? or removed from Vector4?

src/osp/Active/SysNewton.h
z-adams: I considered that, but the interface is a little ugly. `compute_hier_CoM<true>()` isn't the most intuitive thing; I guess in either case you still have to check the header to see what the function actually does though, so maybe it's worth it.

src/osp/Active/SysNewton.h
```diff
@@ -285,9 +263,8 @@ class SysNewton : public IDynamicSystem
 };
 
 template<class TRIANGLE_IT_T>
-void SysNewton::shape_create_tri_mesh_static(ACompCollisionShape &shape,
-                                             TRIANGLE_IT_T const& start,
-                                             TRIANGLE_IT_T const& end)
+void SysNewton::shape_create_tri_mesh_static(ACompShape& rShape,
+    ACompCollider& rCollider, TRIANGLE_IT_T const& start, TRIANGLE_IT_T const& end)
```
jonesmz, line 267: just take `start` by value, iterate it directly, and skip creating `TRIANGLE_IT_T it`

src/osp/Active/SysNewton.h
z-adams: The most common Magnum types are `using`'d in one of the common osp files, maybe Vector4 should be included there.

src/osp/Active/SysNewton.h
z-adams: I didn't write this function, I just updated it since I changed how the colliders work; I'd prefer not to screw with it here.

src/osp/Active/SysVehicle.cpp
```diff
@@ -261,11 +261,31 @@ void SysVehicle::part_instantiate_machines(ActiveScene& rScene, ActiveEnt partEn
...
+        auto const* shape = rScene.get_registry().try_get<ACompShape>(ent);
+        if (shape)
+        {
+            auto const& xform = rScene.reg_get<ACompTransform>(ent);
+            volume += shape_volume(shape->m_shape, xform.m_transform.scaling());
+        }
+        return EHierarchyTraverseStatus::Continue;
+    };
+
+    rScene.hierarchy_traverse(part, checkVol);
```
jonesmz, line 278: std::move(checkVol)

src/osp/Active/SysVehicle.cpp
```diff
@@ -261,11 +261,31 @@ void SysVehicle::part_instantiate_machines(ActiveScene& rScene, ActiveEnt partEn
...
+    float volume = 0.0f;
+    auto checkVol = [&rScene, &volume](ActiveEnt ent)
+    {
+        auto const* shape = rScene.get_registry().try_get<ACompShape>(ent);
+        if (shape)
+        {
+            auto const& xform = rScene.reg_get<ACompTransform>(ent);
+            volume += shape_volume(shape->m_shape, xform.m_transform.scaling());
+        }
+        return EHierarchyTraverseStatus::Continue;
```
jonesmz, line 275: I suspect that the compiler *should* be detecting that this lambda always returns the same value unconditionally,

but if you wanted to be super conservative, we could improve hierarchy_traverse to check the return type of the lambda, and for lambdas that return void, instead of trying to use the return type, it assumes that the lambda would have returned ::Continue.

You'd do something like

template<typename FUNCTOR_T, bool ALWAYS_CONTINUE = std::is_same_t<void, std::result_of_t<FUNCTOR_T(ActiveEnt)>>>
blah foo(FUNCTOR_T)
{
    if constexpr(ALWAYS_CONTINUE)
    {
        // simple version
    }
    else
    {
        // complicated version
    }
}

jonesmz: Or you could use partial template specialization.

```
template<typename FUNCTOR_T,
         bool ALWAYS_CONTINUE = std::is_void_t<std::result_of_t<FUNCTOR_T(ActiveEnt)>>>
blah foo(FUNCTOR_T)
{
    static_assert("Well that's not supposed to happen...");
}

template<typename FUNCTOR_T>
blah foo<FUNCTOR_T, true>(FUNCTOR_T)
{
    // simple version
}

template<typename FUNCTOR_T>
blah foo<FUNCTOR_T, false>(FUNCTOR_T)
{
    // complex version version
}
```

src/osp/Active/SysNewton.h
jonesmz: Use an enum as the template type.

compute_hier_CoM<ERootMass::Include> vs compute_hier_CoM<ERootMass::Ignore>

jonesmz: Shame there is no `switch constexpr(val)` syntax.

src/osp/Active/SysNewton.h
jonesmz: My recommendation is to use Magnum:: everywhere :-)

But consistency is nice, one way or another.

src/osp/Active/SysNewton.cpp
z-adams: EnTT has no concept of trees, that's up to us. Which part did you want to change specifically, the `if` condition?

src/osp/Active/SysNewton.cpp
jonesmz: That's fair

src/osp/Active/SysNewton.cpp
jonesmz: Hrmm, that's really frustrating. It *should* have the concept of a hierarchy.

My goal was to say that the notion of traversing a tree, in the way that you're doing here, should really be something that the library provides us, instead of something we have to do manually.

For example, you're doing
```
        nextChild = childHier.m_siblingNext;
```

at the end of the loop, but it would be substantially easier for a reader of the code to understand what's going on if the concept of traversing a tree, the way you're doing here, was encapsulated inside some kind of iterator pair (or an iterator, and a c++20 iterator-sentinal) that internally embodied the tree traversal order.

z-adams: To be honest, it really shouldn't since ECS is supposed to be a framework removed from any concepts like that. It's not EnTT's job to deal with scene graphs. A fancy iterator scheme might be doable though.

jonesmz: EnTT has far far more features than simply being dumb storage. There's no reason why it shouldn't provide basic "relationship traversal" capabilities out of the box, specifically because it has direct access to it's own internal design details that make it easier to write properly optimized traversal tools.

Capital-Asterisk (approved):

PR #65, opened 2021-02-12
Pipe wires

z-adams:
You can call me the white hat of the wiring system. I abuse the system horribly so we can find out how to stop people like me from breaking things.

The new `Pipe` wire type is, to cap's horror, just a wrapper for the `ActiveEnt` at the source end. Right now, this allows for connections between fuel tanks and engines to be programmatically defined in the ship blueprint (good). The MachineRocket at the sink end uses it (in Rocket.cpp:187) to reach into the source and withdraw whatever resources it demands (bad). The primary function of the pipe wire is to allow pipes to be defined in blueprints; we need to figure out what the interface presented to either end of the pipe should be.

Right now, the pipes originate in MachineContainer, and are accepted as an argument by MachineRocket and stored as a member vector, and are iterated over in the main `MachineRocket::update` loop as before. Hopefully this new use case will help stimulate some brainstorming over how to improve the pipe system.

### DIFF
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -61,16 +61,15 @@ WireOutput* MachineRocket::request_output(WireOutPort port)
 
 std::vector<WireInput*> MachineRocket::existing_inputs()
 {
-    /*std::vector<WireInput*> inputs;
+    std::vector<WireInput*> inputs;
     inputs.reserve(3 + m_resourceLines.size());
 
     inputs.insert(inputs.begin(), {&m_wiGimbal, &m_wiIgnition, &m_wiThrottle});
     for (auto& resource : m_resourceLines)
     {
-        inputs.push_back(&resource.m_lineIn);
+        inputs.push_back(&resource.m_source);
     }
-    return inputs;*/
-    return {&m_wiGimbal, &m_wiIgnition, &m_wiThrottle};
+    return inputs;
 }
 
 std::vector<WireOutput*> MachineRocket::existing_outputs()
@@ -125,18 +124,23 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         bool fail = false;
         for (auto& resource : machine.m_resourceLines)
         {
-            auto const* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
-            if (!src)
+            if (auto const* pipe = resource.m_source.get_if<wiretype::Pipe>();
+                pipe != nullptr)
             {
-                std::cout << "Error: no source found\n";
-                fail = true;
-                break;
+                
+                if (auto const* src = rScene.reg_try_get<MachineContainer>(pipe->m_source);
+                    src != nullptr)
+                {
+                    uint64_t required = resource_units_required(rScene, machine,
+                        pThrotPercent->m_value, resource);
+                    if (src->check_contents().m_quantity > required)
+                    {
+                        continue;
+                    }
+                }
             }
-            if (!src->check_contents().m_quantity > 0)
-            {
-                fail = true;
-                break;
-            }
+            fail = true;
+            break;
         }
         if (fail)
         {
@@ -179,17 +183,17 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         rCompRb.m_inertiaDirty = true;
 
         // Perform resource consumption calculation
-        float massFlowRateTot = thrustMag /
-            (phys::constants::g_0 * machine.m_params.m_specImpulse);
         for (MachineRocket::ResourceInput const& resource : machine.m_resourceLines)
         {
-            float massFlowRate = massFlowRateTot * resource.m_massRateFraction;
-            float massFlow = massFlowRate * m_scene.get_time_delta_fixed();
-            uint64_t required = resource.m_type->resource_quantity(massFlow);
-            auto* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
-            uint64_t consumed = src->request_contents(required);
+            // Pipe must be non-null since we checked earlier
+            const auto& pipe = *resource.m_source.get_if<wiretype::Pipe>();
+            auto& src = rScene.reg_get<MachineContainer>(pipe.m_source);
+
+            uint64_t required = resource_units_required(rScene, machine,
+                pThrotPercent->m_value, resource);
+            uint64_t consumed = src.request_contents(required);
             std::cout << "consumed " << consumed << " units of fuel, "
-                << src->check_contents().m_quantity << " remaining\n";
+                << src.check_contents().m_quantity << " remaining\n";
         }
 
         // Set output power level (for plume effect)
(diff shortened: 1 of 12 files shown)


src/adera/Machines/Rocket.cpp
```diff
@@ -125,7 +124,8 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         bool fail = false;
         for (auto& resource : machine.m_resourceLines)
         {
-            auto const* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
+            auto const* pipe = resource.m_source.get_if<wiretype::Pipe>();
+            auto const* src = rScene.reg_try_get<MachineContainer>(pipe->m_source);
             if (!src)
             {
                 std::cout << "Error: no source found\n";
```
z-adams, line 131: Check not only for the presence of resources, but actually ensure that enough resources are available (in accordance with the consumption ratios) before allowing physics update

-- Capital Asterisk, somewhere I can't find

z-adams:
Fix "quanta" in accordance with https://github.com/TheOpenSpaceProgram/osp-magnum/pull/61#discussion_r564281262

z-adams:
More quanta improvements; `constexpr`-ify things if possible: https://github.com/TheOpenSpaceProgram/osp-magnum/pull/58#discussion_r563229209

src/adera/Machines/Rocket.cpp
```diff
@@ -125,7 +124,8 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         bool fail = false;
         for (auto& resource : machine.m_resourceLines)
         {
-            auto const* src = m_scene.reg_try_get<MachineContainer>(resource.m_sourceEnt);
+            auto const* pipe = resource.m_source.get_if<wiretype::Pipe>();
+            auto const* src = rScene.reg_try_get<MachineContainer>(pipe->m_source);
```
Capital-Asterisk, line 128: `pipe` here can still be null, and causes a crash. Also use an `== nullptr` check instead of an implicit boolean conversion.

Another issue, on line 135, the ! operator is being applied to m_quantity, not the entire if statement. Put brackets, or even better, change to `0 ==`. Silly enough `(!m_quantity) > 0` still works.

Capital-Asterisk (changes requested):

src/adera/Machines/Rocket.cpp
```diff
@@ -61,16 +61,15 @@ WireOutput* MachineRocket::request_output(WireOutPort port)
 
 std::vector<WireInput*> MachineRocket::existing_inputs()
 {
-    /*std::vector<WireInput*> inputs;
+    std::vector<WireInput*> inputs;
```
jonesmz, line 64: seems a bit weird to use a raw pointer for something that you know is always non-null.

std::reference_wrapper might be closer to what you really want.

src/adera/Machines/Rocket.cpp
jonesmz: you probably want something like

if(auto const* pipe = resource.m_source.get_if<wiretype::Pipe>(); nullptr != pipe)
{
    auto const* src = rScene.reg_try_get<MachineContainer>(pipe->m_source);

    .... other stuff
}

jonesmz: And yea, line 135 is wrong. Operator precedence is a bitch.

Had a guy ask me why i'm so insistent on parenthesis usages the other day, and this is basically the reason.

src/adera/Machines/Rocket.cpp
```diff
@@ -183,10 +183,12 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
             (phys::constants::g_0 * machine.m_params.m_specImpulse);
         for (MachineRocket::ResourceInput const& resource : machine.m_resourceLines)
         {
+            const auto* pipe = resource.m_source.get_if<wiretype::Pipe>();
```
jonesmz, line 186: check for null.

Unless somehow you know that you'll always get a non-null

in which case, use a reference instead.

src/adera/Machines/Rocket.cpp
```diff
@@ -247,32 +249,15 @@ Machine& SysMachineRocket::instantiate(ActiveEnt ent, PrototypeMachine config,
     params.m_maxThrust = std::get<double>(config.m_config["thrust"]);
     params.m_specImpulse = std::get<double>(config.m_config["Isp"]);
 
-    MachineRocket::fuel_list_t inputs;
-    Package& pkg = m_scene.get_application().debug_find_package("lzdb");
-    DependRes<ShipResourceType> fuel = pkg.get<ShipResourceType>("fuel");
+    std::string fuelIdent = std::get<std::string>(config.m_config["fueltype"]);
```
jonesmz, line 252: bind temporary lifetime to const&

jonesmz: unless you're planning to std::move it somewhere later.

jonesmz: or at the very least, make it const.

src/adera/Machines/Rocket.cpp
```diff
@@ -247,32 +249,15 @@ Machine& SysMachineRocket::instantiate(ActiveEnt ent, PrototypeMachine config,
...
-            return EHierarchyTraverseStatus::Continue;
-        };
-        auto const& parent = m_scene.reg_get<ACompHierarchy>(ent);
-        m_scene.hierarchy_traverse(parent.m_parent, find_fuel);
-        // endTMP
-
-        ActiveEnt fuelSource = foundTank;
-        MachineRocket::ResourceInput input = {std::move(fuel), 1.0f, fuelSource};
-        inputs.push_back(std::move(input));
+        inputs.push_back({std::move(fuel), 1.0f});
```
jonesmz, line 260: use std::move on line 279/264

src/adera/Machines/Rocket.h
```diff
@@ -81,13 +81,21 @@ class SysMachineRocket :
 class MachineRocket : public osp::active::Machine
 {
     friend SysMachineRocket;
+
+    using input_t = std::pair<osp::DependRes<ShipResourceType>, float>;
     struct ResourceInput
     {
+        ResourceInput(osp::DependRes<ShipResourceType> type,
+            float massRateFraction, osp::active::WireInput&& source)
```
jonesmz, line 89: generally i discourage taking a parameter by &&, preferring to take by value or const& only, unless i have two versions, one for const& and one for &&.

If you want to deviate from that, I recommend providing a comment elaborating on why.

src/adera/Machines/Rocket.h
```diff
@@ -129,20 +137,28 @@ class MachineRocket : public osp::active::Machine
...
 }; // MachineRocket
 
-inline MachineRocket::MachineRocket(Parameters params, fuel_list_t& resources)
+inline MachineRocket::MachineRocket(Parameters params, std::vector<input_t> resources)
     : Machine(true)
     , m_params(params)
-    , m_resourceLines(std::move(resources))
-{ }
+{
+    for (input_t& input : resources)
```
jonesmz, line 153: I recommend signifying that you're planning on moving out of the items in the vector by using std::move() in the for loop like this:

for(input_t && input : std::move(resources))

or if that doesn't compile, then either


for(auto && input : std::move(resources))

or



for(input_t & input : std::move(resources))

jonesmz (approved):

src/adera/Machines/Rocket.cpp
z-adams: @Capital-Asterisk what do you think of this? I was just following the convention of the other `existing_inputs()` functions, should we change all of these to references instead of pointers?

Capital-Asterisk: I want the entire wiring system dead, so there isn't really a point. Changing the syntax won't untangle the spaghetti.

z-adams: Yeah fair enough

Capital-Asterisk (approved):

PR #69, opened 2021-02-13
Update EnTT submodule to v3.6.0 and Improve Package

Capital-Asterisk:
There's some new EnTT features look really useful, along with the code being generally outdated.

Changes:
* Fixed errors caused by version bump
* Reworked Package 'group types' to use EnTT sequential identifiers instead of a map. Accessing a resource now takes 1 less map access.
* Remove the use of groups in ActiveScene. This might have been causing 
* Added and removed a bunch of comments

### DIFF
--- a/src/osp/Resource/Package.h
+++ b/src/osp/Resource/Package.h
@@ -136,7 +136,7 @@ public:
         std::map<std::string, Resource<TYPE_T>, std::less<>> m_resources;
     };
 
-    constexpr ResPrefix_t const& get_prefix() { return m_prefix; }
+    constexpr ResPrefix_t const& get_prefix() const { return m_prefix; }
 
 private:
 
@@ -157,7 +157,7 @@ DependRes<TYPE_T> Package::add(std::string_view path,
     // First unique type added is 0, next is 1, then 2, etc...
     const uint32_t resTypeId = resource_id::type<TYPE_T>;
 
-    // Resize m_groups to ensure that resId a valid index
+    // Resize m_groups to ensure that resTypeId a valid index
     if (m_groups.size() <= resTypeId)
     {
         m_groups.resize(resTypeId + 1);
@@ -191,14 +191,14 @@ DependRes<TYPE_T> Package::add(std::string_view path,
 template<class TYPE_T>
 DependRes<TYPE_T> Package::get(std::string_view path) noexcept
 {
-    const uint32_t resId = resource_id::type<TYPE_T>;
+    const uint32_t resTypeId = resource_id::type<TYPE_T>;
 
-    if (m_groups.size() < resId)
+    if (m_groups.size() < resTypeId)
     {
-        return {}; // Return if resId is not a valid index to m_groups
+        return {}; // Return if resTypeId is not a valid index to m_groups
     }
 
-    entt::any &groupAny = m_groups[resId];
+    entt::any &groupAny = m_groups[resTypeId];
 
     if(groupAny.type() != entt::type_id< GroupType<TYPE_T> >())
     {
@@ -222,11 +222,11 @@ DependRes<TYPE_T> Package::get(std::string_view path) noexcept
 template<class TYPE_T>
 void Package::clear() noexcept
 {
-    const uint32_t resId = resource_id::type<TYPE_T>;
+    const uint32_t resTypeId = resource_id::type<TYPE_T>;
     //m_groups.erase(entt::type_info<TYPE_T>::id());
-    if (m_groups.size() > resId)
+    if (m_groups.size() > resTypeId)
     {
-        m_groups[resId] = entt::any{}; // Destruct GroupType
+        m_groups[resTypeId] = entt::any{}; // Destruct GroupType
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/osp/Resource/Package.h
```diff
@@ -90,121 +71,143 @@ StrViewPair_t decompose_str(std::string_view path, const char delim);
...
 
-    template<class TYPE_T>
-    struct GroupTypeRes : public GroupType
-    {
-        ~GroupTypeRes() = default;
-        //std::vector<TYPE_T> m_types;
         std::map<std::string, Resource<TYPE_T>, std::less<>> m_resources;
     };
 
-    ResPrefix_t get_prefix() const { return m_prefix; }
```
jonesmz, line 144: This should probably not be returning by value.

Be liberal in parameters you accept, and exacting in values you return.

Const& is my recommendation.

Also, the function is just a getter. Make it const.

src/osp/Resource/Package.h
```diff
@@ -90,121 +71,143 @@ StrViewPair_t decompose_str(std::string_view path, const char delim);
...
-    //std::any& groupAny(m_groups[entt::type_info<TYPE_T>::id()]);
-    std::unique_ptr<GroupType>& groupAny(m_groups[entt::type_info<TYPE_T>::id()]);
+    // Runtime generated (global) sequential ID
+    // First resource added is 0, next is 1, then 2, etc...
+    const uint32_t resId = resource_id::type<TYPE_T>;
 
+    // Resize m_groups to ensure that resId a valid index
+    if (m_groups.size() <= resId)
+    {
+        m_groups.resize(resId + 1);
```
jonesmz, line 163: Not a big fan of this. You're causing a new alloction each time. No doubling algorithm for the capacity. 

Use push_back instead? Or at least a call to reserve before resize (which does the doubling algom)

src/osp/Resource/Package.h
```diff
@@ -217,9 +220,14 @@ DependRes<TYPE_T> Package::get(std::string_view path)
...
 template<class TYPE_T>
-void Package::clear()
+void Package::clear() noexcept
 {
-    m_groups.erase(entt::type_info<TYPE_T>::id());
+    const uint32_t resId = resource_id::type<TYPE_T>;
+    //m_groups.erase(entt::type_info<TYPE_T>::id());
+    if (m_groups.size() > resId)
+    {
+        m_groups[resId] = entt::any{}; // Destruct GroupType
```
jonesmz, line 229: Why not resize to one smaller?

src/osp/Resource/Resource.h
```diff
@@ -80,8 +80,16 @@ namespace osp
 template <class TYPE_T>
 struct Resource
 {
-    Resource(bool loaded) : m_data(), m_loaded(loaded), m_refCount(0) {}
-    Resource(bool loaded, TYPE_T&& data) :  m_data(std::move(data)), m_loaded(loaded), m_refCount(0) {}
+    Resource(bool loaded)
+     : m_data()
```
jonesmz, line 84: Don't needed to specify that you are default constructing the member variables.

Put the default value of recount in the class body, not the constructor.

jonesmz (approved):

src/osp/Resource/Package.h
Capital-Asterisk: These IDs are generated per-type. This is called very few times when types are first added to the Package. EnTT does a similar thing with its pools on the inside. Maybe I should change resId to resTypeId, and just make this more clear in general.

src/osp/Resource/Package.h
Capital-Asterisk: There is other groups with a later index still stored in there. Only a single type is being removed.

src/osp/Resource/Package.h
jonesmz: My point was about the reallocations of the vector, since you're circumventing the allocators doubling heuristic.

No way to use push_back for this?

src/osp/Resource/Package.h
```diff
@@ -136,7 +136,7 @@ class Package
         std::map<std::string, Resource<TYPE_T>, std::less<>> m_resources;
     };
 
-    ResPrefix_t get_prefix() { return m_prefix; }
+    constexpr ResPrefix_t const& get_prefix() { return m_prefix; }
```
jonesmz, line 139: constexpr functions aren't, inherently, const. You'll want to add the trailing const keyword as well.

src/osp/Resource/Resource.h
```diff
@@ -30,75 +30,27 @@
...
      : m_data()
      , m_loaded(loaded)
-     , m_refCount(0)
+     , m_refCount()
     { }
     Resource(bool loaded, TYPE_T&& data)
      : m_data(std::move(data))
      , m_loaded(loaded)
-     , m_refCount(0)
+     , m_refCount()
```
jonesmz, line 44: this default initializes m_refCount to whatever m_refCounts type's default is.

You'll need to remove this line entirely to make the class-body default actually apply.

z-adams (approved):
Looks good to me. Does this allow us to use the fancy iteration scheme that skypjack mentioned in that issue?

src/osp/Resource/Package.h
Capital-Asterisk: Nope, resId can be any number above the current capacity of the m_groups vector. It's possible to resize to a slightly higher capacity, but this is a minor optimization anyways.

PR #72, opened 2021-02-14
Add Release Artifact to GitHub CI workflow

Capital-Asterisk:
This tells the CI to upload binaries after they compile for both Windows and Linux. I'm not quite sure if I'm doing this properly. 

Windows version is tested to work in wine.

Linux version needs to set bin/osp-magnum as executable, then in bin run `LD_LIBRARY_PATH=../lib osp-magnum`


jonesmz (approved):
It's annoying that the Linux artefacts aren't able to run out of the box.

For the LD_LIBRARY_PATH problem, we can solve it in 2 ways.

1) Set the rpath of the binary as a post-compile step in the cmake config.
2) Static linking.

Neither of those are really a 2 minute task, so this is fine as is.

PR #73, opened 2021-02-14
Use entt type name

jonesmz:
This PR includes https://github.com/TheOpenSpaceProgram/osp-magnum/pull/69, and should only be looked at after 69 is merged.

Instead of manually defining static const std::string = "", use the entt::type_name type instead.

### DIFF
--- a/src/adera/Machines/RCSController.h
+++ b/src/adera/Machines/RCSController.h
@@ -45,8 +45,6 @@ class SysMachineRCSController :
     public osp::active::SysMachine<SysMachineRCSController, MachineRCSController>
 {
 public:
-    static inline std::string smc_name = "RCSController";
-
     SysMachineRCSController(osp::active::ActiveScene &rScene);
 
     /**
@@ -135,3 +133,14 @@ inline MachineRCSController& MachineRCSController::operator=(MachineRCSControlle
 }
 
 } // adera::active::machines
+
+namespace entt
+{
+    template<>
+    struct type_name<adera::active::machines::SysMachineRCSController>
+    {
+        [[nodiscard]] static constexpr std::string_view value() noexcept {
+            return "RCSController";
+        }
+    };
+} // namespace entt
--- a/src/adera/Machines/Rocket.h
+++ b/src/adera/Machines/Rocket.h
@@ -39,9 +39,6 @@ class SysMachineRocket :
         public osp::active::SysMachine<SysMachineRocket, MachineRocket>
 {
 public:
-
-    static inline std::string smc_name = "Rocket";
-
     SysMachineRocket(osp::active::ActiveScene &scene);
 
     //void update_sensor();
@@ -169,3 +166,14 @@ inline MachineRocket& MachineRocket::operator=(MachineRocket&& move) noexcept
 }
 
 } // namespace adera::active::machines
+
+namespace entt
+{
+    template<>
+    struct type_name<adera::active::machines::SysMachineRocket>
+    {
+        [[nodiscard]] static constexpr std::string_view value() noexcept {
+            return "Rocket";
+        }
+    };
+} // namespace entt
--- a/src/adera/Machines/UserControl.cpp
+++ b/src/adera/Machines/UserControl.cpp
@@ -32,8 +32,6 @@ using namespace adera::active::machines;
 using namespace osp::active;
 using namespace osp;
 
-const std::string SysMachineUserControl::smc_name = "UserControl";
-
 void MachineUserControl::propagate_output(WireOutput* output)
 {
     std::cout << "propagate test: " << output->get_name() << "\n";
--- a/src/adera/Machines/UserControl.h
+++ b/src/adera/Machines/UserControl.h
@@ -40,9 +40,6 @@ class SysMachineUserControl :
         public osp::active::SysMachine<SysMachineUserControl, MachineUserControl>
 {
 public:
-
-    static const std::string smc_name;
-
     SysMachineUserControl(osp::active::ActiveScene &scene,
                           osp::UserInputHandler& userControl);
 
@@ -125,3 +122,14 @@ inline MachineUserControl& MachineUserControl::operator=(MachineUserControl&& mo
 }
 
 } // namespace adera::active::machines
+
+namespace entt
+{
+    template<>
+    struct type_name<adera::active::machines::SysMachineUserControl>
+    {
+        [[nodiscard]] static constexpr std::string_view value() noexcept {
+            return "UserControl";
+        }
+    };
+} // namespace entt
--- a/src/adera/ShipResources.h
+++ b/src/adera/ShipResources.h
@@ -117,8 +117,6 @@ class SysMachineContainer
     : public osp::active::SysMachine<SysMachineContainer, MachineContainer>
 {
 public:
-    static inline std::string smc_name = "Container";
-
     SysMachineContainer(osp::active::ActiveScene& rScene);
 
     static void update_containers(osp::active::ActiveScene& rScene);
@@ -205,3 +203,14 @@ inline MachineContainer& MachineContainer::operator=(MachineContainer&& move) no
 }
 
 } // namespace adera::active::machines
+
+namespace entt
+{
+    template<>
+    struct type_name<adera::active::machines::SysMachineContainer>
+    {
+        [[nodiscard]] static constexpr std::string_view value() noexcept {
+            return "Container";
+        }
+    };
+} // namespace entt
--- a/src/adera/SysExhaustPlume.h
+++ b/src/adera/SysExhaustPlume.h
@@ -40,8 +40,6 @@ struct ACompExhaustPlume
 class SysExhaustPlume : public IDynamicSystem
 {
 public:
-    static inline std::string smc_name = "ExhaustPlume";
-
     SysExhaustPlume(ActiveScene& rScene);
     ~SysExhaustPlume() = default;
 
@@ -64,3 +62,14 @@ private:
 };
 
 } // namespace osp::active
+
+namespace entt
+{
+    template<>
+    struct type_name<osp::active::SysExhaustPlume>
+    {
+        [[nodiscard]] static constexpr std::string_view value() noexcept {
+            return "ExhaustPlume";
+        }
+    };
+} // namespace entt
--- a/src/osp/Active/ActiveScene.h
+++ b/src/osp/Active/ActiveScene.h
@@ -275,7 +275,7 @@ private:
 template<class SYSMACH_T, typename... ARGS_T>
 void ActiveScene::system_machine_create(ARGS_T &&... args)
 {
-    system_machine_add(SYSMACH_T::smc_name,
+    system_machine_add(entt::type_name<SYSMACH_T>::value(),
                        std::make_unique<SYSMACH_T>(*this, args...));
 }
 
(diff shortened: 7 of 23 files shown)


z-adams (approved):

Capital-Asterisk:
Won't this restrict access to the real value of `entt::type_name`? If a type was forward declared and the header with the specialization is not included, would it return the same value? It also becomes ambiguous which names are platform-consistent constants, and which ones are compiler-generated. I don't see how this makes anything more convenient.

jonesmz:
Specializing the type_name template makes the provided specialization the ONLY definition of type_name for that type across the whole program. without regard to forward declarations.

I'm just as fine providing no specialization at all. Unless you plan to use what uses to be smc_name as a config file identifier, in which case providing concrete type_name specializations for those is the right approach.

Capital-Asterisk:
I guess the use of smc_name seems excessive throughout the codebase. The only time they're really needed are to identify Machines from config files, and possibly multiplayer setup; the rest are on the chopping block. Dynamic systems are not needed, and Satellite Types are already removed in my next PR.

How is specializing entt type the 'right approach?' I haven't been given any reason why this is more correct. Is this for consistency? Is a string constant not already a simple and non-ambiguous solution?

jonesmz:
> How is specializing entt type the 'right approach?' I haven't been given any reason why this is more correct. Is this for consistency? Is a string constant not already a simple and non-ambiguous solution?

The `entt::type_info` collection of tools is more consistent than a manually specified static string, yes. But it also comes out of the box with a runtime determined type sequence, and a compile time determined hash of the type name, both of which can be overridden for a particular type.

This allows us to build a map between a type name, and it's hash or sequence. In the case of the hash, we can just calculate that directly, and in the case of the type sequence, we can build the map at runtime in the main function. From that point, any configuration files that need to be able to name specific things can have a single string-comparison or hash happen to get an integer ID that can be compared in maps directly. I haven't dug deep enough to confirm this, but I suspect that a Component in EnTT's ECS also ends up mapping back to the `entt::type_info`, which could allow for some code to be simplified by doing a one-time lookup of what the configuration file says the type name is, and then carry around a typeid thereafter.

There's also the aspect of "dont repeat yourself"

jonesmz:
My recommendation is that in main(), we call some function that builds a std::map<entt::type_name, entt::type_id> of all of the "config file API" types.

When a config file is parsed, if it has an identifier that is not found in that map, abort parsing and consider that config file as bad.

This gives us a single place where specializations for config-file-api types need be given a name, (and should work just fine with forward declarations, even), and a single place to check if a config file is naming a type that isn't part of the public API.

All other types have no need for platform-stable names. Only public API types. So any other type having an unspecialized entt::type_name<> is no problem.

Capital-Asterisk:
> The `entt::type_info` collection of tools is more consistent than a manually specified static string, yes. But it also comes out of the box with a runtime determined type sequence, and a compile time determined hash of the type name, both of which can be overridden for a particular type.

Any string can be already be hashed, it doesn't have to be a type name. It's a nice feature to have a pretty-printed type name, but specializing it for a string constant is removing the feature instead of utilizing it.

One useful entt::type_name specialization I can think of, is to add additional information to the name, such as it's EnTT runtime ID, whether it's copy constructable, or its sizeof.

> This allows us to build a map between a type name, and it's hash or sequence. In the case of the hash, we can just calculate that directly, and in the case of the type sequence, we can build the map at runtime in the main function. From that point, any configuration files that need to be able to name specific things can have a single string-comparison or hash happen to get an integer ID that can be compared in maps directly. I haven't dug deep enough to confirm this, but I suspect that a Component in EnTT's ECS also ends up mapping back to the `entt::type_info`, which could allow for some code to be simplified by doing a one-time lookup of what the configuration file says the type name is, and then carry around a typeid thereafter.

Using entt::type_name isn't needed here. Machines already maps config names to instances of ISysMachine using smc_name, working just fine (ISysMachine will probably be replaced with function pointers soon). Resolving the name into an integer ID on load sounds like a good idea; right now, every part created has to access the string map of ISysMachines. entt::family is perfect for generating runtime IDs, which are also sequential so a vector can be used instead of map. EnTT uses entirely vectors for the core of ECS; maps are only used in meta, snapshot, and resource features (Verified with `grep -r "map"`).

> There's also the aspect of "dont repeat yourself"

How is copying an entire namespace and template specification less repetitive than a single string constant? This is an apparent increase in consistency for the cost of ambiguity and more complicated code.

Instead of making it more ambiguous, a better option would be to make the difference between entt::type_name and smc_name more obvious, renaming "smc_name" to something like "smc_configName". smc_name and entt::type_name just don't serve the same purpose, and should be differentiated from each other.

> My recommendation is that in main(), we call some function that builds a std::map<entt::type_name, entt::type_id> of all of the "config file API" types.
> 
> When a config file is parsed, if it has an identifier that is not found in that map, abort parsing and consider that config file as bad.
> 
> This gives us a single place where specializations for config-file-api types need be given a name, (and should work just fine with forward declarations, even), and a single place to check if a config file is naming a type that isn't part of the public API.
> 
> All other types have no need for platform-stable names. Only public API types. So any other type having an unspecialized entt::type_name<> is no problem.

I'm pretty sure you mean storing an `entt::type_info` instead. I'm not quite sure what "config file API" includes, but a separate map for each base type is more practical. Machine map will only map machines; a system map will only map systems. This also allows each map to store an ID that belong to a unique entt::family. A reverse map/vector may also be needed, to convert IDs back to type names when writing to config files.

jonesmz:
> Any string can be already be hashed, it doesn't have to be a type name. It's a nice feature to have a pretty-printed type name, but specializing it for a string constant is removing the feature instead of utilizing it.

On the contrary, this is the intended use of the entt::type_name functionality, it explicitly supports being specialized for the concerns you brought up of cross-compiler, cross-platform, cross-version, and cross-network stability.

The entt::type_*** infrastructure is intended to be the single place to maintain information about types, with regards to unique ids and names mapped to C++ symbols.

> One useful entt::type_name specialization I can think of, is to add additional information to the name, such as it's EnTT runtime ID, whether it's copy constructable, or its sizeof.

I don't understand why you would inject that information into the name of the type. You can query that information with existing C++ language level tools, or if you really want to provide some type-erased entt::type_id -> type_traits details (because perhaps you no longer have access to the actual C++ symbol) the place for that would be in a std::map built at runtime similar to how entt::type_seq works.

> Using entt::type_name isn't needed here. Machines already maps config names to instances of ISysMachine using smc_name, working just fine (ISysMachine will probably be replaced with function pointers soon). Resolving the name into an integer ID on load sounds like a good idea; right now, every part created has to access the string map of ISysMachines. entt::family is perfect for generating runtime IDs, which are also sequential so a vector can be used instead of map. EnTT uses entirely vectors for the core of ECS; maps are only used in meta, snapshot, and resource features (Verified with `grep -r "map"`).

Ok. I don't really have a response to this paragraph, since it doesn't really apply to my PR, but I'm happy to know what your plans are.

> How is copying a namespace and template specification less repetitive than a single string constant?

I've already explained how adding this specialization in multiple different files wasn't the approach I thought was ideal. Better would be to have that single config_file_api.h file that provides this mapping. I was attempting to provide an example of how to use the specialization in a way that could be easily compared to the current code.

If we centralized these entt::type_name specializations into a single file, a macro turns the ugly specializations into one-liners with clear and concise meaning.

>  This is an apparent increase in consistency for the cost of ambiguity and more complicated code.

It's not ambiguous at all. What ambiguity is there?

The code is not more complicated, there's just more code. Arguably, having an smc_name member is far more complicated from a template metaprogramming perspective due to issues with SFINAE.

> Instead of making it more ambiguous, a better option would be to make the difference between entt::type_name and smc_name more obvious, renaming "smc_name" to something like "smc_configName". smc_name and entt::type_name just don't serve the same purpose, and should be differentiated from each other.

I don't agree that there should exist a static member variable to solve this problem.

I'm fine with entt::type_name being distinct and different from some hypothetical osp::config_name<> lookup system with similar usage semantics to entt::type_name, but I definitely do not agree with using a static member variable.

Better still than even using entt::type_name would be a compile-time map of "config names <-> types <-> hand-maintained identifiers", so that configuration files can be used to lookup the identifier to use at runtime, and code that needs to map from a C++ symbol to an identifier can do so as well. But we shouldn't be using strings as the key for looking things up in the guts of the engine. Instead, strings should be used only when information enters or leaves the engine (e.g. config files, network traffic, so on)

> > My recommendation is that in main(), we call some function that builds a std::map<entt::type_name, entt::type_id> of all of the "config file API" types.
> > When a config file is parsed, if it has an identifier that is not found in that map, abort parsing and consider that config file as bad.
> > This gives us a single place where specializations for config-file-api types need be given a name, (and should work just fine with forward declarations, even), and a single place to check if a config file is naming a type that isn't part of the public API.
> > All other types have no need for platform-stable names. Only public API types. So any other type having an unspecialized entt::type_name<> is no problem.
> 
> I'm pretty sure you mean storing an `entt::type_info` instead. I'm not quite sure what "config file API" includes, but a separate map for each base type is more practical. Machine map will only map machines; a system map will only map systems. This also allows each map to store an ID that belong to a unique entt::family. A reverse map/vector may also be needed, to convert IDs back to type names when writing to config files.

I meant entt::type_id, as in the integral type. But an entt::type_info is also fine.

A config file API would encompass, at a minimum,

- Encoding scheme (presumable UTF-8)
- Markup - E.g. json, yaml, xml, so on.
- Structure - Things like identifers, hierarchies, values, meaning of particular sequences

Inside the "Structure" aspect, you need to have a well defined set of (among other things) identifiers. So if you're providing a mechanism to allow the configuration file to name concepts that exist inside the C++ code (or inside dynamically loaded plugins, as well), you need a place where you inform the configuration system of what those names are, and how to map a parsed "textual identifier" to an internal handle of some kind. Ideally an integral type, for speedy lookups.

Until we grow the ability to dynamically load plugins or execute interpreted script code, all "type name" information is known at compile time, but my intention is to indicate that the typename information be stored in a runtime-queryable format. E.g. instead of a configuration file doing:

```cpp
if(text == ThingA)
else if(text == ThingB)
```

We would instead have one of:
```cpp
if(compileTimeLookupTable.find(text))
{
    // Yay, quick outcome.
}
else if(runtimeLookupTable.find(text))
{
    // Yay, still found it.
}
else
{
    std::abort();
}
```
or
```cpp
if(typeNameToId.find(text))
{
    // Yay
}
else
{
    std::abort();
}
```

Note that I do strongly recommend separating true compile-time type searching from runtime-only type searching. My reason for this is that containers like std::map() can't be optimized all that well for situations where you know the full contents of the map ahead of time, but compile time storage can be optimized down to very fast searches. So you'd want to execute your search on the compile time type table first, and then fall back to the runtime type table, because that will likely be quicker than combining them into a single run time table.

Capital-Asterisk:
I understand what the entt type system is for, but I think the only thing I forgot to clarify here is that smc_name is NOT a type name. Even though the same variable name "smc_name" is used across unrelated classes, they serve different purposes. Note how SysVehicle::smc_name and SatVehicle::smc_name are both "Vehicle"; they don't conflict because they're emplaced into different unrelated maps.

jonesmz:
Ok.

Then lets make a new system. Some osp::config_file_name<> or similar.

z-adams:
What's the conclusion here? Should this PR be closed?

jonesmz:
Closing due to lack of consensus.

PR #88, opened 2021-02-28
Update test universes and DebugCameraController for distant vehicles

Capital-Asterisk:
* Increased default activation radius of planets to 256x their radius. This stops the planet from disappearing at a short distance.
* DebugCameraController was modified to switch between vehicle satellites instead of only already-activated vehicles. This allows switching between very distant vehicles.
* Increased distance between planets in simple test universe
* Replaced moon test vehicles with 4 part vehicles instead of spamcans
* Added roll control to part vehicles by adding RCS "blocks" consisting of two RCS thrusters
* A few other minor cleanups

### DIFF
--- a/src/planet-a/Satellites/SatPlanet.cpp
+++ b/src/planet-a/Satellites/SatPlanet.cpp
@@ -37,8 +37,11 @@ UCompPlanet& SatPlanet::add_planet(
         osp::universe::Universe& rUni, Satellite sat, double radius, float mass,
         float resolutionSurfaceMax, float resolutionScreenMax)
 {
+
+    float activateRadius = float(radius) * 256.0f;
+
     rUni.get_reg().emplace<UCompActivatable>(sat);
-    rUni.get_reg().emplace<UCompActivationRadius>(sat, float(radius));
+    rUni.get_reg().emplace<UCompActivationRadius>(sat, activateRadius);
 
     return rUni.get_reg().emplace<UCompPlanet>(
                 sat, radius, resolutionSurfaceMax, resolutionScreenMax, mass);
--- a/src/test_application/universes/planets.cpp
+++ b/src/test_application/universes/planets.cpp
@@ -62,57 +62,45 @@ void testapp::create_real_moon(osp::OSPApplication& ospApp)
     auto &stationary = rUni.trajectory_create<TrajStationary>(
                                         rUni, rUni.sat_root());
 
-    // Create 10 random vehicles
-    for (int i = 0; i < 10; i ++)
+    // Create 4 part vehicles
+    for (int i = 0; i < 4; i ++)
     {
-        // Creates a random mess of spamcans as a vehicle
-        Satellite sat = debug_add_random_vehicle(rUni, rPkg, "TestyMcTestFace Mk"
-                                                 + std::to_string(i));
+        Satellite sat = testapp::debug_add_part_vehicle(rUni, rPkg, "Placeholder Mk. I " + std::to_string(i));
 
         auto &posTraj = rUni.get_reg().get<UCompTransformTraj>(sat);
 
-        posTraj.m_position = osp::Vector3s(i * 1024l * 5l, 0l, 0l);
+        posTraj.m_position = osp::Vector3s(i * 1024l * 4l, 0l, 0l);
         posTraj.m_dirty = true;
 
         stationary.add(sat);
     }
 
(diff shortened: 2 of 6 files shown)


src/planet-a/Satellites/SatPlanet.cpp
```diff
@@ -37,8 +37,11 @@ UCompPlanet& SatPlanet::add_planet(
         osp::universe::Universe& rUni, Satellite sat, double radius, float mass,
         float resolutionSurfaceMax, float resolutionScreenMax)
 {
+
+    float activateRadius = float(radius) * 256.0f;
+
     rUni.get_reg().emplace<UCompActivatable>(sat);
-    rUni.get_reg().emplace<UCompActivationRadius>(sat, float(radius));
+    rUni.get_reg().emplace<UCompActivationRadius>(sat, activateRadius);
```
jonesmz, line 44: Is the downcast to float intentional for perf reasons?

I'm skeptical that its any faster than doing the calculation as double.

src/test_application/universes/vehicles.cpp
```diff
@@ -223,38 +259,39 @@ osp::universe::Satellite testapp::debug_add_part_vehicle(
...
+                                    rotZ * rotY_090);
+        }
+    }
 
     enum Parts
     {
         CAPSULE = 0,
         FUSELAGE = 1,
-        ENGINE = 2,
-        RCS1 = 3,
```
jonesmz, line 247: Recommend making this an enum class, BTW.

src/test_application/universes/vehicles.cpp
```diff
@@ -223,38 +259,39 @@ osp::universe::Satellite testapp::debug_add_part_vehicle(
...
-    blueprint.add_part(rcs, rotZ_270.transformVector(rcsOsetTop), yz270, scl);
+    std::vector<int> rcsPorts;
 
-    // Top RCS ring
-    blueprint.add_part(rcs, rcsOsetBtm, rotY_090, scl);
-    blueprint.add_part(rcs, rotZ_090.transformVector(rcsOsetBtm), yz090, scl);
-    blueprint.add_part(rcs, rotZ_180.transformVector(rcsOsetBtm), yz180, scl);
-    blueprint.add_part(rcs, rotZ_270.transformVector(rcsOsetBtm), yz270, scl);
+    for (float z = rcsZMin; z < rcsZMax; z += rcsZStep)
+    {
```
jonesmz, line 273: Accuracy issues no big deal here? Guess not, but just double checking.

src/test_application/universes/vehicles.cpp
```diff
@@ -223,38 +259,39 @@ osp::universe::Satellite testapp::debug_add_part_vehicle(
 
     auto& engBP = blueprint.add_part(engine, cfOset + feOset, idRot, scl);
 
-    Quaternion yz090 = rotZ_090 * rotY_090;
-    Quaternion yz180 = rotZ_180 * rotY_090;
-    Quaternion yz270 = rotZ_270 * rotY_090;
+    // Add a shit ton of RCS rings
+
+    Rad rcsRingStep     = 90.0_degf;
```
jonesmz, line 264: Make this variable constexpr?

jonesmz (approved):

src/test_application/universes/vehicles.cpp
z-adams: Enum class is sort of annoying for this because these are supposed to just be named indices, so you'd need `static_cast<int>()` everywhere you want to use the enum. I think it's fine as-is since it's just local to this function scope.

z-adams (approved):

src/test_application/universes/planets.cpp
```diff
@@ -62,57 +62,45 @@ void testapp::create_real_moon(osp::OSPApplication& ospApp)
...
-        }
-    }
+    Satellite moonSat = rUni.sat_create();
+
+    // Create the real world moon
+    float radius = 1.737E+6;
+    float mass = 7.347673E+22;
+
+    float resolutionScreenMax = 0.056f;
+    float resolutionSurfaceMax = 12.0f;
```
EhWhoAmI, line 94: Should these 4 be const?

EhWhoAmI:
Aside from some nitpicky stuff, looks good to me

EhWhoAmI (approved):

Issue #42, opened 2021-01-08
Use DependRes's ability to store a not-yet-loaded resource in PlumeEffect; PrototypePart


z-adams:
Ideally this [PlumeEffectData::meshName] should be `DependRes`. I'm aware this does have the issue of meshes not being added to the Package yet, but there is a not-yet-implemented solution to this. Note that `osp::Resource<>` has an `m_loaded` bool variable. This is intended to be set false for placeholder resources that are initialized empty. If something needs a resource that isn't loaded yet, it can initialize an empty one, where it can be loaded later.

_Originally posted by @Capital-Asterisk in https://github.com/TheOpenSpaceProgram/osp-magnum/pull/36#discussion_r552235308_

Capital-Asterisk:
Added in #90

PR #90, opened 2021-03-05
Add reserved Resources to Package and Add list_pkg CLI command

Capital-Asterisk:
This PR adds support for reserving placeholders in an `osp::Package` for resources that aren't loaded yet (see issue #42).

`osp::Package` maps any type (referred to as a resource) to a string path identifier, by creating an `std::map` for each unique type. These maps store the resource as a member of `osp::Resource<T>` to handle reference counting.

`osp::Resource<T>` had an unused `m_loaded` member intended to handle placeholders, but it isn't adequate as the resource type must still be constructed. Some resources like `Magnum::Trade::ImageData2D` are not default-constructable. Instead of storing the type as a member directly, this PR changes it to an `std::optional` so placeholders can be created without constructing the resource.

If a config refers to a resource that isn't loaded yet, then it can now use `Package::get_or_reserve(path)` to create a non-initialized placeholder with the path, where it can be loaded in a later step.

These changes are cherry-picked from a branch where I'm doing some major changes to how vehicles are loaded.

There is also the `list_pkg` command and a few other minor changes added to main.cpp.

### DIFF
--- a/src/osp/Resource/Package.h
+++ b/src/osp/Resource/Package.h
@@ -82,6 +82,9 @@ class Package
 
 public:
 
+    template<class TYPE_T>
+    using group_t =  std::map<std::string, osp::Resource<TYPE_T>, std::less<>>;
+
     Package(std::string prefix, std::string packageName);
     Package(Package&& move) = default;
     Package(const Package& copy) = delete;
@@ -111,6 +114,21 @@ public:
     template<class TYPE_T>
     DependRes<TYPE_T> get(std::string_view path) noexcept;
 
+    /**
+     * Get a resource by path identifier. If it isn't found, then reserve the
+     * path to be loaded later.
+     *
+     * @tparam TYPE_T Type of resource to get
+     *
+     * @param path [in] String used to identify the target resource
+     * @return Reference counted dependency to the found or unloaded resource
+     */
+    template<class TYPE_T>
+    DependRes<TYPE_T> get_or_reserve(std::string_view path) noexcept;
+
+    template<class TYPE_T>
+    group_t<TYPE_T> const& group_get();
+
     // TODO: function for removing specific resources
 
     /**
@@ -133,7 +151,7 @@ public:
         // entt experimental branch allows move-only types.
         GroupType(GroupType const& copy) { assert(0); };
 
-        std::map<std::string, Resource<TYPE_T>, std::less<>> m_resources;
+        group_t<TYPE_T> m_resources;
     };
 
     constexpr ResPrefix_t const& get_prefix() const { return m_prefix; }
@@ -175,17 +193,30 @@ DependRes<TYPE_T> Package::add(std::string_view path,
 
     GroupType<TYPE_T> &group = entt::any_cast< GroupType<TYPE_T>& >(groupAny);
 
+    // Check if resource already exists
+    auto resIt = group.m_resources.find(path); // Find resource by path
+
+    if (resIt != group.m_resources.end())
+    {
+        // Resource already exists, replace if not loaded
+        if (!resIt->second.m_data.has_value())
+        {
+            resIt->second.m_data.emplace(std::forward<ARGS_T>(args)...);
+        }
+        return {};
+    }
+
     // Emplace without needing copy constructor
-    auto final = group.m_resources
-        .try_emplace(std::string{path}, false, std::forward<ARGS_T>(args)...);
+    auto newIt = group.m_resources
+        .try_emplace(std::string{path}, construct_tag(), std::forward<ARGS_T>(args)...);
 
     // check if emplace fails, Resource already exists
-    if (!final.second)
+    if (!newIt.second)
     {
         return {}; // return empty if so
     }
 
-    return DependRes<TYPE_T>(final.first);
+    return DependRes<TYPE_T>(newIt.first);
 }
 
 template<class TYPE_T>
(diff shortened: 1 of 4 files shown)


src/osp/Resource/Resource.h
```diff
@@ -24,29 +24,37 @@
...
-    Resource(bool loaded)
+    Resource(reserve_tag)
      : m_data()
-     , m_loaded(loaded)
     { }
-    Resource(bool loaded, TYPE_T&& data)
+    Resource(construct_tag)
+     : m_data(TYPE_T())
+    { }
+    Resource(construct_tag, TYPE_T&& data)
```
Capital-Asterisk, line 47: This is my first time using a tag dispatch; I kind of don't know what I'm doing. `Resource()` became ambiguous, which can either refer to default-constructing TYPE_T, or initializing std::optional as null.

src/osp/Resource/Resource.h
```diff
@@ -24,29 +24,37 @@
...
      : m_data(std::move(data))
-     , m_loaded(loaded)
     { }
     Resource(Resource&& move) = default;
     Resource(const Resource& copy) = delete;
+    ~Resource()
+    {
+        assert(m_refCount == 0);
```
Capital-Asterisk, line 54: I forgot I put this here. It shouldn't cause any issues if DependRes is doing it's job properly

z-adams:
Does any of this change the existing way we interact with packages and resources, or does it just add to it?

src/test_application/main.cpp
```diff
@@ -106,7 +108,7 @@ int main(int argc, char** argv)
 
     load_a_bunch_of_stuff();
 
-    if( ! args.value("scene").empty())
+    if(args.value("scene") != "none")
```
Capital-Asterisk, line 111: `!args.value("scene").empty()` would always return true, as "simple" is set as the default value. Doing `--scene none` would just show the CLI.

src/test_application/main.cpp
```diff
@@ -309,13 +317,61 @@ void debug_print_help()
...
+        debug_print_resource_group<osp::BlueprintVehicle>(rPkg);
+
+        debug_print_resource_group<Magnum::Trade::ImageData2D>(rPkg);
+        debug_print_resource_group<Magnum::Trade::MeshData>(rPkg);
+        debug_print_resource_group<Magnum::GL::Texture2D>(rPkg);
+        debug_print_resource_group<Magnum::GL::Mesh>(rPkg);
+
+        debug_print_resource_group<adera::active::machines::ShipResourceType>(rPkg);
+        debug_print_resource_group<adera::shader::PlumeShader>(rPkg);
+        debug_print_resource_group<adera::shader::Phong>(rPkg);
```
Capital-Asterisk, line 371: I guess it's possible to add some polymorphic functionality to Package to list resource types, but this works for now.

src/test_application/main.cpp
```diff
@@ -309,13 +317,61 @@ void debug_print_help()
...
+    if (group.empty())
+    {
+        return;
+    }
+
+    std::cout << "  * TYPE: " << entt::type_name<RES_T>().value() << "\n";
+
+    for (auto const& [key, resource] : group)
+    {
+        std::cout << "    * " << (resource.m_data.has_value() ? "LOADED" : "RESERVED") << ": " << key << "\n";
```
Capital-Asterisk, line 342: Now that I think of it, the reference count can be printed here too.

Also, sample output:
```
* PACKAGE: lzdb
  * TYPE: osp::PrototypePart
    * LOADED: part_phCapsule
    * LOADED: part_phEngine
    * LOADED: part_phFuselage
    * LOADED: part_phLinRCS
    * LOADED: part_spamcan
    * LOADED: part_stomper
  * TYPE: Magnum::Trade::ImageData<2>
    * LOADED: Material Metallic
    * LOADED: OSPData/adera/noise1024.png
    * LOADED: OSPData/adera/noise256.png
    * LOADED: ph_base
    * LOADED: spamcanhd
    * LOADED: stomper
    * LOADED: stomper_metal-stomper_rough
  * TYPE: Magnum::Trade::MeshData
    * LOADED: OSPData/adera/ph_capsule:Cylinder.001
    * LOADED: OSPData/adera/ph_engine:Cylinder.002
    * LOADED: OSPData/adera/ph_fuselage:Cylinder
    * LOADED: OSPData/adera/ph_plume:Cylinder.003
    * LOADED: OSPData/adera/ph_rcs:Cube
    * LOADED: OSPData/adera/ph_rcs_plume:Cylinder.005
    * LOADED: OSPData/adera/spamcan:Cylinder.001
    * LOADED: OSPData/adera/stomper:Cylinder.002
  * TYPE: adera::active::machines::ShipResourceType
    * LOADED: fuel

```

Capital-Asterisk:
> Does any of this change the existing way we interact with packages and resources, or does it just add to it?

None of the existing functions have changed in the current cases that we're using them in. No placeholders are ever created so far. However, DependRes right now doesn't care if it's referring to a placeholder or not; it would be referring to an empty `std::optional` which would be an issue.

z-adams (approved):

src/osp/Resource/Package.h
```diff
@@ -175,17 +193,30 @@ DependRes<TYPE_T> Package::add(std::string_view path,
...
+            resIt->second.m_data.emplace(std::forward<ARGS_T>(args)...);
+        }
+        return {};
+    }
+
     // Emplace without needing copy constructor
-    auto final = group.m_resources
-        .try_emplace(std::string{path}, false, std::forward<ARGS_T>(args)...);
+    auto newIt = group.m_resources
+        .try_emplace(std::string{path}, construct_tag(), std::forward<ARGS_T>(args)...);
```
jonesmz, line 211: It may not make sense to have the parameter to the function Package::add take a std::string_view if you're always going to allocate a new string when you call try_emplace.

src/osp/Resource/Resource.h
jonesmz: I recommend collapsing the two construct_tag constructors into one, with TYPE_T&& data = {}.

src/osp/Resource/Resource.h
jonesmz: I don't understand what this reference count is doing, if nothing in the Resource class modifies the reference count?

jonesmz (approved):

src/osp/Resource/Package.h
```diff
@@ -211,14 +242,74 @@ DependRes<TYPE_T> Package::get(std::string_view path) noexcept
...
+    }
+
+    GroupType<TYPE_T> &group = entt::any_cast<GroupType<TYPE_T>&>(groupAny);
+
+    auto resIt = group.m_resources.find(path); // Find resource by path
+
+    if (resIt == group.m_resources.end())
+    {
+        // Not found, reserve with m_data as empty optional
+        auto newIt = group.m_resources.try_emplace(std::string{path},
```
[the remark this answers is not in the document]
Capital-Asterisk: There was an issue with `emplace` requiring a copy constructor somehow. `try_emplace` can add move-only types to the map.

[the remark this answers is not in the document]
jonesmz: Sure. But that doesn't address any of my questions ;)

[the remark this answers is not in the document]
Capital-Asterisk: Oh I see, this branch just isn't needed at all. About the string views, I'll go with the template parameter since the function is already a template.

src/osp/Resource/Resource.h
Capital-Asterisk: The ref count is modified by DependRes. This assert is making sure that there is no references to the resource when it's destructed.

src/osp/Resource/Resource.h
Capital-Asterisk: I'll combine the constructors, but I'll keep the tags as-is. I'm not sure if I fully understand what you mean, but I feel that parameter pack would end up moving construct-related compiler error messages deep into Resource, which I found caused headaches. Isn't nullopt_t already a tag? The terms reserve and construct are also more explicit to this application, and makes sure the first argument has a consistent meaning.

jonesmz: std::nullopt_t is natively understood by std::optional to not construct anything.

You're basically just reinventing functionality that std::optional already has built in.

Capital-Asterisk: This is unlikely, but if one of TYPE_T's constructors accepts a std::nullopt_t, then it would become ambiguous whether Resource(std::nullopt_t()) should construct or not. When there's two separate tags for construct and reserve, the first argument will never be ambiguous.

jonesmz: You would out the std::nullopt inside of an initializer list, so that the std::optional didn't pick it out.

Capital-Asterisk: In this rare case, Package::add causes unintended behavior of constructing the optional as empty.
```cpp
package.add<AcceptsNullopt>("name", std::nullopt);
// calls Resource<AcceptsNullopt>(std::nullopt_t)
```

Needing an initializer list specifically for this case is inconsistent.
```cpp
package.add<AcceptsNullopt>("name", {std::nullopt});
```

Anyways, it turns out stl already dealt with this issue with two separate tags. Is this good? 
```cpp
template <class TYPE_T>
struct Resource
{
    Resource(std::nullopt_t)
     : m_data(std::nullopt)
    { }
    template<typename ... ARGS_T>
    Resource(std::in_place_t, ARGS_T&& ... args)
     : m_data(std::in_place, std::forward<ARGS_T>(args)...)
    { }
    Resource(Resource&& move) = default;
    Resource(const Resource& copy) = delete;
    ~Resource()
    {
        assert(m_refCount == 0);
    }

    std::optional<TYPE_T> m_data;

    int m_refCount{0};
};
```

z-adams:
What's left to change here before it's ready to be merged?

Capital-Asterisk:
> What's left to change here before it's ready to be merged?

I'm waiting for @jonesmz to respond to the latest changes. He already approved a month ago anyways, so merge any time.

PR #91, opened 2021-03-06
Imgui integration

z-adams:
Add ImGui, ImPlot, and a simple SysGui to test them


src/test_application/OSPMagnum.cpp
```diff
@@ -45,21 +46,51 @@ OSPMagnum::OSPMagnum(const Magnum::Platform::Application::Arguments& arguments,
...
+
     m_timeline.start();
 
 }
 
 OSPMagnum::~OSPMagnum()
 {
+    // Free ImPlot
+    // ImPlot is not part of Magnum's imgui integration, and is handled manually
+    ImPlot::DestroyContext(m_implot);
```
z-adams, line 63: @jonesmz suggested a `unique_ptr` with a custom deleter

EhWhoAmI:
Have you tested if ImPlot works? I tested it, and it doesn't work on my computer.

z-adams:
Good point, I'll test it tonight. I tested ImGui and it worked, and the ImPlot stuff was just copied from my other branch where it also works. I'll let you know.

z-adams:
@EhWhoAmI I tested it and it works fine, by default the plot scales are weird so double click on the plot area and it should auto-fit the data. I added a graph to the fps counter window, if you can't see it then expand the window until it's visible.

EhWhoAmI:
Alright, great!

modules/FindMagnumIntegration.cmake
```diff
@@ -0,0 +1,309 @@
+#.rst:
```
jonesmz, line 1: No way to avoid copying these two files into our repo? We can't use the ones in the magnum repo directly somehow?

src/CMakeLists.txt
```diff
@@ -93,6 +98,9 @@ endif()
 # Include ENTT (header only lib)
 target_include_directories(osp-magnum PRIVATE ../3rdparty/entt/src)
 
+# Include ImPlot (simple library that depends on magnum-integration's imgui)
+target_include_directories(osp-magnum PRIVATE ../3rdparty/implot)
```
jonesmz, line 102: Doesn't this automatically get picked up when you add the magnum-imgui library to the list of libraries to link to?

src/test_application/OSPMagnum.cpp
jonesmz: Could you remind me what the context was?

The point of having a unique_ptr with a custom deleter is that when the unique_ptr goes out of scope the custom deleter gets triggered, which can either delete the thing, or cause whatever arbitrary outcomes to happen.

jonesmz (approved):
Great start!

EhWhoAmI (approved):

src/CMakeLists.txt
EhWhoAmI: I believe that implot is separate from ImGui, so you'll have to include it separately

jonesmz: That's true, but it's not what I was asking about.

Normally in CMake, when you have a library, even if it's header-only, you create a "library" in cmake that has all the include directories for the library tied to it.

Then, instead of having your own project use "target_include_directories", you simply do "target_link_libraries" and that automatically populates any shared libraries, static libraries, dependencies, and include directories for you.

So my confusion was "Why aren't we using the cmake "library" for imgui/implot, instead of directly adding their include directories?" If we're doing it this way because imgui/implot don't have cmake "libraries", then that's terrible and we should probably send them patches to fix their broken.

z-adams: I can't remember why I did it this way, but generally implot/imgui are meant to be included directly since they're small libraries with no dependencies. Feel free to explain in more detail how I should go about trying what you're suggesting.

Capital-Asterisk (commented):
I'm a bit skeptical of how well this will work out further on; it doesn't seem to fit in the ECS that well. I think the main issue is how the state of the gui stored in OSPMagnum, but are all processed from ActiveScene. Also consider how GUIs should be renderable onto textures and stuff.

Some other possible scheme I thought about:
* Entity 1:
  * ACompImguiContext: component that stores the imgui context
  * ACompGuiToScreen: tag component to tell a system to render this gui to the screen
* Entity 2:
  * ACompGui: refers to an entity that contains a ACompImguiContext (entity 1)
  * ACompGuiDebug: tag struct to render the "Debug" gui here
  * ACompGuiShipStatus: tag struct to render "Ship Status"

src/test_application/OSPMagnum.cpp
z-adams: ImGui's context is managed by OSPMagnum since it's a magnum plugin that integrates seamlessly. ImPlot is an external library, so its context needs to be manually freed. You suggested the `unique_ptr` so that we can use the default `~OSPMagnum`.

src/test_application/OSPMagnum.h
```diff
@@ -75,6 +77,10 @@ class OSPMagnum : public Magnum::Platform::Application
     constexpr MapActiveScene_t& get_scenes() { return m_scenes; }
 
 private:
+    Magnum::ImGuiIntegration::Context m_imgui{Magnum::NoCreate};
+    ImPlotContext* m_implot;
```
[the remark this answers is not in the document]
z-adams: Yes, OSPMagnum is where the key events are polled (see OSPMagnum.cpp), and need to be passed directly to the `m_imgui` member. The ImGui context is more akin to the OpenGL context (similarly stored internal to `Magnum::Platform::Application`) and doesn't belong to any particular scene.

src/osp/Active/SysGUI.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "osp/Active/activetypes.h"
+#include <Magnum/ImGuiIntegration/Context.hpp>
+
+namespace osp::active
+{
+
+struct ACompGUIWindow
+{
+    std::function<void(ActiveScene&)> m_function;
```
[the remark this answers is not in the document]
z-adams: So get rid of the GUI system, and just insert the lambdas directly? I guess that could work.

z-adams:
@Capital-Asterisk part of the weirdness comes from the way ImGui works in general. It's completely stateless and has a global context, so the weird `std::function` components do actually store all the GUI state (in mutable capture variables), they just also have to store the instructions that draw the pane. I guess it would be worth thinking about how a non-immediate GUI system should be structured though, and see if we can cram imgui into that framework.

Drawing to textures has nothing to do with ImGui as it will just draw to whatever framebuffer is actively bound, does our existing renderer have support for that yet in general? We might need to do that first before we can worry about it for the GUIs.

Capital-Asterisk:
@z-adams 

> It's completely stateless and has a global context, ...

Imgui is very not stateless; its current context and window is stored as global variables. There can be multiple contexts that can be switched between using SetCurrentContext. 

>  the weird `std::function` components do actually store all the GUI state (in mutable capture variables), they just also have to store the instructions that draw the pane. 

The state of the windows (window size, position, and stuff) is stored in the context. The windows are identified by their title. I don't see any lambda captures in the std::functions, and I really doubt they store anything.

> Drawing to textures has nothing to do with ImGui as it will just draw to whatever framebuffer is actively bound....

This barebones GUI doesn't need all the features yet, but should at least be structured in a way that supports them. Your current implementation doesn't specify where the GUIs are going to be drawn, and rendering onto virtual instrument screens is one of the intended features. This is why I suggested storing contexts in components, and having a `ACompGuiToScreen` tag to specify where the GUI is rendered to.

Capital-Asterisk:
The way Magnum Integration encapsulates the Imgui contexts might make the changes I suggested a bit more difficult. Magnum::ImGuiIntegration::Context can be thrown around as a component, but it also bundles a full-on application with it, with OpenGL dependent resources. It looks possible to handle and switch between multiple instances of Context. The on-screen gui component would have to store a reference to OSPMagnum.

Like the physics world, contexts can be temporary states that can be deleted and just get regenerated if they're not present. Window states can be saved some other way that Imgui probably has a solution for.

z-adams:
@Capital-Asterisk Is window position and size the only thing that the ImGui context stores? Values within the windows (collapse state, tickmark states, slider values, etc) are all stored by the user, which is what I have in my functions (the `std::array`s initialized in the FPS counter window).

If what you're saying is true, then I'm sort of tempted to either ditch Magnum's ImGui integration, or just start writing our own GUI. For the MVP we literally need like 5 numbers as far as GUI goes, so we aren't really in immediate need of much.

Ditching Magnum's ImGui integration might be the best option as it would allow us to keep using ImGui but retain full control of the contexts. I've already done it this way for ImPlot, so it wouldn't be much different. Thoughts?

z-adams:
Just pushed a commit (most recent one) with a sort of intermediate experiment in decoupling the GUI from OSPMagnum a bit. I left the event handling inside OSPMagnum, but moved the GUI contexts to components which live in the ActiveScene. The GUI system activates the relevant contexts at the beginning of its update, and then calls the lambdas in the components I had before. It's pretty rough still; in particular, I'm not quite sure where OSPMagnum should decide which scene's GUI contexts to act on (ultimately this would be determined by whichever scene the mouse is currently captured by). I made a function in ActiveScene to fetch the GUI context, but for now I just manually assign it to OSPMagnum's member pointer in flight.cpp.

I have some cleanup to do and I just realized I can move more of the GUI rendering code from OSPMagnum to SysGUI (the frame clearing and drawing stuff), but maybe the overall scheme will stimulate some ideas. I suspect @Capital-Asterisk won't like OSPMagnum extracting the GUI information from ActiveScene this way but it's pretty convenient and I haven't figured out a clean way to freeze the control inputs which should be handled by ImGui from within the ActiveScene. I'm open to suggestions though.

src/CMakeLists.txt
jonesmz: You should make an "interface" library (or use the one that imgui and implot define, in their own cmake file), and then use "target_link_library(MyImPlotInterfaceLib)" to add that library.

The interface library doesn't have any cpp files or shared librarys / static libraries. It just exists to provide an ergnomic way to refer to a header-only-library and all of it's dependencies.

src/test_application/OSPMagnum.h
```diff
@@ -74,12 +75,14 @@ class OSPMagnum : public Magnum::Platform::Application
...
     constexpr MapActiveScene_t& get_scenes() { return m_scenes; }
 
+    void set_active_GUI(osp::active::ActiveScene& rScene);
 private:
-
     void drawEvent() override;
 
     osp::UserInputHandler m_userInput;
 
+    Magnum::ImGuiIntegration::Context* m_activeImgui{nullptr};
```
Capital-Asterisk, line 84: This doesn't seem to point to stable memory

Capital-Asterisk (commented):
Considering rendering GUIs to textures and instrument screens, there should be components to determine which context the ACompGUIWindows are being rendered to. Either a vector stored along side the context that lists entities containing windows, or components in the window entities that point to a context. I'm not quite sure if it's that important to decided that right now though.

jonesmz:
@z-adams I know you're not really actively working on this, but if you have time, could you rebase this PR on the latest code so that the compile errors are taken care of?

z-adams:
@jonesmz This was last updated before the current rendering system was made so it'll take a little while to bring up to date. I may have time to do it this weekend, I just can't do it at the moment

jonesmz:
@z-adams Closing, because it's been open for several months without action.

Feel free to re-open if you decide to work on the imgui stuff again, but I know your FPS game is taking most of your time.

PR #92, opened 2021-03-16
Integrate taskflow (draft)

z-adams:
I've taken a first pass at integrating taskflow just to try it out, but it leads to some bizarre errors deep inside EnTT. Try it out and see if you can repro.

Most of the changes are modifying all systems to have fully static update functions, and adding a questionable `smc_update` member to the systems as a temporary place to store the description of the update process for that system.


jonesmz:
If you have time could you resolve the merge conflicts? ;)

z-adams:
@jonesmz it's way too early to worry about that, not only does it not work yet, but even once we get it working, by then Cap's new system code will be merged which conflicts almost exactly with this branch so the better option is to make this branch work and then splice the Taskflow integration code (which is very little, basically just ActiveScene.cpp) over into cap's branch.

jonesmz:
Regarding the smc_update stuff.

I think a much more reliable way to handle this is to declare a list of components that are going to be read-from the system, and written to by the system.

These dependencies can be used to automatically schedule the task in the right order. A system that reads component X must always come after a system that writes component X.

I can't think of a reason why we would want two systems to be allowed to write to the same components. And think that we would instead want to break that component into two. One that the first system writes to, and the other system reads from that component and writes to its component.

This let's us ultimately break down everything in ENTT to provide an acylic directed graph of tasks for each frame.

jonesmz:
Did you try setting the number of threads that taskflow is allowed to use to 1? I couldn't find any code that was trying to set the threads.

z-adams:
@jonesmz where did you see something about configuring taskflow threads? The basic usage instructions on their git readme don't mention that.

jonesmz:
```cpp
tf::Executor executor(4);  // creates an executor of 4 workers
```
https://taskflow.github.io/taskflow/LimitTheMaximumConcurrency.html

z-adams:
By default it uses the maximum number of hardware concurrent threads. I'm going to try sticking taskflow into a new branch with Cap's static systems just in case the issue with this branch is induced by something I did unrelated to taskflow

jonesmz:
Right. But I'm saying that if you're having what might be concurrent access issues, reducing the thread count to 1 may make it stop doing that.

Then once you know you're having concurrent access issues, it becomes easier to debug.

z-adams:
Ah understood, I'll give that a try.

z-adams:
Closing until we get Taskflow working on our own threads

PR #93, opened 2021-03-17
Remove IDynamicSystem and make all systems static

Capital-Asterisk:
Might as well clear the spaghetti now rather than later so we can stop building new changes on top of it.

Systems are suppose to be just static functions, but for some reason I decided to use classes at the start of the project. This led to the `IDynamicSystem` interface (just a virtual destructor) that allows the lifetime of these system classes to be managed by `ActiveScene`. They are also stored in a map so they can be accessed with strings.

This is a rather messy and horrible system, and we've been taking steps towards fully static systems (see #46). Since only static functions are used, there is no longer a purpose in initializing system classes. The only purpose left for system classes is to manage the lifetime of `FunctionOrderHandles`, which are pointlessly ref-counted handles to a function in a `FunctionOrder`.

This PR removes `IDynamicSystem` entirely, and makes all affected system classes static.
* Removed `IDynamicSystem`-related functions in `ActiveScene`, replaced with debug_update_add and debug_render_add to create `FunctionOrderHandle`s. These are denoted with "debug" as `FunctionOrder` would likely be replaced soon (see #2, #92)
* All the constructors were replaced with an add_functions static function instead
* Fix dangling pointer issue in `SysNewton`
* A few other minor cleanups

This issue progresses on #46, but doesn't fix it entirely. Machines follow a similar scheme for their systems, but that will be handled in a different PR. 

This also conflicts with #92 with many systems being changed to static.

### DIFF
--- a/src/osp/Active/ActiveScene.h
+++ b/src/osp/Active/ActiveScene.h
@@ -48,14 +48,11 @@ namespace osp::active
 /**
  * An ECS 3D Game Engine scene that implements a scene graph hierarchy.
  *
- * Components are prefixed with AComp, for "Active Component."
+ * The state of scene is represented with Active Entities (ActiveEnt) which are
+ * compositions of Active Components (AComp-prefixed structs). Behaviours are
+ * added by injecting System functions into the Update Order (a list of
+ * functions called in a specific order, use debug_update_add).
  *
- * Features are added through "Dynamic Systems" (IDynamicSystem), which are
- * classes that add functions to the UpdateOrder.
- *
- *
- * In ECS, "System" refers to the functions; but here, systems refer to the
- * classes containing those functions too.
  */
 class ActiveScene
 {
@@ -196,7 +193,6 @@ public:
     template<class SYSMACH_T, typename... ARGS_T>
     void system_machine_create(ARGS_T &&... args);
 
-
     /**
      * Find a registered SysMachine by name. This accesses a map.
      * @param name [in] Name used as a key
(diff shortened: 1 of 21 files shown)


src/osp/Active/SysNewton.cpp
```diff
@@ -626,10 +619,33 @@ void SysNewton::on_shape_destruct(ActiveReg_t& reg, ActiveEnt ent)
 
 void SysNewton::on_world_destruct(ActiveReg_t& reg, ActiveEnt ent)
 {
+
+
     NewtonWorld const *world = reg.get<ACompNwtWorld>(ent).m_nwtWorld;
     if (world != nullptr)
     {
-        NewtonDestroyAllBodies(world);
```
Capital-Asterisk, line 632: When NewtonDestroyAllBodies was previously called, it would invalidate all the pointers in the ACompCollider, leading to a crash when they're destructed.

src/osp/Active/SysWire.cpp
```diff
@@ -51,21 +48,6 @@ void WireInput::doErase()
...
-}
-
-
-void SysWire::update_propagate(ActiveScene& rScene)
-{
-    for (DependentOutput &output : m_dependentOutputs)
-    {
-        output.m_element->propagate_output(output.m_output);
-    }
-}
```
Capital-Asterisk, line 68: This doesn't actually do anything. All of the propagate_output functions are blank or just prints. Wiring system needs a ton of redesigning too.

src/osp/Active/ActiveScene.h
```diff
@@ -206,31 +202,13 @@ class ActiveScene
...
-     * @tparam SYSTEM_T Type of registered IDynamicSys
-     * @return Reference to specified IDynamicSys
-     */
-    template<class SYSTEM_T>
-    SYSTEM_T& dynamic_system_find();
-
-    bool dynamic_system_it_valid(MapDynamicSys_t::iterator it);
+    template<typename... ARGS_T>
+    void debug_render_add(ARGS_T &&... args)
+    { m_renderHandles.emplace_back(std::forward<ARGS_T>(args)...); }
```
jonesmz, line 211: I believe that emplace back returns a reference.

You might as well return what it returns. Something might want to use the reference in the future.

src/osp/Active/SysWire.h
```diff
@@ -287,11 +287,11 @@ class WireOutput : private LinkedList<WireInput>
 //-----------------------------------------------------------------------------
 
 
-class SysWire : public IDynamicSystem
+class SysWire
 {
 public:
 
-    static const std::string smc_name;
+    static inline std::string smc_name = "Wire";
```
z-adams, line 294: The other systems seem to have lost this member, should this one have stayed?

For the rest, is the `smc_name` concept being removed entirely?

src/osp/Active/SysNewton.cpp
jonesmz: Ugh. That's a terrible interface. Shame on newton.

z-adams (approved):
Merge when ready

jonesmz (approved):

jonesmz:
Big improvement! Awesome.

PR #95, opened 2021-03-23
Add spdlog

EhWhoAmI:
This adds `spdlog`, and removes most of the `std::cout`s


src/osp/OSPApplication.cpp
```diff
@@ -24,8 +24,14 @@
  */
 #include "OSPApplication.h"
 
+#include "spdlog/sinks/stdout_color_sinks.h"
+
 using namespace osp;
 
+osp::OSPApplication::OSPApplication()
+{
+    m_logger = spdlog::stdout_color_mt("console");}
```
z-adams, line 33: Messed up bracket

src/osp/OSPApplication.h
```diff
@@ -61,10 +63,14 @@ class OSPApplication
     size_t debug_num_packages() const { return m_packages.size(); }
 
     universe::Universe& get_universe() { return m_universe; }
+    
+    std::shared_ptr<spdlog::logger> get_logger() { return m_logger; };
```
z-adams, line 67: I wonder if it's worth returning a raw pointer or a reference or something, to prevent incrementing the `std::shared_ptr` ref counter, @jonesmz any thoughts?

z-adams:
I've never integrated something like this before, what are the tradeoffs between having a universal global instance in main or something vs storing an instance in a specific place? Just abstraction?

src/adera/Machines/Rocket.cpp
```diff
@@ -222,12 +223,12 @@ void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
...
     {
-        std::cout << "ERROR: could not find plume anchor for MachineRocket "
-            << ent << "\n";
+        m_scene.get_application().get_logger()->error("ERROR: could not find plume anchor for MachineRocket {}", ent);
         return;
     }
-    std::cout << "MachineRocket " << ent << "'s associated plume: "
-        << plumeNode << "\n";
+
+    m_scene.get_application().get_logger()->info("MachineRocket {0}\'s associated plume: {1}", ent, plumeNode);
```
jonesmz, line 230: why are you doing {0} and {1} here but in other places you don't do that, and just use {}.

src/adera/Machines/Rocket.cpp
```diff
@@ -236,8 +237,8 @@ void SysMachineRocket::attach_plume_effect(ActiveEnt ent)
     DependRes<PlumeEffectData> plumeEffect = pkg.get<PlumeEffectData>(effectName);
     if (plumeEffect.empty())
     {
-        std::cout << "ERROR: couldn't find plume effect " << effectName << "!\n";
-        return;
+        m_scene.get_application().get_logger()->error("ERROR: couldn't find plume effect  {}", effectName);
+            return;
```
jonesmz, line 241: dont change indentation without reason.

src/adera/Machines/UserControl.cpp
```diff
@@ -36,7 +36,7 @@ const std::string SysMachineUserControl::smc_name = "UserControl";
 
 void MachineUserControl::propagate_output(WireOutput* output)
 {
-    std::cout << "propagate test: " << output->get_name() << "\n";
+    spdlog::info("Propagate test: {}", output->get_name());
```
jonesmz, line 39: Why are you tracing with a global logger instance here, but you use a specific handle in other places?

Capital-Asterisk (commented):
All the loggers should be initialized and managed in one place. spdlog's global registry may be a good solution, or perhaps just a global or application-wide sink for all the loggers to connect to. I'm not really sure what the best solution is here; I'll think about it a bit more.

src/osp/OSPApplication.h
jonesmz: @EhWhoAmI @z-adams you should return a const& to the m_logger variable.
Be wide in what you accept, and narrow in what you return.

@Capital-Asterisk that's how spdlog works, basically.

jonesmz:
> I've never integrated something like this before, what are the tradeoffs between having a universal global instance in main or something vs storing an instance in a specific place? Just abstraction?

In my opinion, we need to have multiple component-level log object handles.

For example, from the spdlog documentation:

```cpp
Create a multiple loggers all sharing the same file (sink) aka categories
#include <iostream>
#include "spdlog/spdlog.h"
#include "spdlog/sinks/daily_file_sink.h"
int main(int, char* [])
{
    try
    {
        auto daily_sink = std::make_shared<spdlog::sinks::daily_file_sink_mt>("logfile", 23, 59);
        // create synchronous  loggers
        auto net_logger = std::make_shared<spdlog::logger>("net", daily_sink);
        auto hw_logger  = std::make_shared<spdlog::logger>("hw",  daily_sink);
        auto db_logger  = std::make_shared<spdlog::logger>("db",  daily_sink);      

        net_logger->set_level(spdlog::level::critical); // independent levels
        hw_logger->set_level(spdlog::level::debug);
         
        // globally register the loggers so the can be accessed using spdlog::get(logger_name)
        spdlog::register_logger(net_logger);
    }
    catch (const spdlog::spdlog_ex& ex)
    {
        std::cout << "Log initialization failed: " << ex.what() << std::endl;
    }
}
```

Each "system" should have it's own log handle instance, so that we can use the configuration file to individually set what level of detail each "system" will log at.

It would also be nice to be able to specify in the OSP config file which systems log to the same log file, and which log to their own individual files, but that's not necessary to support at this moment.

I'm a bit frustrated that spdlog doesn't support the ability to do hierarchical components, or components as a parameter to the log call.

I'm also frustrated that they have individual functions for trace / note / info / critical / warning. In my opinion, these should be parameters to the log function, not their own function.

But, it is what it is, I suppose. They've traded loss of flexibility in exchange for easier (and potentially faster) code.

src/osp/CommonPhysics.cpp
```diff
@@ -52,9 +54,10 @@ float shape_volume(ECollisionShape shape, Vector3 scale)
     case ECollisionShape::TERRAIN:
     case ECollisionShape::COMBINED:
     default:
-        std::cout << "Error: unsupported shape for volume calc\n";
+        spdlog::error("Unsupported shape for volume calc");
         assert(false);
-    }
+        return 0;
```
jonesmz, line 59: Don't add a return here. That should be done in it's own changelist.

jonesmz:
You should use these macros, instead of the individual calls.

How to remove all debug statements at compile time ?
Define SPDLOG_ACTIVE_LEVEL to the desired log level before including spdlog.h and use the SPDLOG_ macros:

#define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_INFO // All DEBUG/TRACE statements will be removed by the pre-processor
#include "spdlog/spdlog.h"
...
SPDLOG_DEBUG("debug message to default logger"); // removed at compile time
SPDLOG_LOGGER_TRACE(my_logger, "trace message"); // removed at compile time
SPDLOG_INFO("info message to default logger");   // included


That will allow us to disable more verbose tracing at compil;e time.

src/osp/Resource/AssetImporter.h
```diff
@@ -157,6 +159,17 @@ PluginManager;
                                Magnum::UnsignedInt parentProtoIndex,
                                Magnum::UnsignedInt childGltfIndex);
 
+    static std::shared_ptr<spdlog::logger> get_logger() 
+    {
+        if (logger == nullptr) {
+            logger = spdlog::stdout_color_mt("assetimporter");
```
jonesmz, line 165: don't do lazy initialization.

Require that the logger be passed in when the assertimporter is constructed, and assert that it is not null.

z-adams: AssetImporter is used basically fully statically, which is probably why he did it that way

src/osp/UserInputHandler.cpp
```diff
@@ -131,9 +133,11 @@ ButtonVar::ButtonVar(ButtonMap::iterator button, ButtonVarConfig cfg) :
 
 
 UserInputHandler::UserInputHandler(int deviceCount) :
-        m_deviceToButtonRaw(deviceCount)
+    m_deviceToButtonRaw(deviceCount)
 {
     m_controls.reserve(7);
+
+    logger = spdlog::stdout_color_mt("userinput");
```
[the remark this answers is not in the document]
jonesmz:
```cpp
Create a multiple loggers all sharing the same file (sink) aka categories
#include <iostream>
#include "spdlog/spdlog.h"
#include "spdlog/sinks/daily_file_sink.h"
int main(int, char* [])
{
    try
    {
        auto daily_sink = std::make_shared<spdlog::sinks::daily_file_sink_mt>("logfile", 23, 59);
        // create synchronous  loggers
        auto net_logger = std::make_shared<spdlog::logger>("net", daily_sink);
        auto hw_logger  = std::make_shared<spdlog::logger>("hw",  daily_sink);
        auto db_logger  = std::make_shared<spdlog::logger>("db",  daily_sink);      

        net_logger->set_level(spdlog::level::critical); // independent levels
        hw_logger->set_level(spdlog::level::debug);
         
        // globally register the loggers so the can be accessed using spdlog::get(logger_name)
        spdlog::register_logger(net_logger);
    }
    catch (const spdlog::spdlog_ex& ex)
    {
        std::cout << "Log initialization failed: " << ex.what() << std::endl;
    }
}
```

src/osp/UserInputHandler.cpp
```diff
@@ -368,8 +371,8 @@ void UserInputHandler::event_raw(DeviceId deviceId, int buttonEnum,
         return; // button not registered
     }
 
-    std::cout << "sensitive button pressed\n";
-
+    logger->trace("sensitive button pressed");
+
```
jonesmz, line 375: unnecessary whitespace change

src/osp/UserInputHandler.cpp
```diff
@@ -419,15 +422,15 @@ void UserInputHandler::update_controls()
...
+                logger->trace("RELEASE");
             }
         }
         else if (control.m_triggered)
         {
             // start holding down the control. control.m_exprRelease should
             // have been generated earlier
             control.m_held = true;
-            std::cout << "HOLD\n";
+            logger->trace("HOLD");
```
jonesmz, line 433: the logger has shared-pointer semanitcs.

it should be named "pLogger".

src/planet-a/Active/SysPlanetA.cpp
```diff
@@ -258,17 +258,17 @@ void SysPlanetA::update_geometry(ActiveScene& rScene)
...
             for (chindex_t i = 0; i < rPlanetGeo.get_chunk_count(); i ++)
             {
                 debug_create_chunk_collider(ent, planet, i);
-                std::cout << "* completed chunk collider: " << i << "\n";
+                rScene.get_application().get_logger()->info("* completed chunk collider: {}", i);
             }
 
-            std::cout << "Planet colliders done\n";
+            rScene.get_application().get_logger()->info("Completed planet colliders");
+
```
jonesmz, line 271: unnecessary whitepace

src/planet-a/IcoSphereTree.cpp
```diff
@@ -223,7 +225,6 @@ trindex_t IcoSphereTree::ancestor_at_depth(trindex_t start, uint8_t targetDepth)
 
 int IcoSphereTree::subdivide_add(trindex_t triInd)
 {
-    //std::cout << "m_icoTree->subdivide_add(" << t << ");\n";
```
jonesmz, line 226: log these at debug.

z-adams:
@jonesmz 
>That will allow us to disable more verbose tracing at compile time.

You're telling me spdlog doesn't do this internally? Isn't that the whole point of logging libraries like this...?

jonesmz:
Don't forget to call spdlog::shutdown at the end of main().

```
Asynchronous logging hangs on application exit under Windows
At the end of main, call spdlog::shutdown();
```

jonesmz:
and make sure you update your PR and fix the merge conflicts.

jonesmz:
> @jonesmz
> 
> > That will allow us to disable more verbose tracing at compile time.
> 
> You're telling me spdlog doesn't do this internally? Isn't that the whole point of logging libraries like this...?
> 
> Edit: can't we do that with `if constexpr` instead of macros?

If-constexpr could have the same outcome, yes. But why bother when the macros already exist?

The logging library *DOES* do this internally, but it does it as a runtime check, not compile time.

The macro let's us completely disable verbose logging, if we wanted to, at compile time without effecting runtime logging of other log levels. 

This let's us skip the function call overhead, essentially.

Why would we do this? Its a cheap code style to enforce, and it can be useful for eliminating the logging as a source of performance problems. We wouldn't typically want to build with the verbose tracing being disabled, but we probably will from time to time.

src/osp/Resource/AssetImporter.h
jonesmz: Then we should probably store the logger in entt for the asset importer, or initialize a global variable for the asset importer to pick up.

src/adera/Machines/UserControl.cpp
EhWhoAmI: Because `MachineUSerControl` is not tied to any application, so I believe there is no logger to use, and I don't think it's a good idea to create a whole new logger just for this.

EhWhoAmI:
> All the loggers should be initialized and managed in one place. spdlog's global registry may be a good solution, or perhaps just a global or application-wide sink for all the loggers to connect to. I'm not really sure what the best solution is here; I'll think about it a bit more.
I can probably create a global static logger, or just use the global static logger. I believe there's even a macro that enables you to see the line number in the log, so maybe we can use that instead

src/osp/Resource/AssetImporter.h
```diff
@@ -157,6 +159,17 @@ PluginManager;
...
+        }
+        return logger;
+    }
+    /**
+    * Unique logger for AssetImporter because it's important
+    */
+    static std::shared_ptr<spdlog::logger> logger;
```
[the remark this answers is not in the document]
EhWhoAmI: Assetimporter does not have a reference to the application, so it makes it a bit awkward to link them over.

jonesmz:
> > All the loggers should be initialized and managed in one place. spdlog's global registry may be a good solution, or perhaps just a global or application-wide sink for all the loggers to connect to. I'm not really sure what the best solution is here; I'll think about it a bit more.
> 
> I can probably create a global static logger, or just use the global static logger. I believe there's even a macro that enables you to see the line number in the log, so maybe we can use that instead

You're referring to the __LINE__ macro?

That's superceeded by the `std::source_location` type, which provides even better information (at the expensive of polluting the language with *TERRIBLE* layering violations, but oh well).

EhWhoAmI:
How about this, I add a function to main.cpp to initialize the loggers, and then I use `spdlog::get` to get all the loggers

EhWhoAmI:
These two commits change the loggers so that they are in one place, in OSPApplication.cpp, and gives the loggers more information.

EhWhoAmI:
Whoops I might have not tested the build before committing, I'll fix that later

Capital-Asterisk (approved):
It feels a bit inconsistent to have some parts using the longest log function of all c++ projects in existence `rScene.get_application().get_logger()`, and other parts being something else. All I can think of is shortening it to `rScene.get_logger()`.

A logger for each system class would be cool, but structurally, the additions here are all good.

src/osp/UserInputHandler.cpp
```diff
@@ -368,8 +370,8 @@ void UserInputHandler::event_raw(DeviceId deviceId, int buttonEnum,
         return; // button not registered
     }
 
-    std::cout << "sensitive button pressed\n";
-
+    SPDLOG_LOGGER_TRACE(p_logger, "sensitive button pressed");
```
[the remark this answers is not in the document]
EhWhoAmI: They are traces, so they are ignored by default. I think we can leave them in there.

z-adams (approved):

jonesmz:
I think this is sufficiently well done to merge.

jonesmz (approved):

Issue #83, opened 2021-02-22
Zoom does not zoom well, change to a log scale. 


EhWhoAmI:
The scroll wheel zooms in and out annoyingly slow, so to improve the zoom, we can switch the scroll to a log scale. 

Something like this:
`double new_zoom = scroll_amount^constant * current_zoom` would make it scroll much faster.

PR #96, opened 2021-04-02
zooming set to +/- 0.3 times current zoom

flindt:
Suggested alternative to issue #83

### DIFF
--- a/src/test_application/DebugObject.cpp
+++ b/src/test_application/DebugObject.cpp
@@ -252,8 +252,8 @@ void DebugCameraController::update_physics_post()
     Vector3 posRelative = xform.translation() - xformTgt.translation();
 
     // set camera orbit distance
-    constexpr float distSensitivity = 1.0f;
-    m_orbitDistance += distSensitivity * static_cast<float>(-m_scrollInput.dy());
+    constexpr float distSensitivity = 0.3f;
+    m_orbitDistance += m_orbitDistance * distSensitivity * static_cast<float>(-m_scrollInput.dy());
     
     // Clamp orbit distance to avoid producing a degenerate m_orbitPos vector
     constexpr float minDist = 5.0f;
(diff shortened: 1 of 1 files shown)


z-adams:
I think the point of that issue was actually that the scroll is too slow at large zoom values, so this way would only fix it in the other direction. Maybe @EhWhoAmI can clarify what he meant.

flindt:
Yes, I'd agree it was way too slow.

The point of my suggestion is that zoom change is proportional with current zoom level - increasing the steps at large zoom values.  
This way one click on the scroll wheel feels the same whether you are close or far away from the ship.

EhWhoAmI:
Yeah, I meant it to be proportion to what the zoom level is. My suggestion was based off my guess of the zoom in code before I took a look, so it may be inaccurate

EhWhoAmI:
I tested it out, and if you zoom out far enough, it will keep zooming in and out by itself, making it a bit epileptic. 

But otherwise, the zoom is much more bearable.

Capital-Asterisk:
This is a nice change.

> I tested it out, and if you zoom out far enough, it will keep zooming in and out by itself, making it a bit epileptic.
> 
> But otherwise, the zoom is much more bearable.

When you zoom out too far, the vehicle unloads. Since the vehicle is still selected by the camera controller, it will move towards it until it is back in the active area.

On line 181, change this
```cpp
// round to nearest (floatingOriginThreshold)
Vector3s tra = Vector3s(xform.translation() / floatingOriginThreshold)
             * floatingOriginThreshold;
```

To this, so that the floating origin is centered on the vehicle instead of the camera.
```cpp
// round to nearest (floatingOriginThreshold)
Vector3s tra = Vector3s((xform.translation() - m_orbitPos) / floatingOriginThreshold)
             * floatingOriginThreshold;
```

flindt:
How did this break the windows build?

Capital-Asterisk:
> How did this break the windows build?

This happens a few times, it's github changing things in the build environment. The build is actually successful, it's the artifact release that failed because it seems the files are now built in a different directory.

EhWhoAmI:
Yeah, don't worry

jonesmz (approved):

EhWhoAmI:
Is this good to merge? If it is I'll just merge it.

EhWhoAmI (approved):

PR #102, opened 2021-04-17
Rendering system

z-adams:
My progress so far on the new rendering system. It doesn't yet support offscreen rendering or complicated multipass, but it currently has feature parity with the existing rendering scheme, but is structured in a much better way.

Objects are given an `ACompVisible` component to indicate they should be processed by the renderer.

Shader instances no longer exist; the rendering system (or an external system, if a nonstandard shader is being used) is responsible for creating an `ACompShader` which stores a function pointer to a static shader draw function, which is now responsible for collecting any information necessary to set uniforms and draw the object.


bin/OSPData/adera/Shaders/FullscreenTri.frag
```diff
@@ -0,0 +1,37 @@
...
+layout(location = 0, index = 0) out vec3 color;
+
+layout(location = 0) uniform sampler2D framebuffer;
+
+in vec2 uv;
+
+void main()
+{
+    color = texture(framebuffer, uv).rgb;
+}
```
jonesmz, line 37: no newline at end of file?

makes a lot of editors complain.

src/adera/Shaders/PlumeShader.cpp
```diff
@@ -24,31 +24,53 @@
...
+#include <osp/Active/SysRender.h>
+#include <osp/Resource/AssetImporter.h>
+
 #include <Magnum/GL/Version.h>
 #include <Magnum/GL/Shader.h>
 #include <Corrade/Containers/Reference.h>
 #include <Magnum/GL/Texture.h>
 
-using namespace Magnum;
+using Magnum::Int;
```
jonesmz, line 36: ugly ;)

src/adera/Shaders/PlumeShader.cpp
```diff
@@ -24,31 +24,53 @@
...
+
+    using Magnum::GL::Texture2D;
+    using Magnum::GL::Mesh;
+    // Collect uniform data
+    auto& transform = rScene.reg_get<ACompTransform>(e);
+    auto& rPlumeComp = rScene.reg_get<ACompExhaustPlume>(e);
+    auto const& plumedata = *rPlumeComp.m_effect;
+    Mesh& rMesh = *rScene.reg_get<ACompMesh>(e).m_mesh;
+
+    constexpr std::string_view texName = "OSPData/adera/noise1024.png";
```
jonesmz, line 56: no need to change it for this PR, but eventually we should make sure either:
1) Textures like this are directly imported into the binary, there are various techniques for how to do this, so i won't detail the mechanism that's used, but the point is that it's not loaded from disk.
2) Resources loaded from disk are described in a config file, so that they can be changed by the end user.

src/osp/Active/ActiveScene.cpp
```diff
@@ -216,25 +216,33 @@ void ActiveScene::update_hierarchy_transforms()
...
-    //Matrix4 cameraInverse;
-
-    // TODO: check if camera has the right components
-    ACompCamera& cameraComp = reg_get<ACompCamera>(camera);
-    ACompTransform& cameraTransform = reg_get<ACompTransform>(camera);
-
-    //cameraProject = cameraComp.m_projection;
-    cameraComp.m_inverse = cameraTransform.m_transformWorld.inverted();
+    auto renderView = m_registry.view<ACompRenderingAgent, ACompPerspective3DView, ACompRenderer>();
+    for (auto [e, agent, view, renderer] : renderView.each())
```
jonesmz, line 226: imho, this would be better as `auto const&`, but whether that actually results in different code gen depends on the specific types of the variables.

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+        std::array<float, 12> surfData
+        {
+            // Vert position    // UV coordinate
+            -1.0f,  1.0f,       0.0f,  1.0f,
+            -1.0f, -3.0f,       0.0f, -1.0f,
+             3.0f,  1.0f,       2.0f,  1.0f
+        };
+
+        GL::Buffer surface(std::move(surfData), GL::BufferUsage::StaticDraw);
```
jonesmz, line 91: std::array can't be moved from. It's equivelent to a raw c-style array on the stack.

As a result, you should probably make surfData a static constexpr variable, to give the compiler a better chance at not fucking up.

You could go a step further and make it an std::initializer_list<float> as well.

src/adera/Shaders/PlumeShader.cpp
z-adams: This will all be config'd eventually, although importing into the binary is an interesting idea for loading shaders.

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+        resources.add<GL::Framebuffer>("offscreen_fbo", std::move(fbo));
+
+    rScene.reg_emplace<ACompRenderTarget>(rScene.hier_get_root(),
+        viewSize, std::move(fboRes));
+    rScene.reg_emplace<ACompFBOColorAttachment>(rScene.hier_get_root(), colorRes);
+
+
+    // Create the default render pipeline
+    resources.add<RenderPipeline>("default",
+        std::move(create_render_pipeline()));
```
jonesmz, line 132: if create_render_pipeline returns an object by value, or by const&, then this std::move() will have no effect.

Functions that return an object by value will automatically have the returned value converted to && in this context.

std::move() silently ignores const& parameters, returning a const&.

So the only time this std::move() will have an effect on the program is if create_render_pipeline() returns a value by non-const reference, and while that's not inherently problematic, given the name of the function, i'd argue that's not what this function should return.

So, long story short, the std::move() here can be removed.

z-adams: Good to know, thanks

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+            Renderer::setFaceCullingMode(Renderer::PolygonFacing::Front);
+            SysRender::draw_group(rScene, transparentView, camera);
+
+            // Draw frontfaces
+            Renderer::setFaceCullingMode(Renderer::PolygonFacing::Back);
+            SysRender::draw_group(rScene, transparentView, camera);
+        }
+    );
+
+    return {pipeline};
```
jonesmz, line 191: I recommend returning a braced initializer list instead of giving the vector a name.


Uhhhhhh, also, why aren't you using the vector of handles?

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    auto view = rScene.get_registry().view<ACompVisible>();
+    for (ActiveEnt e : view)
+    {
+        auto* shader = rScene.reg_try_get<ACompShader>(e);
+
+        if (shader == nullptr)
+        {
+            // Drawable entity has not been initialized yet
+            constexpr ShaderDrawFnc_t defaultDraw = shader::Phong::draw_entity;
```
jonesmz, line 204: imho not really useful to have this defaultDraw variable.

src/osp/Active/SysRender.cpp
z-adams: I had it there originally, but it causes a crash when the resource is deleted at program exit, and I have no actual use for the handles (the only way to create a FunctionOrder is via constructing FunctionOrderHandles) so I discarded them.

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+            shader = &rScene.reg_emplace<ACompShader>(e, defaultDraw);
+            rScene.reg_emplace<ACompOpaque>(e);
+        }
+    }
+
+    auto agentView = rScene.get_registry().view<ACompRenderingAgent>(entt::exclude<ACompRenderer>);
+    for (ActiveEnt e : agentView)
+    {
+        constexpr std::string_view defaultRenderer = "default";
+        rScene.reg_emplace<ACompRenderer>(e, std::string{defaultRenderer});
```
jonesmz, line 214: I created the SharedString class specifically for write-once, read-many strings like this.

src/osp/Active/SysRender.cpp
z-adams: I separated it out like that because I'm not sure if I'm going to just hard-code it, or if it should be fetched from elsewhere.

src/osp/Active/SysRender.cpp
z-adams: I'll look into it, but I think ultimately these strings should probably be replaced by compile-time hashed keys that are easier to work with than strings.

src/osp/Active/SysVehicle.cpp
```diff
@@ -366,20 +365,11 @@ std::pair<ActiveEnt, std::vector<SysVehicle::MachineDef>> SysVehicle::part_insta
...
-            rScene.reg_emplace<Phong::ACompPhongInstance>(currentEnt, std::move(shader));
-
-            CompDrawableDebug& bBocks = rScene.reg_emplace<CompDrawableDebug>(
-                        currentEnt, meshRes, &Phong::draw_entity, 0x0202EE_rgbf);
-
-            //new DrawablePhongColored(*obj, *m_shader, *mesh, 0xff0000_rgbf, m_drawables);
+            // by now, the mesh and texture should both exist so we emplace them
+            // as components to be consumed by the Phong shader
+            rScene.reg_emplace<ACompMesh>(currentEnt, meshRes);
+            rScene.reg_emplace<ACompDiffuseTex>(currentEnt, std::move(textureResources[0]));
```
jonesmz, line 371: you move only the first texture? Why bother creating a vector if you are only ever using the first item in it?

src/osp/Shaders/FullscreenTriShader.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    set_framebuffer(texture);
+    draw(surface);
+}
+
+FullscreenTriShader& FullscreenTriShader::set_framebuffer(GL::Texture2D& rTex)
+{
+    rTex.bind(static_cast<Int>(TextureSlot::FramebufferSlot));
+    return *this;
+}
```
jonesmz, line 59: missing end of file newline

src/osp/Shaders/FullscreenTriShader.h
```diff
@@ -0,0 +1,75 @@
...
+{
+public:
+    // Vertex attribs
+    typedef Magnum::GL::Attribute<0, Magnum::Vector2> Position;
+    typedef Magnum::GL::Attribute<1, Magnum::Vector2> TextureCoordinates;
+
+    // Outputs
+    enum : Magnum::UnsignedInt
+    {
+        ColorOutput = 0
```
jonesmz, line 43: Interesting trick,
 but enums that aren't enum classes can be implicitly converted to other types, can't they? so this doesn't buy you much.

src/osp/Active/SysVehicle.cpp
z-adams: Temporary ugliness. Eventually we'll be loading several textures, and the code above this is sort of sitting around waiting to be hammered into a proper PBR resource loader. It was like that before this PR.

src/osp/Shaders/FullscreenTriShader.h
```diff
@@ -0,0 +1,75 @@
...
+    // Uniforms
+    enum class UniformPos : Magnum::Int
+    {
+        FramebufferSampler = 0
+    };
+
+    // Texture2D slots
+    enum class TextureSlot : Magnum::Int
+    {
+        FramebufferSlot = 0
```
jonesmz, line 65: Since you always have to refer to this as TextureSlot::FramebufferSlot, maybe drop the trailing Slot suffix?

src/osp/Shaders/FullscreenTriShader.h
z-adams: I'm just following Magnum's convention, this is how they suggest designing shader program wrappers, so I've stuck with it for the sake of consistency even though it's a bit verbose.

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+            // Draw backfaces
+            Renderer::setFaceCullingMode(Renderer::PolygonFacing::Front);
+            SysRender::draw_group(rScene, transparentView, camera);
+
+            // Draw frontfaces
+            Renderer::setFaceCullingMode(Renderer::PolygonFacing::Back);
+            SysRender::draw_group(rScene, transparentView, camera);
+        }
+    );
```
Capital-Asterisk, line 189: So these two render passes is now the "debug renderer"

src/osp/Active/SysRender.h
```diff
@@ -0,0 +1,164 @@
...
+template<typename VIEW_T>
+void SysRender::draw_group(ActiveScene& rScene, VIEW_T& drawlist, ACompCamera const& camera)
+{
+    for (ActiveEnt e : drawlist)
+    {
+        auto shader = drawlist.template get<ACompShader>(e);
+
+        shader.m_drawCall(e, rScene, camera);
+    }
+}
```
Capital-Asterisk, line 162: For stuff like filling a g-buffer or common objects, there should be a "draw group using this shader only" function too.

Capital-Asterisk (commented):
A step in the right direction. Still expecting lots of stuff discussed in the Discord server.

src/adera/SysExhaustPlume.h
```diff
@@ -35,6 +35,8 @@ struct ACompExhaustPlume
 {
     ActiveEnt m_parentMachineRocket{entt::null};
     DependRes<PlumeEffectData> m_effect;
+    float m_time{0.0f};
+    float m_powerLevel{0.0f};
```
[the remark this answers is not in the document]
z-adams: PlumeEffectData was supposed to represent a plume config file, whose fields are just read straight into a struct. This will probably all shift around as we figure out what sorts of properties are needed to define a plume, or if multiple shaders are needed, etc. I put the time in there since before it was stored in the shader instance, although I suppose once ActiveScene provides a way to get the global time we could get rid of it.

src/osp/Active/SysRender.cpp
z-adams: Yep, or default. Ultimately it probably won't end up looking much different from this.

src/osp/Active/SysRender.h
z-adams: That would usually be handled as one of the render passes. Can you provide an example?

src/osp/Active/SysRender.cpp
jonesmz: Ah. That's terrible :-)

src/osp/Active/SysVehicle.cpp
jonesmz: understood

src/osp/Shaders/FullscreenTriShader.h
jonesmz: Understood.

Kind of feel like you should promote this to a proper enum class, but you do you.

z-adams: Oh I thought it was an enum class, that's just a typo. The Magnum way uses enum classes and I have everywhere except apparently here. I'll fix it.

src/osp/Active/ActiveScene.cpp
z-adams: This doesn't really seem to change the types that get used, which are `T&&` for the entity type (`e`) and `T&` for all the component types. This doesn't change even if I add `const`. I'm going to leave it as-is since this is the way skypjack himself seems to write it.

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,274 @@
...
+    /* Configure pipeline resources */
+
+    // Add an offscreen framebuffer
+
+    Vector2i viewSize = GL::defaultFramebuffer.viewport().size();
+
+    GL::Texture2D color;
+    color.setStorage(1, GL::TextureFormat::RGB8, viewSize);
+    DependRes<GL::Texture2D> colorRes =
+        resources.add<GL::Texture2D>("offscreen_fbo_color", std::move(color));
```
Capital-Asterisk, line 111: Package::add is actually an emplace, you can do this:
```cpp
DependRes<GL::Texture2D> colorRes =
        resources.add<GL::Texture2D>("offscreen_fbo_color");
colorRes->setStorage(1, GL::TextureFormat::RGB8, viewSize);
```

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    auto agentView = rScene.get_registry().view<ACompRenderingAgent>(entt::exclude<ACompRenderer>);
+    for (ActiveEnt e : agentView)
+    {
+        constexpr std::string_view defaultRenderer = "default";
+        rScene.reg_emplace<ACompRenderer>(e, std::string{defaultRenderer});
+    }
+}
+
+void SysRender::render(ActiveScene& rScene, ACompCamera const& camera)
```
Capital-Asterisk, line 217: Since render order is no longer being called, this function is no longer being called and is leftover?

src/osp/Active/SysRender.cpp
```diff
@@ -0,0 +1,274 @@
...
+        auto* shader = rScene.reg_try_get<ACompShader>(e);
+
+        if (shader == nullptr)
+        {
+            // Drawable entity has not been initialized yet
+            constexpr ShaderDrawFnc_t defaultDraw = shader::Phong::draw_entity;
+            shader = &rScene.reg_emplace<ACompShader>(e, defaultDraw);
+            rScene.reg_emplace<ACompOpaque>(e);
+        }
+    }
```
Capital-Asterisk, line 207: This is a lot sketchier than I imagined. Since ACompShader just stores a function pointer now, it's safe to set the draw function on initialization.

src/osp/Active/SysRender.cpp
z-adams: Yep, just forgot to remove it

src/osp/Active/SysRender.cpp
z-adams: Elaborate?

Capital-Asterisk: `shader::Phong::draw_entity` can be stored somewhere in the scene or package as the "default shader". SysVehicle can use this this to make an ACompShader.

Capital-Asterisk (approved):

src/test_application/flight.cpp
```diff
@@ -121,15 +134,18 @@ void testapp::test_flight(std::unique_ptr<OSPMagnum>& pMagnumApp,
...
 
     // Create the camera entity
     ActiveEnt camera = rScene.hier_create_child(rScene.hier_get_root(),
                                                        "Camera");
     auto &cameraTransform = rScene.reg_emplace<ACompTransform>(camera);
-    auto &cameraComp = rScene.get_registry().emplace<ACompCamera>(camera);
+    auto &cameraComp = rScene.reg_emplace<ACompCamera>(camera);
+    rScene.reg_emplace<ACompRenderingAgent>(camera, SysRender::get_default_rendertarget(rScene));
+    rScene.reg_emplace<ACompPerspective3DView>(camera, camera);
+    rScene.reg_emplace<ACompRenderer>(camera);
```
Capital-Asterisk, line 148: It would really help to organize the setup in one readable section of code, such as:
```cpp
    // Create the camera entity
    ActiveEnt camera = rScene.hier_create_child(rScene.hier_get_root(),
                                                "Camera");
    auto &cameraTransform = rScene.reg_emplace<ACompTransform>(camera);
    auto &cameraComp = rScene.reg_emplace<ACompCamera>(camera);
    rScene.reg_emplace<ACompFloatingOrigin>(camera);
    cameraTransform.m_transform = Matrix4::translation(Vector3(0, 0, 25));
    cameraComp.m_far = 1u << 24;
    cameraComp.m_near = 1.0f;
    cameraComp.m_fov = 45.0_degf;
    cameraComp.calculate_projection();
    cameraComp.m_viewport
            = Vector2(Magnum::GL::defaultFramebuffer.viewport().size());

    // Add the debug camera controller to the scene. This adds controls
    auto camObj = std::make_unique<DebugCameraController>(rScene, camera);

    // Add a ACompDebugObject to camera to manage camObj's lifetime
    rScene.reg_emplace<ACompDebugObject>(camera, std::move(camObj));

    // Setup default render system
    osp::active::SysRender::setup(rScene);
    load_shaders(rScene);
    
    // Setup scene renderer, added to the camera component
    rScene.reg_emplace<ACompRenderingAgent>(camera, SysRender::get_default_rendertarget(rScene));
    rScene.reg_emplace<ACompPerspective3DView>(camera, camera);
    rScene.reg_emplace<ACompRenderer>(camera);
```

z-adams: Sounds good, I will make that change and then merge

PR #103, opened 2021-04-17
Automatically filter filetypes for Visual Studio

z-adams:
Adds CMake code to separate source files by type so that they can be put into Visual Studio filters in the project explorer. Makes it way easier to open header files and shader source files. Let me know if there's a better way to do this, but it works as-is.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -42,11 +42,22 @@ find_package(MagnumPlugins REQUIRED
 find_package(Threads)
 
 # Gather paths to OSP headers and sources
-file (GLOB_RECURSE CPP_FILES *.cpp)
-file (GLOB_RECURSE H_FILES *.h)
-set (SOURCE_FILES ${CPP_FILES} ${H_FILES})
+add_executable(osp-magnum)
+file (GLOB_RECURSE CPP_FILES CONFIGURE_DEPENDS *.cpp)
+file (GLOB_RECURSE H_FILES CONFIGURE_DEPENDS *.h)
+target_sources(osp-magnum PRIVATE "${CPP_FILES}" "${H_FILES}")
 
-add_executable(osp-magnum ${CPP_FILES})
+# For some reason, source_group() requires all files to be added to the executable
+set(SHADERS_DIR "../bin/OSPData/adera/Shaders")
+file (GLOB_RECURSE SHADER_FILES CONFIGURE_DEPENDS
+    "${SHADERS_DIR}/*.vert"
+    "${SHADERS_DIR}/*.frag"
+    "${SHADERS_DIR}/*.comp"
+)
+target_sources(osp-magnum PRIVATE "${SHADER_FILES}")
+
+# Set osp-magnum as startup project
+set_property(DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT osp-magnum)
 
 target_include_directories(osp-magnum PRIVATE .)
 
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -84,10 +97,17 @@ elseif(MINGW)
...
-  target_compile_options(osp-magnum PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
+	# Set the MSVC debug working directory
+  	set_property(TARGET osp-magnum PROPERTY VS_DEBUGGER_WORKING_DIRECTORY ../../bin)
+
+  	# Enforce conformance mode for osp-magnum
+  	target_compile_options(osp-magnum PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
+
+  	# Segregate headers, shaders into filters
+  	source_group("Shader Files" FILES ${SHADER_FILES})
+  	source_group("Header Files" FILES ${H_FILES})
```
EhWhoAmI, line 110: Can you also add `set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT osp-magnum)`
to this? This makes it so you don't have to change the startup project, which makes it a bit more convenient.

z-adams: Yep, on it

Capital-Asterisk (commented):
This doesn't affect me but this kind of looks a little bit hacky to add GLSL shaders as source files. I guess this is just the way it is with MSVC? Nothing seems to be broken so I'm okay with it.

Capital-Asterisk (approved):

z-adams:
Yep, apparently it's sort of a weird fluke of VS that you have to add any filtered files to the source files. Otherwise it doesn't work.

src/CMakeLists.txt
```diff
@@ -46,7 +46,23 @@ file (GLOB_RECURSE CPP_FILES *.cpp)
...
+	# For some reason, source_group() requires all files to be added to the
+	# executable. I did this conditionally in case it messes up other platforms
+	set(SHADERS_DIR "../bin/OSPData/adera/Shaders")
+  	file (GLOB_RECURSE SHADER_FILES
+  		"${SHADERS_DIR}/*.vert"
+  		"${SHADERS_DIR}/*.frag"
+  		"${SHADERS_DIR}/*.comp"
+  		)
+  	list(APPEND SOURCE_FILES ${SHADER_FILES})
+	add_executable(osp-magnum ${SOURCE_FILES})
```
jonesmz, line 59: Better might be
```cmake
add_executable(osp-magnum)
file (GLOB_RECURSE H_FILES CONFIGURE_DEPENDS *.h)
file (GLOB_RECURSE CPP_FILES CONFIGURE_DEPENDS *.cpp)
target_sources(osp-magnum "${CPP_FILES}" "${H_FILES}")

# For some reason, source_group() requires all files to be added to the executable
	set(SHADERS_DIR "../bin/OSPData/adera/Shaders")
  	file (GLOB_RECURSE SHADER_FILES CONFIGURE_DEPENDS
  		"${SHADERS_DIR}/*.vert"
  		"${SHADERS_DIR}/*.frag"
  		"${SHADERS_DIR}/*.comp"
  		)
target_sources(osp-magnum "${SHADER_FILES}")
```

jonesmz: And I wouldn't make it conditional on the platform. Unless someone adds a toolchain file to cmake that teaches cmake how to compile .vert, .frag and .comp files, cmake will ignore them.

src/CMakeLists.txt
```diff
@@ -84,10 +100,17 @@ elseif(MINGW)
 
 endif()
 
-# Set the MSVC debug working directory, enforce conformance mode for osp-magnum
+# MSVC quality of life improvements for Visual Studio
 if(MSVC)
-  set_property(TARGET osp-magnum PROPERTY VS_DEBUGGER_WORKING_DIRECTORY ../../bin)
-  target_compile_options(osp-magnum PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
+	# Set the MSVC debug working directory
+  	set_property(TARGET osp-magnum PROPERTY VS_DEBUGGER_WORKING_DIRECTORY ../../bin)
```
jonesmz, line 106: imho it's unnecessary to make these items conditional on if(MSVC) None of them should cause problems outside of the MSVC block.

jonesmz (approved):
Approved assuming some action is taken regarding my comments
(Ignoring my comments is an acceptable, but not highly recommended, option)

src/CMakeLists.txt
z-adams: I guess that's true, but if they have no effect outside of MSVC, why not make them conditional? Just for cleanliness' sake. I don't really know CMake so I was just playing it safe.

jonesmz: MSVC can be used to compile for linux

and MSVC probably isn't the only IDE that would look at that flag. It might be the only one now, but others might look in the future.

EhWhoAmI (approved):

PR #99, opened 2021-04-07
Rework Vehicles

Capital-Asterisk:
Merge #90 first

This PR restructures how vehicles are stored and activated into the physics world,. Using a queue-like scheme for better performance and possibly simpler code, this makes way for a new wiring system and easier multi threading in the future.

Previously, activating vehicles into the physics world was done by instantiating each individual part. To instantiate a part's machines, each part accesses a map of ISysMachine interfaces to call a virtual function to create the required components. This isn't good for the ECS and the multi threading schemes described elsewhere (#2 #92), as various ECS components are created in an unpredictable order. Instantiating parts individually also has issues with the vehicle being in an incomplete state, requiring #61.

In these new changes, Activating a vehicle will create part colliders and meshes but no machines. The vehicle will have an `ACompVehicleInConstruction`, telling other systems that it is an incomplete vehicle. System functions then add the required machine components by reading the vehicle's blueprint.

Aside from just adding machines, adding meshes and physics setup can be separated into their own systems too. FunctionOrder is pretty much used like one big dependency injection for everything needed for vehicle initialization.

Some temporary workarounds were also used to work with the current wiring system. 

Other notable changes:
* Rewrite prototypes and blueprints into Structure-of-Arrays style
* Added "Machine IDs" which use entt::family to index different machine types
* ISysMachine is removed, essentially fulfilling #46. 
* Separated BlueprintVehicle building functions into a new VehicleBuilder class
* Machines are now in their own entities parented to the part
* Moved MachineContainer into its own file

Update order for vehicle-related functions looks like this:
```
* vehicle_activate (Create the vehicles)
* mach_container_construct (Add required MachineContainers)
* mach_rcs_construct (Add required MachineRCSControllers)
* mach_rocket_construct (Add required MachineRockets)
* mach_usercontrol_construct (Add required MachineUserControls)
* vehicle_modification (Handle separations, part removals, etc)
...
```

Vehicles now look like this:
```
  ->[4]: Vehicle
  ->  ->[115]: part_phLinRCS
  ->  ->  ->[120]: Machine
  ->  ->  ->[119]: Machine
  ->  ->  ->[118]: linrcsblk
  ->  ->  ->[117]: fx_plume_phRCSplume
  ->  ->  ->[116]: col_RCSblk
  ->  ->[109]: part_phLinRCS
  ->  ->  ->[114]: Machine
  ->  ->  ->[113]: Machine
  ->  ->  ->[112]: linrcsblk
[...]

PR #108, opened 2021-04-27
Node Wiring System(s)

Capital-Asterisk:
Depends on #99

The wiring system is the last bit of old code to de-spaghettify. See issue #56; however, the implementation in this PR does not resemble anything described in the issue.

## Structure

The new way wires are organized looks like this:

image

* Values are passed around by reading and writing into Nodes
  * One Machine writes into the node, another Machine reads the node
  * Usually there will be separate read and write fields to prevent race conditions and for correct propagation speeds
* Machine entities also have Panel components that contain Ports
* Panels can contain multiple Ports
* Each Port can connect to only one Node
* Nodes can connect to multiple Ports; connections are called Links which are stored in the Node
* Links can have custom states

There are no longer any dedicated input and output classes like before.

This system is entirely templated. Propagation mechanics and topology is up to the user. Templates functions for common logical signals are provided, known as Signals. The system can be further templated for things like serial connections, events, and resource flow.

This example diagram shows a UserControl and a Container connected to a Rocket:

image

## Updating

Mentioned earlier, Nodes need to store a read value and a write value. Consider two machines A and B: A writes to Node, and B reads the Node.
* If A updates first, then B would immediately get the new value written by A.
* If B updates first, then it would read the previous value written by A.
* If they update in parallel, then there would be a race condition.
If A writes into a specialized write field in the Node instead, then this problem would be eliminated. This also means that the Node needs its own update function where it can replace its read field with the new written value.

Wire components must also update multiple times per frame. 1/60 second propagation delays are simply unacceptable for controlling fast-moving spacecraft in real-time.

Updating wire components in the scene alternates between two different steps:

* Node update - Nodes consider their new written values and update their read values
* Machine Calculate update - Machines read connected input nodes, do calculations, and write to connected output nodes 

## Performance

So far, the current wire update in the 'simple' scene takes 0.003ms to update on a 3.0Ghz processor. 

In a modified scene with 100 vehicles each with 16 RCS controllers, 17 rockets, and a user controller, it takes only 0.3ms to update everything while RCS is firing and throttle is changing. 

## Notes:

The idea of needing an intermediate value may seem unnecessary at first, but it greatly simplifies many calculations. It also stops Machine components from having to write into each other's memory. This prevents the terrible ECS component polymorphism needed before.

The term "Node" comes from Electrical Engineering: https://en.wikipedia.org/wiki/Node_(circuits)

This PR adds the right infrastructure, but leaves a few opened gaps:
* Fuel flow is disabled entirely
* No way to delete nodes yet

### DIFF
--- a/src/adera/Machines/Container.cpp
+++ b/src/adera/Machines/Container.cpp
@@ -130,13 +130,13 @@ void SysMachineContainer::update_construct(ActiveScene &rScene)
         for (BlueprintMachine &mach : rVehConstr.m_blueprint->m_machines[id])
         {
             // Get part
-            ActiveEnt partEnt = rVeh.m_parts[mach.m_blueprintIndex];
+            ActiveEnt partEnt = rVeh.m_parts[mach.m_partIndex];
 
             // Get machine entity previously reserved by SysVehicle
             auto& machines = rScene.reg_get<ACompMachines>(partEnt);
             ActiveEnt machEnt = machines.m_machines[mach.m_protoMachineIndex];
 
-            BlueprintPart const& partBp = vehBp.m_blueprints[mach.m_blueprintIndex];
+            BlueprintPart const& partBp = vehBp.m_blueprints[mach.m_partIndex];
             instantiate(rScene, machEnt,
                         vehBp.m_prototypes[partBp.m_protoIndex]
                                 ->m_protoMachines[mach.m_protoMachineIndex],
--- a/src/adera/Machines/Container.cpp
+++ b/src/adera/Machines/Container.cpp
@@ -141,7 +141,6 @@ void SysMachineContainer::update_construct(ActiveScene &rScene)
                         vehBp.m_prototypes[partBp.m_protoIndex]
                                 ->m_protoMachines[mach.m_protoMachineIndex],
                         mach);
-            rScene.reg_emplace<ACompMachineType>(machEnt, id);
         }
     }
 }
--- a/src/adera/Machines/Container.cpp
+++ b/src/adera/Machines/Container.cpp
@@ -35,31 +35,6 @@ using namespace adera::active::machines;
 
 /* MachineContainer */
 
-void MachineContainer::propagate_output(WireOutput* output)
-{
-
-}
-
-WireInput* MachineContainer::request_input(WireInPort port)
-{
-    return nullptr;
-}
-
-WireOutput* MachineContainer::request_output(WireOutPort port)
-{
-    return &m_outputs;
-}
-
-std::vector<WireInput*> MachineContainer::existing_inputs()
-{
-    return {};
-}
-
-std::vector<WireOutput*> MachineContainer::existing_outputs()
-{
-    return {&m_outputs};
-}
-
 uint64_t MachineContainer::request_contents(uint64_t quantity)
 {
     if (quantity > m_contents.m_quantity)
@@ -166,11 +141,7 @@ void SysMachineContainer::update_construct(ActiveScene &rScene)
                         vehBp.m_prototypes[partBp.m_protoIndex]
                                 ->m_protoMachines[mach.m_protoMachineIndex],
                         mach);
-            rScene.reg_emplace<ACompMachineType>(machEnt, id,
-                    [] (ActiveScene &rScene, ActiveEnt ent) -> Machine&
-                    {
-                        return rScene.reg_get<MachineContainer>(ent);
-                    });
+            rScene.reg_emplace<ACompMachineType>(machEnt, id);
         }
     }
 }
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -1,6 +1,6 @@
 /**
  * Open Space Program
- * Copyright © 2019-2020 Open Space Program Project
+ * Copyright Â© 2019-2020 Open Space Program Project
  *
  * MIT License
  *
@@ -40,6 +40,8 @@ using namespace osp;
 using Magnum::Vector3;
 using Magnum::Matrix4;
 
+using osp::active::wiretype::AttitudeControl;
+using osp::active::wiretype::Percent;
 
 void SysMachineRCSController::add_functions(ActiveScene& rScene)
 {
@@ -121,50 +123,94 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
     }
 }
 
-#if 0
-void SysMachineRCSController::update_controls(ActiveScene& rScene)
+
+void SysMachineRCSController::update_calculate(ActiveScene& rScene)
 {
+    ACompWireNodes<Percent> &rNodesPercent = SysWire::nodes<Percent>(rScene);
+    std::vector<ActiveEnt>& rToUpdate
+            = SysWire::to_update<MachineRCSController>(rScene);
+
+    std::vector< nodeindex_t<wiretype::Percent> > propagatePercent;
+
     auto view = rScene.get_registry().view<MachineRCSController>();
 
-
-    for (ActiveEnt ent : view)
+    for (ActiveEnt ent : rToUpdate)
     {
         auto& machine = view.get<MachineRCSController>(ent);
-        if (!machine.m_enable)
+
+        float influence = 0.0f;
+
+        auto const *panelAtCtrl = rScene.get_registry()
+                .try_get< ACompWirePanel<AttitudeControl> >(ent);
+
+        if (panelAtCtrl != nullptr)
         {
-            continue;
+            WirePort<AttitudeControl> const *portCommand
+                    = panelAtCtrl->port(MachineRCSController::smc_wiCommandOrient);
+
+            // Read AttitudeControl Command Input
+            if (portCommand != nullptr)
+            {
+                auto const &nodesAttCtrl
+                        = rScene.reg_get< ACompWireNodes<AttitudeControl> >(
+                            rScene.hier_get_root());
+                WireNode<AttitudeControl> const &nodeCommand
+                        = nodesAttCtrl.get_node(portCommand->m_nodeIndex);
+
+                // Get rigidbody ancestor and its transformation component
+                auto const *rbAncestor
+                        = SysPhysics_t::try_get_or_find_rigidbody_ancestor(
+                            rScene, ent);
+                auto const &compRb = rScene.reg_get<ACompRigidBody_t>(
+                            rbAncestor->m_ancestor);
+                auto const &compTf = rScene.reg_get<ACompTransform>(
+                            rbAncestor->m_ancestor);
+
+                Matrix4 transform = rbAncestor->m_relTransform;
+
+                // TODO: RCS translation is not currently implemented, only
+                // rotation
+                Vector3 commandTransl = Vector3{0.0f};
+                Vector3 commandRot = nodeCommand.m_state.m_value.m_attitude;
+                Vector3 thrusterPos = transform.translation()
+                                    - compRb.m_centerOfMassOffset;
+                Vector3 thrusterDir = transform.rotation()
+                                    * Vector3{0.0f, 0.0f, 1.0f};
+
+                if (commandRot.length() > 0.0f || commandTransl.length() > 0.0f)
+                {
+                    influence = thruster_influence(thrusterPos, thrusterDir,
+                                                   commandTransl, commandRot);
+                }
+            }
         }
 
-        using wiretype::AttitudeControl;
-        using Magnum::Vector3;
+        auto const *panelPercent = rScene.get_registry()
+                .try_get< ACompWirePanel<Percent> >(ent);
 
-        // Get rigidbody ancestor and its transformation component
-        auto const* pRbAncestor =
-            SysPhysics_t::try_get_or_find_rigidbody_ancestor(rScene, ent);
-        auto const& rCompRb = rScene.reg_get<ACompRigidBody_t>(pRbAncestor->m_ancestor);
-        auto const& rCompTf = rScene.reg_get<ACompTransform>(pRbAncestor->m_ancestor);
-
-        if (WireData *gimbal = machine.m_wiCommandOrient.connected_value())
+        if (panelPercent != nullptr)
         {
-            AttitudeControl *attCtrl = std::get_if<AttitudeControl>(gimbal);
+            WirePort<Percent> const *portThrottle
+                    = panelPercent->port(MachineRCSController::m_woThrottle);
 
-            Matrix4 transform = pRbAncestor->m_relTransform;
-
-            // TODO: RCS translation is not currently implemented, only rotation
-            Vector3 commandTransl = Vector3{0.0f};
-            Vector3 commandRot = attCtrl->m_attitude;
-            Vector3 thrusterPos = transform.translation() - rCompRb.m_centerOfMassOffset;
-            Vector3 thrusterDir = transform.rotation() * Vector3{0.0f, 0.0f, 1.0f};
-
-            float influence = 0.0f;
-            if (commandRot.length() > 0.0f || commandTransl.length() > 0.0f)
+            // Write to throttle output
+            if (portThrottle != nullptr)
             {
-                influence =
-                    thruster_influence(thrusterPos, thrusterDir, commandTransl, commandRot);
+                WireNode<Percent> &nodeThrottle
+                        = rNodesPercent.get_node(portThrottle->m_nodeIndex);
+
+                // Write possibly new throttle value to node
+                SysSignal<Percent>::signal_assign(
+                        rScene, {influence}, nodeThrottle,
+                        portThrottle->m_nodeIndex, propagatePercent);
             }
-            std::get<wiretype::Percent>(machine.m_woThrottle.value()).m_value = influence;
         }
     }
 
+    // Request propagation if wire nodes were modified
+    if (!propagatePercent.empty())
+    {
+        rNodesPercent.propagate_request(std::move(propagatePercent));
+        rScene.reg_get<ACompWire>(rScene.hier_get_root()).request_update();
+    }
 }
-#endif
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -40,30 +40,6 @@ using namespace osp;
 using Magnum::Vector3;
 using Magnum::Matrix4;
 
-void MachineRCSController::propagate_output(WireOutput* output)
-{
-
-}
-
-WireInput* MachineRCSController::request_input(osp::WireInPort port)
-{
-    return existing_inputs()[port];
-}
-
-WireOutput* MachineRCSController::request_output(osp::WireOutPort port)
-{
-    return existing_outputs()[port];
-}
-
-std::vector<WireInput*> MachineRCSController::existing_inputs()
-{
-    return {&m_wiCommandOrient};
-}
-
-std::vector<WireOutput*> MachineRCSController::existing_outputs()
-{
-    return {&m_woThrottle};
-}
 
 void SysMachineRCSController::add_functions(ActiveScene& rScene)
 {
@@ -141,11 +117,7 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
             ActiveEnt machEnt = machines.m_machines[mach.m_protoMachineIndex];
 
             rScene.reg_emplace<MachineRCSController>(machEnt);
-            rScene.reg_emplace<ACompMachineType>(machEnt, id,
-                    [] (ActiveScene &rScene, ActiveEnt ent) -> Machine&
-                    {
-                        return rScene.reg_get<MachineRCSController>(ent);
-                    });
+            rScene.reg_emplace<ACompMachineType>(machEnt, id);
         }
     }
 }
@@ -154,6 +126,7 @@ void SysMachineRCSController::update_controls(ActiveScene& rScene)
 {
     auto view = rScene.get_registry().view<MachineRCSController>();
 
+#if 0
     for (ActiveEnt ent : view)
     {
         auto& machine = view.get<MachineRCSController>(ent);
@@ -192,4 +165,5 @@ void SysMachineRCSController::update_controls(ActiveScene& rScene)
             std::get<wiretype::Percent>(machine.m_woThrottle.value()).m_value = influence;
         }
     }
+#endif
 }
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -43,8 +43,8 @@ using Magnum::Matrix4;
 
 void SysMachineRCSController::add_functions(ActiveScene& rScene)
 {
-    rScene.debug_update_add(rScene.get_update_order(), "mach_rcs", "wire", "controls",
-                            &SysMachineRCSController::update_controls);
+    //rScene.debug_update_add(rScene.get_update_order(), "mach_rcs", "wire", "controls",
+    //                        &SysMachineRCSController::update_controls);
     rScene.debug_update_add(rScene.get_update_order(), "mach_rcs_construct", "vehicle_activate", "vehicle_modification",
                             &SysMachineRCSController::update_construct);
 }
@@ -110,7 +110,7 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
         for (BlueprintMachine &mach : rVehConstr.m_blueprint->m_machines[id])
         {
             // Get part
-            ActiveEnt partEnt = rVeh.m_parts[mach.m_blueprintIndex];
+            ActiveEnt partEnt = rVeh.m_parts[mach.m_partIndex];
 
             // Get machine entity previously reserved by SysVehicle
             auto& machines = rScene.reg_get<ACompMachines>(partEnt);
@@ -122,11 +122,12 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
     }
 }
 
+#if 0
 void SysMachineRCSController::update_controls(ActiveScene& rScene)
 {
     auto view = rScene.get_registry().view<MachineRCSController>();
 
-#if 0
+
     for (ActiveEnt ent : view)
     {
         auto& machine = view.get<MachineRCSController>(ent);
@@ -165,5 +166,6 @@ void SysMachineRCSController::update_controls(ActiveScene& rScene)
             std::get<wiretype::Percent>(machine.m_woThrottle.value()).m_value = influence;
         }
     }
-#endif
+
 }
+#endif
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -117,7 +117,6 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
             ActiveEnt machEnt = machines.m_machines[mach.m_protoMachineIndex];
 
             rScene.reg_emplace<MachineRCSController>(machEnt);
-            rScene.reg_emplace<ACompMachineType>(machEnt, id);
         }
     }
 }
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -34,14 +34,38 @@
 #include <Magnum/Math/Vector.h>
 #include <functional>
 
-using namespace adera::active::machines;
-using namespace osp::active;
-using namespace osp;
-using Magnum::Vector3;
-using Magnum::Matrix4;
+using adera::active::machines::SysMachineRCSController;
+using adera::active::machines::MachineRCSController;
+using adera::wire::Percent;
+using adera::wire::AttitudeControl;
 
-using osp::active::wiretype::AttitudeControl;
-using osp::active::wiretype::Percent;
+using osp::active::ActiveScene;
+using osp::active::ActiveEnt;
+using osp::active::ACompMachines;
+using osp::active::ACompRigidBody_t;
+using osp::active::ACompTransform;
+using osp::active::SysPhysics_t;
+
+using osp::active::ACompWirePanel;
+using osp::active::ACompWireNodes;
+using osp::active::ACompWire;
+using osp::active::SysWire;
+using osp::active::SysSignal;
+using osp::active::WireNode;
+using osp::active::WirePort;
+using osp::active::UpdNodes_t;
+using osp::nodeindex_t;
+
+using osp::BlueprintMachine;
+
+using osp::Package;
+using osp::DependRes;
+using osp::Path;
+
+using osp::machine_id_t;
+using osp::NodeMap_t;
+using osp::Matrix4;
+using osp::Vector3;
 
 void SysMachineRCSController::add_functions(ActiveScene& rScene)
 {
@@ -96,7 +120,7 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
             .view<osp::active::ACompVehicle,
                   osp::active::ACompVehicleInConstruction>();
 
-    machine_id_t const id = mach_id<MachineRCSController>();
+    machine_id_t const id = osp::mach_id<MachineRCSController>();
 
     for (auto [vehEnt, rVeh, rVehConstr] : view.each())
     {
@@ -128,7 +152,7 @@ void SysMachineRCSController::update_calculate(ActiveScene& rScene)
     std::vector<ActiveEnt>& rToUpdate
             = SysWire::to_update<MachineRCSController>(rScene);
 
-    std::vector< nodeindex_t<wiretype::Percent> > updPercent;
+    UpdNodes_t<Percent> updPercent;
 
     auto view = rScene.get_registry().view<MachineRCSController>();
 
@@ -137,44 +161,45 @@ void SysMachineRCSController::update_calculate(ActiveScene& rScene)
 
         float influence = 0.0f;
 
-        auto const *panelAtCtrl = rScene.get_registry()
+        auto const *pPanelAtCtrl = rScene.get_registry()
                 .try_get< ACompWirePanel<AttitudeControl> >(ent);
 
-        if (panelAtCtrl != nullptr)
+        if (pPanelAtCtrl != nullptr)
         {
-            WirePort<AttitudeControl> const *portCommand
-                    = panelAtCtrl->port(MachineRCSController::smc_wiCommandOrient);
+            WirePort<AttitudeControl> const *pPortCommand
+                    = pPanelAtCtrl->port(MachineRCSController::smc_wiCommandOrient);
 
             // Read AttitudeControl Command Input
-            if (portCommand != nullptr)
+            if (pPortCommand != nullptr)
             {
                 auto const &nodesAttCtrl
                         = rScene.reg_get< ACompWireNodes<AttitudeControl> >(
                             rScene.hier_get_root());
                 WireNode<AttitudeControl> const &nodeCommand
-                        = nodesAttCtrl.get_node(portCommand->m_nodeIndex);
+                        = nodesAttCtrl.get_node(pPortCommand->m_nodeIndex);
 
                 // Get rigidbody ancestor and its transformation component
-                auto const *rbAncestor
+                auto const *pRbAncestor
                         = SysPhysics_t::try_get_or_find_rigidbody_ancestor(
                             rScene, ent);
                 auto const &compRb = rScene.reg_get<ACompRigidBody_t>(
-                            rbAncestor->m_ancestor);
+                            pRbAncestor->m_ancestor);
                 auto const &compTf = rScene.reg_get<ACompTransform>(
-                            rbAncestor->m_ancestor);
+                            pRbAncestor->m_ancestor);
 
-                Matrix4 transform = rbAncestor->m_relTransform;
+                Matrix4 transform = pRbAncestor->m_relTransform;
 
                 // TODO: RCS translation is not currently implemented, only
                 // rotation
                 Vector3 commandTransl = Vector3{0.0f};
-                Vector3 commandRot = nodeCommand.m_state.m_value.m_attitude;
+                Vector3 commandRot = nodeCommand.m_state.m_attitude;
                 Vector3 thrusterPos = transform.translation()
                                     - compRb.m_centerOfMassOffset;
                 Vector3 thrusterDir = transform.rotation()
                                     * Vector3{0.0f, 0.0f, 1.0f};
 
-                if (commandRot.length() > 0.0f || commandTransl.length() > 0.0f)
+                if ((commandRot.length() > 0.0f)
+                        || (commandTransl.length() > 0.0f))
                 {
                     influence = thruster_influence(thrusterPos, thrusterDir,
                                                    commandTransl, commandRot);
@@ -182,24 +207,24 @@ void SysMachineRCSController::update_calculate(ActiveScene& rScene)
             }
         }
 
-        auto const *panelPercent = rScene.get_registry()
+        auto const *pPanelPercent = rScene.get_registry()
                 .try_get< ACompWirePanel<Percent> >(ent);
 
-        if (panelPercent != nullptr)
+        if (pPanelPercent != nullptr)
         {
-            WirePort<Percent> const *portThrottle
-                    = panelPercent->port(MachineRCSController::m_woThrottle);
+            WirePort<Percent> const *pPortThrottle
+                    = pPanelPercent->port(MachineRCSController::m_woThrottle);
 
             // Write to throttle output
-            if (portThrottle != nullptr)
+            if (pPortThrottle != nullptr)
             {
-                WireNode<Percent> &nodeThrottle
-                        = rNodesPercent.get_node(portThrottle->m_nodeIndex);
+                WireNode<Percent> const &nodeThrottle
+                        = rNodesPercent.get_node(pPortThrottle->m_nodeIndex);
 
                 // Write possibly new throttle value to node
                 SysSignal<Percent>::signal_assign(
                         rScene, {influence}, nodeThrottle,
-                        portThrottle->m_nodeIndex, updPercent);
+                        pPortThrottle->m_nodeIndex, updPercent);
             }
         }
     }
@@ -207,7 +232,8 @@ void SysMachineRCSController::update_calculate(ActiveScene& rScene)
     // Request to update any wire nodes if they were modified
     if (!updPercent.empty())
     {
-        rNodesPercent.update_request(std::move(updPercent));
+        std::sort(std::begin(updPercent), std::end(updPercent));
+        rNodesPercent.update_write(updPercent);
         rScene.reg_get<ACompWire>(rScene.hier_get_root()).request_update();
     }
 }
--- a/src/adera/Machines/RCSController.cpp
+++ b/src/adera/Machines/RCSController.cpp
@@ -45,8 +45,6 @@ using osp::active::wiretype::Percent;
 
 void SysMachineRCSController::add_functions(ActiveScene& rScene)
 {
-    //rScene.debug_update_add(rScene.get_update_order(), "mach_rcs", "wire", "controls",
-    //                        &SysMachineRCSController::update_controls);
     rScene.debug_update_add(rScene.get_update_order(), "mach_rcs_construct", "vehicle_activate", "vehicle_modification",
                             &SysMachineRCSController::update_construct);
 }
@@ -130,13 +128,12 @@ void SysMachineRCSController::update_calculate(ActiveScene& rScene)
     std::vector<ActiveEnt>& rToUpdate
             = SysWire::to_update<MachineRCSController>(rScene);
 
-    std::vector< nodeindex_t<wiretype::Percent> > propagatePercent;
+    std::vector< nodeindex_t<wiretype::Percent> > updPercent;
 
     auto view = rScene.get_registry().view<MachineRCSController>();
 
     for (ActiveEnt ent : rToUpdate)
     {
-        auto& machine = view.get<MachineRCSController>(ent);
 
         float influence = 0.0f;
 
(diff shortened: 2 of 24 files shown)


src/adera/Machines/Container.cpp
```diff
@@ -99,32 +79,68 @@ void SysMachineContainer::update_containers(ActiveScene& rScene)
...
-Machine& SysMachineContainer::instantiate(ActiveEnt ent,
-    PrototypeMachine config, BlueprintMachine settings)
+MachineContainer& SysMachineContainer::instantiate(
+        osp::active::ActiveScene& rScene,
+        osp::active::ActiveEnt ent,
+        osp::PCompMachine const& config,
+        osp::BlueprintMachine const& settings)
 {
-    float capacity = std::get<double>(config.m_config["capacity"]);
+    float capacity = std::get<double>(config.m_config.at("capacity"));
```
jonesmz, line 88: Did you want it to throw an exception if it can't find it?

src/osp/Active/SysWire.cpp
```diff
@@ -27,32 +27,53 @@
...
-    wireFrom.insert(&wireTo);
-    // TODO: check for dependent outputs and add to list and sort
+    auto view = rScene.get_registry().view<ACompWireNeedUpdate>();
+
+    ActiveReg_t::poly_storage t = rScene.get_registry().storage(entt::type_id<ACompWireNeedUpdate>());
+
+
+    auto const& wire = rScene.reg_get<ACompWire>(rScene.hier_get_root());
+
+    int updateLimit = 16;
```
jonesmz, line 57: Recommend against signed. And recommend explicitly specifying bit width.

src/adera/Machines/Container.cpp
```diff
@@ -155,22 +130,17 @@ void SysMachineContainer::update_construct(ActiveScene &rScene)
...
+            BlueprintPart const& partBp = vehBp.m_blueprints[mach.m_partIndex];
             instantiate(rScene, machEnt,
                         vehBp.m_prototypes[partBp.m_protoIndex]
                                 ->m_protoMachines[mach.m_protoMachineIndex],
                         mach);
-            rScene.reg_emplace<ACompMachineType>(machEnt, id,
-                    [] (ActiveScene &rScene, ActiveEnt ent) -> Machine&
-                    {
-                        return rScene.reg_get<MachineContainer>(ent);
-                    });
```
Capital-Asterisk, line 173: These lambda functions were previously used only by the old wiring system for polymorphically accessing the machine component. This is no longer needed, and ACompMachineType is now created in SysVehicle

src/adera/Machines/RCSController.cpp
```diff
@@ -1,6 +1,6 @@
 /**
  * Open Space Program
- * Copyright � 2019-2020 Open Space Program Project
+ * Copyright © 2019-2020 Open Space Program Project
```
Capital-Asterisk, line 3: My IDE had been bothering me about the encoding of this unicode copyright symbol. Also noting bumping the copyright to 2021.

src/adera/Machines/RCSController.h
```diff
@@ -80,62 +83,29 @@ class SysMachineRCSController
...
-    osp::active::WireOutput* request_output(osp::WireOutPort port) override;
-
-    std::vector<osp::active::WireInput*> existing_inputs() override;
-    std::vector<osp::active::WireOutput*> existing_outputs() override;
-
-private:
-    osp::active::WireInput m_wiCommandOrient{this, "Orient"};
-    osp::active::WireOutput m_woThrottle{this, "Throttle"};
-
-    osp::active::ActiveEnt m_rigidBody{entt::null};
```
Capital-Asterisk, line 115: I find it's quite beautiful how the RCS controller had been made stateless, with all common properties separated into their own components. Like a logic gate, it's only purpose is to read a value from its input, and write a value to its output.

src/adera/Machines/RCSController.cpp
```diff
@@ -134,62 +110,104 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
...
+                        portThrottle->m_nodeIndex, updPercent);
             }
-            std::get<wiretype::Percent>(machine.m_woThrottle.value()).m_value = influence;
         }
     }
+
+    // Request to update any wire nodes if they were modified
+    if (!updPercent.empty())
+    {
+        rNodesPercent.update_request(std::move(updPercent));
```
Capital-Asterisk, line 210: Nodes can be marked for update from multiple threads (concurrency not yet implemented though). To reduce the number of times the mutexes are locked, changes are accumulated into a vector (updPercent in this case), and are merged into the main vector after updating.

src/osp/Active/SysWire.cpp
Capital-Asterisk: This doesn't seem that important. It's a simple count that won't go any higher than the value assigned to it now, and it's also in a temporary place where it should be made configurable from elsewhere eventually.

src/adera/Machines/UserControl.cpp
```diff
@@ -111,6 +84,11 @@ void SysMachineUserControl::update_sensor(ActiveScene &rScene)
     // InputDevice.IsActivated()
     // Combination
     auto const &usrCtrl = rScene.reg_root_get_or_emplace<ACompUserControl>(rScene.get_user_input());
+    ACompWireNodes<Percent> &rNodesPercent = SysWire::nodes<Percent>(rScene);
+    ACompWireNodes<AttitudeControl> &rNodesAttCtrl = SysWire::nodes<AttitudeControl>(rScene);
+
+    std::vector< nodeindex_t<wiretype::Percent> > updPercent;
+    std::vector< nodeindex_t<wiretype::AttitudeControl> > updAttCtrl;
```
z-adams, line 91: Is there any way we could wrap this stuff up into a cleaner interface? `SysWire::get_node<T>(index)` for the node groups? I guess passing the update list into `signal_assign()` isn't terrible, but it would be nice if it were more opaque from the users's point of view.

src/adera/Machines/UserControl.cpp
```diff
@@ -129,42 +107,79 @@ void SysMachineUserControl::update_sensor(ActiveScene &rScene)
...
-        {
-            throttlePos = std::clamp(throttlePos + delta, 0.0f, 1.0f);
-        }
-
-        if (usrCtrl.m_throttleLess.trigger_hold())
-        {
-            throttlePos = std::clamp(throttlePos - delta, 0.0f, 1.0f);
-        }
+        // Get the Attitude Control Panel which contains the Throttle Port
+        auto *panelAttCtrl = rScene.reg_try_get< ACompWirePanel<AttitudeControl> >(ent);
```
z-adams, line 160: Not a huge fan of the `fnc< AnotherTemplatedType<T> >()` formatting to be honest, kind of screws with my mental template disambiguator. I agree that template args are difficult to read when you have a ton of nesting but I don't think this makes it any easier to read; if it's really a problem then do
```cpp
SomeTemplateType<
    std::pair<A, B>,
    std::vector<T>,
    RandomType
   > myVariable
```
but inserting spaces around the angle brackets just makes them easy to mistake for comparison operators

src/adera/Machines/Rocket.cpp
```diff
@@ -225,6 +197,7 @@ void SysMachineRocket::update_physics(ActiveScene& rScene)
         rCompRb.m_inertiaDirty = true;
 
         // Perform resource consumption calculation
+        /*
```
z-adams, line 200: How much work would it take to get this stuff working again in the new system? I suppose it will muddy this PR a bit so I guess putting it off for a later one is fine.

z-adams (approved):

src/adera/Machines/UserControl.cpp
Capital-Asterisk: The machines we have so far only have one connection of each type. Future machines may have two percent outputs for example. Having an `SysWire::get_node<T>(index)` would mean obtaining the ACompWireNodes component multiple times in many cases. This will only save one line of code anyways.

For the vector, a `using` may make this cleaner
```cpp
template <typename WIRETYPE_T>
using NodeUpdates_t = std::vector< nodeindex_t<WIRETYPE_T> >;
```

src/adera/Machines/UserControl.cpp
Capital-Asterisk: I vaguely remember them causing compile errors before. They end up becoming bitshifts in some cases.

src/adera/Machines/RCSController.cpp
```diff
@@ -134,62 +110,104 @@ void SysMachineRCSController::update_construct(ActiveScene &rScene)
...
+                // TODO: RCS translation is not currently implemented, only
+                // rotation
+                Vector3 commandTransl = Vector3{0.0f};
+                Vector3 commandRot = nodeCommand.m_state.m_value.m_attitude;
+                Vector3 thrusterPos = transform.translation()
+                                    - compRb.m_centerOfMassOffset;
+                Vector3 thrusterDir = transform.rotation()
+                                    * Vector3{0.0f, 0.0f, 1.0f};
+
+                if (commandRot.length() > 0.0f || commandTransl.length() > 0.0f)
```
jonesmz, line 177: Parenthesis around sub-parts of complex conditionals.

src/adera/Machines/Rocket.cpp
Capital-Asterisk: A new system (a node update) would be needed for fuel flow which should, yes, go into a different PR. Nodes would be far more capable in handling multiple fuel tanks; I got this nice diagram if I haven't posted it before.
image

src/osp/Active/SysWire.h
```diff
@@ -116,261 +127,488 @@ namespace wiretype
...
+    {
+        WireNode<WIRETYPE_T> &rNode = rNodes.get_node(nodeIndex);
+
+        // Overwrite value with value to write
+        rNode.m_state.m_value = rNode.m_state.m_write;
+
+        // Add connected machine inputs to machToUpdate vectors, as they now
+        // require a calculation update
+        for (auto it = std::next(std::begin(rNode.m_links));
+             it != std::end(rNode.m_links); it++)
```
jonesmz, line 591: if you're going to use std::next and std::begin, then use std::advance(it) instead of it++.

src/adera/Machines/UserControl.cpp
z-adams: That shouldn't be happening in any situation where it's ambiguous, if it's ambiguous it will require the `template` keyword. Was there a specific place that happened?

src/osp/Active/SysWire.h
```diff
@@ -116,261 +127,488 @@ namespace wiretype
...
-inline WireOutput::WireOutput(IWireElement *element, WireOutput&& move)
- : LinkedList<WireInput>(std::move(move))
- , m_value(std::move(move.m_value))
- , m_element(element)
- , m_name(std::move(move.m_name))
-{ }
+    rNodes.m_needUpdate.clear();
+
+    // Commit machines to update to the scene-wide vectors
+    for (int i = 0; i < machToUpdate.size(); i ++)
```
jonesmz, line 602: for (size_t const i : std::views::iota(1u, machToUpdate.size()))

jonesmz: Or using the ranges-v3 library (most of the rest of which is pending inclusion into the stdlib for 23)
```cpp
for(auto && [rVec, mutex, entToCalculate] : ranges::zip(machToUpdate, rWire.m_entToCalculateMutex, m_entToCalculate))
{
    if( ! rVec.empty())
    {
        std::lock_guard<std::mutex> guard(mutex);
        vecset_merge(entToCalculate, rVec);
        rWire.request_update();
    }
}
```

src/osp/Resource/machines.h
```diff
@@ -25,4 +27,49 @@ struct RegisteredMachine
...
+using nodeindex_t = typename WireTypes<WIRETYPE_T>::node;
+
+template<class WIRETYPE_T>
+using linkindex_t = typename WireTypes<WIRETYPE_T>::link;
+
+template<class WIRETYPE_T>
+using portindex_t = typename WireTypes<WIRETYPE_T>::port;
+
+template<class TYPE_T>
+constexpr TYPE_T nullvalue() { return TYPE_T(std::numeric_limits<typename std::underlying_type<TYPE_T>::type>::max()); }
```
jonesmz, line 61: std::underlying_type_t<> ?

src/test_application/VehicleBuilder.cpp
```diff
@@ -80,17 +80,162 @@ partindex_t VehicleBuilder::add_part(
...
+
+std::pair<mach_t, osp::BlueprintMachine*> VehicleBuilder::machine_find(
+        osp::machine_id_t id, part_t part) noexcept
+{
+    if (id >= m_vehicle.m_machines.size())
+    {
+        return {osp::nullvalue<mach_t>(), nullptr}; // no machines of type
+    }
+
+    for (size_t i = 0; i < m_vehicle.m_machines[id].size(); i ++)
```
jonesmz, line 127: for(osp::BlueprintMachine& machineBp : m_vehicle.m_machines[id])
{

}

src/test_application/VehicleBuilder.cpp
```diff
@@ -80,17 +80,162 @@ partindex_t VehicleBuilder::add_part(
...
+        part_t part, mach_t mach, portindex_t<void> port)
+{
+    if (m_vehicle.m_wireNodes.size() <= id)
+    {
+        // No node of type is even registered
+        return osp::nullvalue< linkindex_t<void> >();
+    }
+
+    // Keep track of wire connections
+    auto nodeIt = m_nodeIndexMap.try_emplace({id, part, mach, port}, node);
```
jonesmz, line 170:
```cpp
auto const& [it, success] = m_nodesIndexMap.try_emplace({id, part, mach, port}, node);
if( ! success)
{
    return osp::nullvalue<linkindex_t<void>>();
}
```

src/test_application/VehicleBuilder.cpp
```diff
@@ -80,17 +80,162 @@ partindex_t VehicleBuilder::add_part(
...
+    rNode.m_links.emplace_back(
+            osp::BlueprintWireLink{config, uint32_t(part), uint16_t(mach),
+                                   uint16_t(port)});
+
+    // Update panel. There's no easy way to find the correct BlueprintWirePanel
+    // from m_vehicle.m_wirePanels from a wire index, part index, and machine
+    // index. We can search the vector if the right panel already exists, but
+    // here we're using a map
+    uint32_t newIndex = m_vehicle.m_wirePanels.size();
+    auto panelIt = m_panelIndexMap.try_emplace({id, part, mach}, newIndex);
```
jonesmz, line 193: use structured binding

src/test_application/VehicleBuilder.cpp
```diff
@@ -80,17 +80,162 @@ partindex_t VehicleBuilder::add_part(
...
+    uint16_t minPortCount = uint16_t(port) + 1;
+
+    if (panelIt.second)
+    {
+        // new panel entry added
+        m_vehicle.m_wirePanels.resize(std::max(m_vehicle.m_wirePanels.size(),
+                                               size_t(id + 1)));
+        m_vehicle.m_wirePanels[size_t(id)].emplace_back(
+                    osp::BlueprintWirePanel{uint32_t(part), uint16_t(mach),
+                                            uint16_t(minPortCount)});
```
jonesmz, line 204: why do you have so much casting going on here? minPortCount is already a uint16_t

src/test_application/VehicleBuilder.cpp
```diff
@@ -80,17 +80,162 @@ partindex_t VehicleBuilder::add_part(
...
+        part_t partOut, mach_t machOut, portindex_t<void> portOut,
+        part_t partIn, mach_t machIn, portindex_t<void> portIn)
+{
+    if (m_nodeIndexMap.find({id, partIn, machIn, portIn}) != m_nodeIndexMap.end())
+    {
+        // Input already connected
+        return osp::nullvalue< nodeindex_t<void> >();
+    }
+
+    auto nodeOutIt = m_nodeIndexMap.find({id, partOut, machOut, portOut});
```
jonesmz, line 228: structured bindings

src/test_application/VehicleBuilder.h
```diff
@@ -48,48 +71,85 @@ class VehicleBuilder
...
     osp::BlueprintVehicle&& export_move() { return std::move(m_vehicle); };
     osp::BlueprintVehicle export_copy() { return m_vehicle; };
 
 private:
+
+    // Map used to obtain a Machine's Panel
+    // panelIndex = m_panelMap[{id, part, machine}]
+    // panel = m_vehicle.m_wirePanels[panelIndex]
+    std::map<std::tuple<osp::wire_id_t, part_t, mach_t>,
+             uint32_t> m_panelIndexMap;
```
jonesmz, line 146: this really looks like you need boost::multi_index...

src/adera/Machines/RCSController.cpp
Capital-Asterisk: > I still strongly recommend against using mutexes for cross-thread communication.

How much of a problem will this be anyways? A conflict would only happen if two update functions finish at the same time and write to the same nodes which is unlikely.

> Instead, you should break the machines into threads based on some kind of logical partitioning...

This partitioning scheme sounds daunting and complex to implement, especially when noting connections across different wire types. Won't this also limit the number of threads per ship? How often would there even be non-connected parts of the ship? I don't even know where to start with this, nor will it even be any more performant than what I have now.

> If there does need to be some kind of communication between partition, then i recommend doing something like https://codereview.stackexchange.com/questions/219389/c-lock-free-single-producer-single-consumer-queue

Atomic-powered ring buffer? I'll keep this in mind.

> Or alternatively, using a dynamically added task in taskflow that has thread affinity for the thread that the machine in question is running on, or otherwise ensures that the machine in question isn't conncurrently running somehow.

Ordering tasks by using their thread IDs? I'm not sure what you're doing with the machines though.


Anyways, the wiring is already really fast and maximizing its performance isn't really something we need to deal with right now.

src/adera/Machines/Rocket.cpp
```diff
@@ -110,58 +71,72 @@ void SysMachineRocket::update_construct(ActiveScene& rScene)
...
+
+        // Check for nonzero throttle, continue otherwise
+        if (0 >= machine.m_powerOutput)
         {
             continue;
         }
 
+        /*
         // Check for adequate resource inputs
         bool fail = false;
```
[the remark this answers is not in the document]
Capital-Asterisk: A bit of this is likely to still be used later on

src/adera/Machines/UserControl.cpp
```diff
@@ -129,42 +107,79 @@ void SysMachineUserControl::update_sensor(ActiveScene &rScene)
...
-        float throttleRate = 0.5f;
-        auto delta = throttleRate * rScene.get_time_delta_fixed();
+        // Get the Percent Panel which contains the Throttle Port
+        auto *panelPercent = rScene.reg_try_get< ACompWirePanel<Percent> >(ent);
 
-        if (!machine.m_enable)
+        if (panelPercent != nullptr)
         {
-            continue;
+            WirePort<Percent> const *portThrottle
```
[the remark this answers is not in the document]
Capital-Asterisk: I remember consts not needing the Hungarian prefixes; I guess that's only for references?

The reasoning for using "port for throttle" instead of "throttle port" is that I usually put the common word within a sequence of similar variables first to improve readability. In this case there's `portThrottle` and `portAttitude`. It would be harder to tell the variables are similar if they were names `throttlePort` and `attitudePort`.

src/osp/Active/SysWire.h
```diff
@@ -116,261 +127,488 @@ namespace wiretype
...
-    // For use in move constructors / move operators of classes
-    // that aggregate WireInput. Use at own risk!!!
-    WireInput(WireInput&& move) = default;
-    WireInput& operator=(WireInput&& move) = default;
-
-    WireInput(WireInput const& copy) = delete;
-    WireInput& operator=(WireInput const& move) = delete;
+    WirePort<WIRETYPE_T> const* port(portindex_t<WIRETYPE_T> portIndex) const noexcept
+    {
+        if (m_ports.size() <= size_t(portIndex))
```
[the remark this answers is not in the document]
Capital-Asterisk: `portindex_t` is an enum class. It doesn't compile otherwise

src/adera/Machines/RCSController.cpp
jonesmz: > How much of a problem will this be anyways? A conflict would only happen if two update functions finish at the same time and write to the same nodes which is unlikely.

I think it will be a noticeable cause of performance issues in the future. Both for the single player game. As well as for multiplayer servers. E.g. multi tenancy related performance problems.

I work on graph processing systems at work all the time and mutexes are frequently a cause of subpar performance. In fact I spent between December and April rewriting a bunch of code to eliminate the use of mutexes entirely because perf testing showed the mutexes being a substantial cause of slowdowns.

My concern is that the wiring system is a **foundational** piece of code for the OSP engine. It needs to be designed very carefully with both single-circuit performance as well as scalability to many many thousands, even many hundreds of thousands of complex circuits. 

Ryzen style "alllllll the threads!" Chips are probably just the beginning of the future explosion of processing cores on a CPU. I very much don't predict that we'll see much growth in terms of single core frequency. So the only practical way to scale up is by scaling out, and mutexes are not going to give good performance for scaling out. Message passing using atomic single produces multi-consumer queues, or explicit job scheduling will scale MUCH better.

And don't think that the single-produces multi-consumer queue is a better solution than using explicit job scheduling (partitioning of the circuits). Its better than mutexes, but it has its own limitations.

jonesmz: > This partitioning scheme sounds daunting and complex to implement, especially when noting connections across different wire types. Won't this also limit the number of threads per ship?

You don't want more threads per ship. Making a single ship faster on the wall-clock at the cost of more CPUs doing things us counter productive. You want aggregate throughput. You want to mininize the per frame computation time for when you have 100,000 to 1,000,000 circuits in a scene. The way to do that is to have a circuit calculate to 100% completion in a single thread, with circuits that aren't connected being run in parallel on different threads.

Frankly I don't really think that there should be any cross circuit communications that take less than one frame to complete. So that the circuit can read whatever inputs might come from external to the circuit at the beginning of the frame, and then when the calculations are finished then that circuit doesn't get reactivated until the next frame regardless of events from outside the circuit.

 How often would there even be non-connected parts of the ship? I don't even know where to start with this, nor will it even be any more performant than what I have now.

jonesmz: > Anyways, the wiring is already really fast and maximizing its performance isn't really something we need to deal with right now.

Be careful not to build on top of a shakey foundation.

src/adera/Machines/UserControl.cpp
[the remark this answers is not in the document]
jonesmz: The reason for the p is to indicate its a pointer. Local constant variables don't need "c_" or anything.

src/adera/Machines/RCSController.cpp
Capital-Asterisk: Are you suggesting to optimize for numerous small unconnected ships/circuits? I'm expecting our common case to be less than 10 separate ships each with possibly thousands of machines and connections, ie. large crafts with many parts such as space stations. For the case of a large space battle, optimizing for thousands of simpler circuits does make sense.

One thread per ship is reasonably fast, but there will often be cases where one big complex ship will act like a bottleneck. Every other thread of the wire update would just have to wait for this one big ship to finish.

Though I do acknowledge that there are quite a few issues with my systems so far. Multiple threads would end up reading and writing off the same Node buffers. [Signal Nodes here](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/7d5dfb7983fd8bca6959b628bb150abf3fcf63b2/src/osp/Active/SysWire.h#L56
) introduce a false sharing cache miss. `m_read` is read of multiple threads, and `m_write` is written by one thread only while being on the same cache block.

Luckily, this is a good opportunity for me to get rid of the mutexes as well. Instead of just marking nodes for update after writing their values, all changes can just be recorded to a thread-local buffer or atomic queue (or both) that can be fed to the node update afterwards without needing a write member in the node itself.

Partitioning per-vehicle does sound kind of practical; the player would never need to know every circuit element in the world at the same time. The complications I suggest have to do with managing many separate buffers along side EnTT:
* Panel components can be kept unchanged, as they're read-only during wire update.
* Nodes are already stored in vectors, and would probably be easy to separate into each vehicle.
* Machine components may be problematic. They're stored in the registry and need to be modified across multiple threads. However, machines can be made stateless and use only nodes instead (this sounds a little extreme but may really be worth it). 

However, I still believe that a partition system will end up being far more complicated to manage without any or much of a speedup, perhaps even slower if a single ship can bottleneck the entire update.

I've already aggregating the problem, since each type of node and machine are stored in their own buffers. All the data is laid out to be easily processed: every rocket in the scene is one continuous buffer; every fuel tank in the scene is also one continuous buffer. It's simple and easy to manage, and it wouldn't be that hard to have multiple threads process parts of the same buffer if needed. My system doesn't care which vehicles each of these machines belongs to, the only data it's given is the ones that are important.

src/osp/Active/SysWire.h
Capital-Asterisk: I can almost hear a C programmer laughing at us right now with all these almost-pythonic ways of doing for loops. The cmake is still configured to C++17, and I don't intend to switch the project to C++20 in this PR however, and I'll just keep these in mind.

src/adera/Machines/RCSController.cpp
jonesmz: > Are you suggesting to optimize for numerous small unconnected ships/circuits? I'm expecting our common case to be less than 10 separate ships each with possibly thousands of machines and connections, ie. large crafts with many parts such as space stations. For the case of a large space battle, optimizing for thousands of simpler circuits does make sense.

My intention is to say that optimizing for a single ship will come at the cost of aggregate throughput. For example, halving the runtime of one ship, but at the cost of multiple ships having an aggregate runtime increase of 10% is worse over all.

The ship you describe that has many thousands of machines and connections, in my opinion, strongly implies that you are working at the wrong abstraction level. As we talked about, you're thinking like an electrical engineer, but i'm trying to think like a network engineer. Both are valid, but it's my opinion that everything you would want to do can be done with a more abstract model than what you've built, at an improved runtime performance.

Instead of operating at the level of and-gates and or-gates, the use of application specific elements / nodes / items where each such "gate" has application specific code such as multiplexers, routers, ALUs, and so on, provides exactly the same level of "you can make literally anything" power without having the runtime penalty that your approach has, while also:

- Making it much easier to do proper partitioning of disjoint components of the "circuit".
- Not preventing the end user of the game from being silly and plopping down and-gates and similar
- Making it substantially easier for modders and/or players to make interesting circuits without needing to, themselves, be electrical engineers.

For example, an ALU, under the hood, is simply a series of fundamental logic gates (I would know, i had to design one in verilog. ugh.) but no one wants to be designing ALUs in whatever janky visual editor we end up providing. They'd be much much happier plopping down an ALU and calling it good.

> One thread per ship is reasonably fast, but there will often be cases where one big complex ship will act like a bottleneck. Every other thread of the wire update would just have to wait for this one big ship to finish.

That's why we would need to provide to the end user some kind of de-sync item for the circuit to allow them to manually break their ship up into smaller pieces for sake of eliminating bottlenecks like that.

Even though you don't think it's a good idea, i still advocate for an approach that introduces more deliberate latency for gameplay reasons. Though I'm not as concerned about that.

> Luckily, this is a good opportunity for me to get rid of the mutexes as well. Instead of just marking nodes for update after writing their values, all changes can just be recorded to a thread-local buffer or atomic queue (or both) that can be fed to the node update afterwards without needing a write member in the node itself.

I also don't think that needing to mark individual nodes for update is a good strategy either. You'd be better off running the entire circuit based on whether the inputs to that circuit have had any action, and then running the circuit to completion. Doing things at the granularity of individual nodes is going to get expensive quickly, as there will probably end up being millions of nodes in a particularly large multiplayer server.

Capital-Asterisk: > My intention is to say that optimizing for a single ship will come at the cost of aggregate throughput. For example, halving the runtime of one ship, but at the cost of multiple ships having an aggregate runtime increase of 10% is worse over all.

I don't quite understand why this would be the case. There are N machines to update per frame, scattered across some unspecified number of ships. Ideally, if all the threads are always updating something without waiting, then it would take the same amount of time to finish all updates for that frame regardless of how they're divided between ships. Is there something here I'm missing?

Also note that this is a realtime simulation where all the ships run and update synchronously. They all need to calculate the right values for the physics update, and other systems that need to run afterwards. We can't just update one ship more or less times. There's a fixed minimum number of times its machines need to update.

> The ship you describe that has many thousands of machines and connections, in my opinion, strongly implies that you are working at the wrong abstraction level....

The example I used is a space station, let me elaborate. There would be constant resource flow everywhere: solar panels generating power, oxygen being consumed, fuel being processed etc... There would be a logic system to control some thrusters, docking ports, and maybe some robotic arms. And of course there are spacecraft docked to it, each with maybe a hundred parts. Reaching thousands of machines and connections doesn't sound unrealistic at all.

> Instead of operating at the level of and-gates and or-gates, the use of application specific elements / nodes / items where each such "gate" has application specific code...

This is exactly the level of abstraction I'm already working with. Logic gates are just simple examples. I'm not using booleans here; so far I have Percent and AttitudeControl. Have you looked at the RCSController? Multiplexers, routers, and ALUs is exactly what I have in mind to implement in the future. 

> That's why we would need to provide to the end user some kind of de-sync item for the circuit to allow them to manually break their ship up into smaller pieces for sake of eliminating bottlenecks like that.

When will there ever be a single ship with unconnected sections? I can't think of any practical example where this would happen. Having this de-sync item sounds like an unnecessarily complicated solution for an entirely avoidable problem.

jonesmz: > I don't quite understand why this would be the case. There are N machines to update per frame, scattered across some unspecified number of ships. Ideally, if all the threads are always updating something without waiting, then it would take the same amount of time to finish all updates for that frame regardless of how they're divided between ships. Is there something here I'm missing?

The cross-thread synchronization using atomic variables.

If you allow individual machines / nodes in side a single circuit to be updated on multiple threads simultaneously, you must synchronize them to prevent data races.

This means that for a simulation with a single ship, the calculations, spread across multiple threads, will complete more quickly in terms of actual human time, but at the cost of substantially more CPU time across all cores.

When you scale this up to thousands of ships, you're overall throughput for all ships goes down, because of the extra work to synchronize across threads.

> Also note that this is a realtime simulation where all the ships run and update synchronously. They all need to calculate the right values for the physics update, and other systems that need to run afterwards. We can't just update one ship more or less times. There's a fixed minimum number of times its machines need to update.

I don't quite follow where we were discussing updating one ship more or fewer times than another.

But again, i think that you're model is not a good fit if you need to update things multiple times in a single frame. The graph / circuit should be designed in such a way that it will reach a steady state on it's own.

Maybe we need to revisit the the design from a different direction. My proposed system several months ago had each node inside the circuit do a single calculate and a single propagate per frame.

Yours does multiple calculates and propagates per frame

But both of our models had a an external driver that visited the entire collection of nodes in a circuit and poked them one by one.

Maybe we need to instead consider a different approach of using a purely event based circuit, where the circuit receives an input, and that input makes its way through all of the other nodes in the circuit via direct function calls, until all activity in the circuit stops (whether that causes some kind of output or not is irrelevant). Any data storage happens entirely as an internal implementation detail of the nodes in question, and the external driver is ignorant of whether or not nodes store items in EnTT.

This also means that while links between items can be stored in EnTT, data that's passing from one node to another via a link would *not* be stored in EnTT, but instead would involve direct function calls to the next node.

This design allows for cycles (with the explicit requirement that there be some mechanism to prevent recursive function calls blowing out the stack), and it also has the attractive property of the circuit needing to be invoked only when there is an input for it to consume, and the circuit can run as long as it needs to because when it finishes, it's at a steady state until it receives it's next input.

jonesmz: > The example I used is a space station, let me elaborate. There would be constant resource flow everywhere: solar panels generating power, oxygen being consumed, fuel being processed etc... There would be a logic system to control some thrusters, docking ports, and maybe some robotic arms. And of course there are spacecraft docked to it, each with maybe a hundred parts. Reaching thousands of machines and connections doesn't sound unrealistic at all.

Sure, that's fair assessment of the number of components / machines that'll be on a particular ship / station. But I'm sceptical that they need to be synchronously connected.

So we have:

- Power generation
- Life support consumption
- Fuel consumption
- Robotics (docking, manipulation, whatever)
- Docked ships interconnected with all the above in some manner.

Power generation, as a "circuit" consumes a "fuel" item (actual fuel, or photons in the case of solar panels) and produces electricity.  That electricity does not need to then synchronously enter the circuit that consumes it. It can be asynchronously posted into some kind of buffer object via a cross-thread message, or a single-producer-single-consumer atomic queue.

The consuming circuit can then asynchronously consume the electricity in it's buffer until it runs out, and when it runs out, it can post an event back to the power producer circuit that it needs more juice.

Similarly, fuel consumption can be handled much the same way. The circuit that requires fuel doesn't need to synchronously pull fuel out of the storage tank. It can have a buffer that's sufficient for it's needs that gets asynchronously filled by the "fuel production" circuit, or "fuel storage" circuit.

Oxygen production and consumption. Water / coolant / sewage and so on, they all have plenty of opportunities for being made asynchronous.

And on a particularly large station, there's no need for all of the electricity circuits to be connected together in one giant rats nest of wires. You'd want a main-bus that has high voltage and current, and then transformers to down-convert the high voltage to a lower voltage for general usage by appliances. Those transformers can, themselves, be async, and represent an opportunity to partition a single station into multiple parallel-y processing circuits.

> This is exactly the level of abstraction I'm already working with. Logic gates are just simple examples. I'm not using booleans here; so far I have Percent and AttitudeControl. Have you looked at the RCSController? Multiplexers, routers, and ALUs is exactly what I have in mind to implement in the future.

I was evaluating it more for C++ code smells than for architectural concerns. My code reviews tend to happen while i'm getting ready for bed on my phone, so it's not always easy to grab a solid architectural understanding.

> When will there ever be a single ship with unconnected sections? I can't think of any practical example where this would happen. Having this de-sync item sounds like an unnecessarily complicated solution for an entirely avoidable problem.

See the proposal above with regards to producers asynchronously inserting into "buffers' (don't have to be buffers in the sense of an std::array, but more in the sense of a "This circuit has this many thingies available right now" counter variable.

Capital-Asterisk: > When you scale this up to thousands of ships, you're overall throughput for all ships goes down, because of the extra work to synchronize across threads.

How about limiting the number of threads, so there's less things to synchronize? Dividing loads by ship, or dividing by type wouldn't really matter then. I prefer not to run ships individually for other practical reasons I've already described.

> But again, i think that you're model is not a good fit if you need to update things multiple times in a single frame. The graph / circuit should be designed in such a way that it will reach a steady state on it's own.

There are multiple cycles of updates, but each update acts on a different set of machines and nodes each time.

For example, the default part vehicle updates like this when the player yaws, pitches, or rolls the vehicle:
1. Before wire update: MachineUserControl reads inputs for rotation, and requests to modify its AttitudeControl output.
2. Wire update starts
3. Node update: AttitudeControl node writes its new value, and marks connected MachineRCSControllers
4. Calculate Update: MachineRCSControllers update, each reading a new value in AttitudeControl, and requests to modify their connected Throttle outputs
5. Node update: Throttle nodes write new values, and mark the connected MachineRockets to update.
6. Calculate Update: MachineRockets get new throttle values, and calculates their thrust output
7. Wire update ends. No more updates requested. 

> Maybe we need to revisit the the design from a different direction. My proposed system several months ago had each node inside the circuit do a single calculate and a single propagate per frame.

1-frame delays are simply unacceptable for control systems, especially in aerospace. Hovering a drone, landing a booster, or simply controlling by hand would suffer greatly from delays. PID controllers are known to osscilate uncontrollably when there's delay in their feedback. 1-frame for the whole system may be acceptable, but 1 frame for every single machine connected along a line would stack up and become very problematic. This may be acceptable for a different game, but definitly not for level of realism I'm aiming for in this project.

> Maybe we need to instead consider a different approach of using a purely event based circuit, where the circuit receives an input, and that input makes its way through all of the other nodes in the circuit via direct function calls...

This was my original plan for the Urho3D repo. This approach only supports combinational logic, is difficult to manage, unpredictable, and butchers the benefits of having an entity component system by pretty much using it like a heap without the systems. There is also stuff like instruction cache misses everywhere among other issues. Allowing cycles would just be a huge headache. I already have a working system here; if there's a huge glaring flaw with it then it's entirely my fault.

jonesmz: > How about limiting the number of threads, so there's less things to synchronize? Dividing loads by ship, or dividing by type wouldn't really matter then. I prefer not to run ships individually for other practical reasons I've already described.

That just means that you reduce framerates because you're not maximizing CPU utilization.


> There is also stuff like instruction cache misses everywhere among other issues. 

I have no idea why you keep indicating that there are instruction cache misses. Where did you measure that?

Basically if you want to submit this, then submit it. I disagree that it's the right approach, but i don't have the time to produce code or examples for a better system beyond what i've already explained.

jonesmz: > 1-frame delays are simply unacceptable for control systems, especially in aerospace.

You keep getting hung up on this. Nothing prevented my original proposal from being run multiple times per frame, just like your system does.

Capital-Asterisk: > ... That electricity does not need to then synchronously enter the circuit that consumes it. It can be asynchronously posted into some kind of buffer object via a cross-thread message, or a single-producer-single-consumer atomic queue

All this just to avoid the 'problem' of being synchronous? Would a buffer really help with multiple machines consuming and producing a resource? From the way you've described it, it really sounds like the buffer object here is a Node.

I strongly believe that there is a really high performance cost in trying to asynchronously manage resource flows like this.

> And lets certainly not forget that unlike electricity, gas and liquid do have noticeable propagation delay properties in pipes / ducts, so treating them as electricity with a "nearly instant" transmission delay isn't an accurate paradigm

Long pipes are best handled with separate machines, not at the system level. There's been a few discussions with z-adams on this topic. Each Node becomes a point of instantaneous pressure when dealing with fluids. A pipe is some machine with some pressure on each end, and an internal state. In electronics, long wires (transmissions lines) are used often in radio frequency design as separate components. Every other connection is just instant.

> And on a particularly large station, there's no need for all of the electricity circuits to be connected together in one giant rats nest of wires. You'd want a main-bus that has high voltage and current, and then transformers to down-convert the high voltage to a lower voltage for general usage by appliances. Those transformers can, themselves, be async, and represent an opportunity to partition a single station into multiple parallel-y processing circuits.

It doesn't matter how cleanly they're connected, they're still connected nonetheless. The idea of having Nodes cleans up a lot of these issues, especially buses. I'm not sure how making these asynchronous would help with anything.

My system handles this issue like this:
1. Calculate update: machines will request to add or subtract resources from a Resource Node by accumulating requests into a thread-local buffer. This buffer is then somehow published to be exposed to the resource node update.
2. Node update: Resource Node will read each request and calculates its net loss or gain. It will then mark its connected producers or consumers to update to replenish or accept resources.

Capital-Asterisk: > I have no idea why you keep indicating that there are instruction cache misses. Where did you measure that?

I do admit this kind of came out of nowhere. You're implying the use of polymorphism for machines, which doesn't work well with the entity component system. Different types of machines would end up being updated in an unspecified order. A single thread would have to know about many different types of machines dynamically. In my system, one thread only updates one kind of machine. A thread updating rockets will only touch rockets.

> Basically if you want to submit this, then submit it. I disagree that it's the right approach, but i don't have the time to produce code or examples for a better system beyond what i've already explained.

I'm very confident that I've covered a wide enough range of use cases in a flexible and performant manner. It is at least superior to the previous wiring system that this PR replaces. The use of templates here would allow you to entirely implement your mechanics if you want. The 'signals' part of this PR is optional, and I'm working on separating it a bit right now.

jonesmz: > All this just to avoid the 'problem' of being synchronous? Would a buffer really help with multiple machines consuming and producing a resource? From the way you've described it, it really sounds like the buffer object here is a Node.
> I strongly believe that there is a really high performance cost in trying to asynchronously manage resource flows like this.

There's two related topics here.

- "multiple circuits being calculated simultaneously" and what the maximum acceptable load the game engine can handle in terms of number of circuits on a particular CPU at e.g. 60 frames per second.
- The runtime cost of a single very large circuit.

Using std::mutex, in either of the situations above, is 100% going to be less performant than some other (non-terribly designed) solution.

For the runtime cost of a single very large circuit, simulating different parts of the circuit on different CPUs sounds attractive because more CPU makes things go quicker. And that's true, up to a point, certainly.

However, that level of cross CPU communication is not going to work well. mutexes cause cache flush events because of the atomic writes, they cause blocking on threads (and potentially sleeping of threads if they can't acquire a lock right away), they cause poor predictability of performance, and framerates that vary maybe not wildly, but certainly more than they would need to.

They're also terrible from a density perspective.

The wall-clock runtime cost of a single very large circuit isn't really relevant when you have many thousands of circuits that all need to be calculated. The aggregate throughput of all circuits in that case is far more important.

The way you get good aggregate throughput for that kind of thing is by having as much "single threaded" calculation happen as you can, by ensuring that the parts of your algorithm / code that touch on other threads are isolated to carefully selected parts of the algorithm. E.g. manually injecting points into the circuits where an asynchrony happens.

> Long pipes are best handled with separate machines, not at the system level. There's been a few discussions with z-adams on this topic. Each Node becomes a point of instantaneous pressure when dealing with fluids. A pipe is some machine with some pressure on each end, and an internal state. In electronics, long wires (transmissions lines) are used often in radio frequency design as separate components. Every other connection is just instant.

That sounds similar to what I was saying, with different terminology.

> It doesn't matter how cleanly they're connected, they're still connected nonetheless. The idea of having Nodes cleans up a lot of these issues, especially buses. I'm not sure how making these asynchronous would help with anything.

It helps in terms of allowing the system to be simulated asynchronously in a predictable manner, to increase aggregate throughput of the overall circuit simulation code.

> My system handles this issue like this: ...........

Which can be a point of asynchronous by having that request / reply happen with a pair of atomic queues.

> I do admit this kind of came out of nowhere. You're implying the use of polymorphism for machines, which doesn't work well with the entity component system. 

I am not implying that currently no. My previous design used polymorphism, which really has no meaningful effect on the instruction cache, but my current concern is with the multi-threaded aspect, which doesn't inherently need polymorphism.

Why do you think polymorphism has any meaningful effect on the instruction cache? That is contrary to my experience in industry and I've never encountered anything to indicate this.

> Different types of machines would end up being updated in an unspecified order. A single thread would have to know about many different types of machines dynamically. In my system, one thread only updates one kind of machine. A thread updating rockets will only touch rockets.

I don't see how this is beneficial at all. What is motivating that as a design goal? Are you saying thread in the meaning of std::thread? If so, how are you handling a situation where there's more types of machines than there are hardware CPU cores? Are you planning to spawn a new std::thread for each of those? Or when taskflow is working for our needs, a new task for each type of machine? That's going to result in absolutely terrible performance.

Using EnTT / ECS is irrelevant, maybe even counter productive, for highly interconnected things like circuits. In an ECS, a particular system should run once, and only once, per frame. That's my opinion anyway, i suppose there could be exceptions to that. But for something like a circuit simulator, or a graph of nodes, ECS is not inherently a useful tool, and you *definitely* should not be doing one "system" per node type if it means needing to run the system over and over again until the nodes reach some steady state.

jonesmz: >  It is at least superior to the previous wiring system that this PR replaces.

I complete agree with you on this.

Capital-Asterisk: > However, that level of cross CPU communication is not going to work well. mutexes cause cache flush events because of the atomic writes, they cause blocking on threads (and potentially sleeping of threads if they can't acquire a lock right away), they cause poor predictability of performance, and framerates that vary maybe not wildly, but certainly more than they would need to.

I'm already in the process of getting rid of mutexes, as well as getting rid of multiple threads writing the same buffer (Nodes) simultaneously. Though the issue with my system is that it constantly needs to pass data between threads after each node or machine update, which happens multiple times per frame. Since each update already touches a different part of memory anyways (noting the part vehicle example), I'm not expecting this to be much of an issue.

> The wall-clock runtime cost of a single very large circuit isn't really relevant when you have many thousands of circuits that all need to be calculated. The aggregate throughput of all circuits in that case is far more important.

Again, numerous small simple circuits won't be our common case. Dividing a large circuit into smaller circuits won't be easy either. 

> Why do you think polymorphism has any meaningful effect on the instruction cache? That is contrary to my experience in industry and I've never encountered anything to indicate this.

I'm repeating things I recall from game developers in the context of ECS and data oriented design. Performance critical sections that rely on virtual functions can suffer from more serious cache misses from needing to load memory from many different types in some unspecified order. It's harder to predict which buffers of memory each function of individual machines read or write into.

> I don't see how this is beneficial at all. What is motivating that as a design goal? Are you saying thread in the meaning of std::thread? If so, how are you handling a situation where there's more types of machines than there are hardware CPU cores? ...

More clearly, each task only updates one type of machine. `MachineRCSController::update_calculate`, for example, will update values for every RCS Controller in the entire scene that needs to be updated. So this is a task for each type. Perhaps even multiple tasks to update the same type in parallel, if enough are marked for update. Where will this terrible performance come from?

> Using EnTT / ECS is irrelevant, maybe even counter productive, for highly interconnected things like circuits. In an ECS, a particular system should run once, and only once, per frame. That's my opinion anyway, i suppose there could be exceptions to that...

It really feels like we're trying to solve entirely different problems here.  I'm using EnTT here to store connections and data, but nothing else I'm doing here is specific to ECS. My scheme doesn't require ECS to work, but EnTT had made many things conveinent. Also, who says an ECS must only run once per frame? This "updating multiple times" thing I keep saying is actually just updaing everything in a certain order (again, noting the part vehicle example where everything is updated only once).

A circuit simulator would benefit from ECS to some extent. Elements in a circuit simulation can be easily represented as entities with components. Stuff like part numbers, temperature, power dissipation, and connections can be components. Implementing higher levels of interconnection is made easier by ECS as it provides a robust method of representing entities. Note that I'm able to tie together logic and resource flows all in one scheme, which would be harder to do without ECS.

jonesmz (approved):

PR #111, opened 2021-06-07
Cleanup Hierarchy and Add Entity Deferred Delete system

Capital-Asterisk:
This PR is mostly just clearing up a bit of technical debt by separating systems and components.

### Hierarchy changes

* Hierarchy functions are moved from ActiveScene to the new SysHierarchy
* ACompName was split off from ACompHierarchy

### Transform component changes

ACompTransform is now just a single Matrix4, as it was split into many other components:

* ACompTransformControlled - For components that are controlled by specific systems. (ie. entities with rigid bodies are fully controlled by physics)
* ACompTransformMutable - For controlled entities that allow their transform to be modified. (ie. to allow teleporting rigid bodies)
* ACompDrawTransform - Holds the world transform used only for drawing. These are only used for drawing. Also consider that this may be used for interpolation for rendering between physics frames.

### Deferred Delete System

Deleting entities right away (using `entt::basic_registry::destroy()`) can be problematic for future concurrency as it modifies every component pool the entity is associated with. To avoid any potential issues, the ACompDelete tag component was added. When added to entities, other systems can operate on them like destructors until it is fully deleted at the end of the frame.

Functions for deleting entities:
* `ActiveScene::mark_delete` adds the delete component to an entity
* `SysHierarchy::mark_delete_cut` Cuts an entity out of the hierarchy, then calls ActiveScene::mark_delete.

Two new systems:
* `SysHierarchy::update_delete` adds the ACompDelete component to the descendants of deleted hierarchy components. This allows any system to delete hierarchy entities while being decoupled from the hierarchy system.
* `ActiveScene::update_delete` finally deletes marked components at the end of the frame

These two are called directly in ActiveScene::update, as it's difficult to get FunctionOrder to put it at the end.

### Other changes

* Remove UserInputHandler from ActiveScene. Only UserControl needs it, so it's injected in flight.cpp instead.
* Add scene.h and drawing.h for components and similar data not bound to any particular system.

TODO: some more comments probably

### DIFF
--- a/src/adera/Machines/Rocket.cpp
+++ b/src/adera/Machines/Rocket.cpp
@@ -30,10 +30,11 @@
 #include "../Shaders/PlumeShader.h"
 
 #include <osp/Active/ActiveScene.h>
-#include <osp/Shaders/Phong.h>
+#include <osp/Active/SysHierarchy.h>
 #include <osp/Active/SysVehicle.h>
 #include <osp/Active/physics.h>
-#include "osp/PhysicsConstants.h"
+#include <osp/Shaders/Phong.h>
+#include <osp/PhysicsConstants.h>
 
 #include <iostream>
 
@@ -44,10 +45,12 @@ using adera::active::machines::MachineRocket;
 using osp::active::ActiveScene;
 using osp::active::ActiveEnt;
 using osp::active::ACompHierarchy;
+using osp::active::ACompName;
 using osp::active::EHierarchyTraverseStatus;
 using osp::active::ACompMachines;
 using osp::active::ACompRigidBody_t;
 using osp::active::ACompTransform;
+using osp::active::SysHierarchy;
 using osp::active::SysPhysics_t;
 
 using osp::active::ACompWirePanel;
@@ -258,9 +261,15 @@ void SysMachineRocket::attach_plume_effect(ActiveScene &rScene, ActiveEnt part,
 
     auto findPlumeHandle = [&rScene, &plumeNode](ActiveEnt ent)
     {
-        auto const& node = rScene.reg_get<ACompHierarchy>(ent);
-        static constexpr std::string_view nodePrefix = "fx_plume_";
-        if (0 == node.m_name.compare(0, nodePrefix.size(), nodePrefix))
+        static constexpr std::string_view namePrefix = "fx_plume_";
+        auto const* name = rScene.reg_try_get<ACompName>(ent);
+
+        if (nullptr == name)
+        {
+            return EHierarchyTraverseStatus::Continue;
+        }
+
+        if (0 == name->m_name.compare(0, namePrefix.size(), namePrefix))
         {
             plumeNode = ent;
             return EHierarchyTraverseStatus::Stop;  // terminate search
@@ -268,7 +277,7 @@ void SysMachineRocket::attach_plume_effect(ActiveScene &rScene, ActiveEnt part,
         return EHierarchyTraverseStatus::Continue;
     };
 
-    rScene.hierarchy_traverse(part, findPlumeHandle);
+    SysHierarchy::traverse(rScene, part, findPlumeHandle);
 
     if (plumeNode == entt::null)
     {
@@ -282,7 +291,7 @@ void SysMachineRocket::attach_plume_effect(ActiveScene &rScene, ActiveEnt part,
 
     // Get plume effect
     Package& pkg = rScene.get_application().debug_find_package("lzdb");
-    std::string_view plumeAnchorName = rScene.reg_get<ACompHierarchy>(plumeNode).m_name;
+    std::string_view plumeAnchorName = rScene.reg_get<ACompName>(plumeNode).m_name;
     std::string_view effectName = plumeAnchorName.substr(3, plumeAnchorName.length() - 3);
     DependRes<PlumeEffectData> plumeEffect = pkg.get<PlumeEffectData>(effectName);
     if (plumeEffect.empty())
--- a/src/osp/Active/ActiveScene.cpp
+++ b/src/osp/Active/ActiveScene.cpp
@@ -56,6 +56,11 @@ ActiveScene::~ActiveScene()
 void ActiveScene::update()
 {
     m_updateOrder.call(*this);
+
+    // TODO: FunctionOrder is way too inconvenient to deal with, so for now the
+    //       delete systems are called here
+    SysHierarchy::update_delete(*this);
+    update_delete(*this);
 }
 
 void ActiveScene::draw(ActiveEnt camera)
--- a/src/osp/Active/ActiveScene.cpp
+++ b/src/osp/Active/ActiveScene.cpp
@@ -48,11 +48,6 @@ ActiveScene::ActiveScene(OSPApplication &app, Package& context) :
     m_root = m_registry.create();
 }
 
-ActiveScene::~ActiveScene()
-{
-    m_registry.clear();
-}
-
 void ActiveScene::update()
 {
     m_updateOrder.call(*this);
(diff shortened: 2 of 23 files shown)


src/osp/Active/scene.h
```diff
@@ -0,0 +1,91 @@
...
+
+namespace osp::active
+{
+
+/**
+ * Component for transformation (in meters)
+ */
+struct ACompTransform
+{
+    osp::Matrix4 m_transform;
```
z-adams, line 40: So `ACompTransform` is now the local coordiates for each entity, and the global transform has moved to `ACompDrawTransform` and is confined to the drawing stuff?

z-adams:
SysRocket.cpp:209:
```cpp
auto const& rCompTf = rScene.reg_get<ACompTransform>(pRbAncestor->m_ancestor);
```
Just wanted to point out that this line requires the assumption that rigidbodies are always top-level hierarchy nodes, that aren't children of anything. Not a problem now, but may be in the future. It was already this way before this PR, I was just reminded of it. May need to change in the future for physics reference frames.

z-adams (approved):

src/adera/Machines/Rocket.cpp
```diff
@@ -258,17 +261,23 @@ void SysMachineRocket::attach_plume_effect(ActiveScene &rScene, ActiveEnt part,
...
-        if (0 == node.m_name.compare(0, nodePrefix.size(), nodePrefix))
+        static constexpr std::string_view namePrefix = "fx_plume_";
+        auto const* name = rScene.reg_try_get<ACompName>(ent);
+
+        if (nullptr == name)
+        {
+            return EHierarchyTraverseStatus::Continue;
+        }
+
+        if (0 == name->m_name.compare(0, namePrefix.size(), namePrefix))
```
jonesmz, line 272: what's wrong with operator= ?

src/osp/Active/ActiveScene.cpp
```diff
@@ -39,196 +40,41 @@ void ACompCamera::calculate_projection()
...
-
     m_root = m_registry.create();
-    ACompHierarchy& hierarchy = m_registry.emplace<ACompHierarchy>(m_root);
-    hierarchy.m_name = "Root Entity";
-
 }
 
 ActiveScene::~ActiveScene()
 {
     m_registry.clear();
```
jonesmz, line 53: Remind me why we need to call .clear()? the ActiveReg_t should be destructed here. Isn't that enough?

src/osp/Active/ActiveScene.cpp
```diff
@@ -39,196 +40,41 @@ void ACompCamera::calculate_projection()
...
+    update_delete(*this);
 }
 
 void ActiveScene::draw(ActiveEnt camera)
 {
+    SysHierarchy::sort(*this);
+    SysRender::update_hierarchy_transforms(*this);
+
     auto renderView = m_registry.view<ACompRenderingAgent, ACompPerspective3DView, ACompRenderer>();
     for (auto [e, agent, view, renderer] : renderView.each())
```
jonesmz, line 72: small preference to change "e" to something more intuitive.

src/osp/Active/ActiveScene.h
```diff
@@ -152,16 +110,33 @@ class ActiveScene
...
 
+
+    /**
+     * Mark an entity for deletion
+     *
+     * @param ent [in] Entity to delete
+     */
+    void mark_delete(ActiveEnt ent)
+    {
+        m_registry.emplace_or_replace<ACompDelete>(ent);
```
jonesmz, line 121: I feel like i've asked this before, but emplacement doesn't invalidate iterators to anything that's already in the registry right?

src/osp/Active/ActiveScene.cpp
z-adams: Could change it to `ent` I guess but this arrangement is used pretty much everywhere and if you're familiar with EnTT it's obvious since the first element of the `each()` closure is always the entity

src/osp/Active/ActiveScene.h
jonesmz: Assuming so, probably want to elaborate on the thread safety in a comment.

src/osp/Active/ActiveScene.h
```diff
@@ -152,16 +110,33 @@ class ActiveScene
...
+     * Delete entities with ACompDelete
+     *
+     * @param rScene [ref] scene with entities
      */
-    void update_hierarchy_transforms();
+    static void update_delete(ActiveScene& rScene)
+    {
+        ActiveReg_t &rReg = rScene.get_registry();
+        auto view = rReg.view<ACompDelete>();
+        rScene.get_registry().destroy(std::begin(view), std::end(view));
```
jonesmz, line 138: use rReg.destroy()

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -165,22 +165,25 @@ void SysAreaAssociate::sat_transform_set_relative(
 void SysAreaAssociate::floating_origin_translate(
         ActiveScene& rScene,Vector3 translation)
 {
+    auto &rReg = rScene.get_registry();
+
     auto view = rScene.get_registry()
```
jonesmz, line 170: use rReg.view<>()

src/osp/Active/SysHierarchy.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+void SysHierarchy::update_delete(ActiveScene& rScene)
+{
+    auto &rReg = rScene.get_registry();
+    auto viewDelHier = rReg.view<ACompDelete, ACompHierarchy>();
+
+    // copy entities to delete into a buffer
+    std::vector<ActiveEnt> toDelete;
+    toDelete.reserve(viewDelHier.size_hint());
+    toDelete.assign(std::begin(viewDelHier), std::end(viewDelHier));
```
jonesmz, line 142: is the call to reserve() actually useful? I suppose i would have expected .assign() to use std::distance() to do a single allocation.

src/osp/Active/SysHierarchy.h
```diff
@@ -0,0 +1,161 @@
...
+            // no children, move to next sibling
+            currentEnt = hier.m_siblingNext;
+        }
+        else if (parentNextSibling.size() > 0)
+        {
+            // last sibling, and not done yet
+            // is last sibling, move to parent's (or ancestor's) next sibling
+            currentEnt = parentNextSibling.top();
+            parentNextSibling.pop();
+        } else
```
jonesmz, line 153: newline

src/osp/Active/SysNewton.cpp
```diff
@@ -47,20 +49,21 @@ void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
...
 
     // Get ACompNwtBody. Every NewtonBody created here is associated with an
     // entity that contains one.
     auto *pBodyComp = static_cast<ACompNwtBody*>(NewtonBodyGetUserData(pBody));
     auto &rTransformComp = scene->reg_get<ACompTransform>(pBodyComp->m_entity);
 
     // Check if transform has been set externally
-    if (rTransformComp.m_transformDirty)
+    auto &tfMutable = scene->reg_get<ACompTransformMutable>(pBodyComp->m_entity);
+    if (tfMutable.m_dirty)
```
jonesmz, line 61: this is a nice pattern, if you're into it.
```cpp
if(std::exchange(tfMutable.m_dirty, false))
{
    NewtonBodySetMatrix(...);
}
```

src/osp/Active/ActiveScene.h
Capital-Asterisk: It only modifies the buffers associated with the templated component. Everything else stays untouched.

src/osp/Active/scene.h
```diff
@@ -0,0 +1,91 @@
...
+struct ACompTransformMutable{ bool m_dirty{false}; };
+
+/**
+ * Added to an entity to mark it for deletion
+ */
+struct ACompDelete{ };
+
+struct ACompName
+{
+    ACompName(std::string name) : m_name(name) { }
```
jonesmz, line 54: move

src/osp/Active/scene.h
```diff
@@ -0,0 +1,91 @@
...
+    std::string m_name;
```
jonesmz, line 55: This would be a good place for the shared_string class. Write-once, read-many.

jonesmz (approved):

src/osp/Active/ActiveScene.cpp
jonesmz: New contributors probably won't realize that right away.

Better to err on the side of welcoming new devs when it's low cost to do.

src/osp/Active/ActiveScene.cpp
Capital-Asterisk: Unfortunately, I can't really pin down why, but ACompDebugObject and DebugCameraController causes a crash on exit when this isn't here. The stack trace shows the error happening in a seemingly unrelated spot that seems to change when small parts of the code are changed. Clearly a sign of undefined behaviour going on.

It probably has something to do with the order at which ActiveScene's members are destructed. Something that DebugCameraController needs.

Capital-Asterisk: Ok, figured it out. It's the unreliable FunctionOrderHandle causing the issues. All of the UpdateOrderHandle_t and RenderOrderHandle_t have to be destructed first, or else they write into an already destructed FunctionOrder messing up something else.

Capital-Asterisk: I mentioned that FunctionOrder is so incredibly unreliably, that one should "hold a gun up to it at all times." Maybe I should actually shoot it next PR. Calling all the system update functions explicitly wouldn't really be a bad idea for the time being.

src/osp/Active/scene.h
Capital-Asterisk: Yes. It was never used anywhere else regardless.

src/adera/Machines/Rocket.cpp
jonesmz: My mistake. I misread your code. 

This is fine.

src/osp/Active/scene.h
Capital-Asterisk: I've considered this, but it doesn't compile on my system at least. You mentioned on the Discord that it might be an outdated version of the file. Would it be a good idea to fix this in this PR?

src/osp/Active/ActiveScene.cpp
Capital-Asterisk: Going to make this `const&` too

src/osp/Active/SysNewton.cpp
Capital-Asterisk: I'll keep this in mind. File has been moved in #112 however.

src/osp/Active/scene.h
jonesmz: hrmm, could you remind me what the compile errors you ran into were?

maybe i need to make a unit test folder for osp-magnum so we can start adding tests.

jonesmz: but if you want to fix it, please feel free. I don't know that i have a different copy really, it's been long enough that it might have gotten deleted from my local machine.

Capital-Asterisk: I'm not sure how to fix it; I put my attempt on [this branch.](https://github.com/Capital-Asterisk/osp-magnum/tree/hierarchy-improvements2-sharedstring)

PR #112, opened 2021-06-08
Decouple Newton Dynamics from physics

Capital-Asterisk:
Depends on #111

The PR entirely separates Newton Dynamics integration from the physics API through generic components. This makes it possible to switch physics engines at runtime as well.

The previous system relied on aliasing Newton-specific components. ie:
```cpp
// SysNewton.h
using ACompRigidBody_t = ACompNwtBody;
```

These new changes instead separate Newton data into their own components, making generic physics components for rigid bodies and colliders. SysNewton is then responsible for adding Newton-specific components to those entities.

New change:
```cpp
// physics.h
struct ACompRigidBody
{
    Vector3 m_inertia{1, 1, 1};
    Vector3 m_netForce{0, 0, 0};
    Vector3 m_netTorque{0, 0, 0};
...
};

// SysNewton.h
struct ACompNwtBody
{
...
private:
    NewtonBody const *m_body;
};
```

Static functions from SysNewton was moved into a new SysPhysics class. SysNewton was then moved to its own folder outside of `src/osp`.

The interface for managing rigid bodies from other systems is almost entirely unchanged.

### Internal SysNewton changes

ACompNwtBody and ACompNwtCollider hold pointers to Newton objects. Pointers are assigned in the constructor, and cannot be reassigned afterwards. The lifetime of their associated Newton objects are managed by these components, and are far easier to follow compared to the previous components, which instead used nullptr.


z-adams (approved):
Much cleaner, looks good

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+    // Get Associated entity
+    uintptr_t userData = reinterpret_cast<uintptr_t>(NewtonBodyGetUserData(pBody));
+    ActiveEnt ent = static_cast<ActiveEnt>(userData);
+
+    auto &rBody = scene->reg_get<ACompRigidBody>(ent);
+    auto &rTransformComp = scene->reg_get<ACompTransform>(ent);
+
+    // Check if transform has been set externally
+    auto &tfMutable = scene->reg_get<ACompTransformMutable>(ent);
+    if (tfMutable.m_dirty)
```
jonesmz, line 70: minor:

if(std::exchange(tfMutable.m_dirty, false))
{
    NewtonBodySetMatrix(...);
}

if you care for that style.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+                    .connect<&SysNewton::on_world_destruct>();
+}
+
+void SysNewton::update_world(ActiveScene& rScene)
+{
+    ActiveReg_t &rReg = rScene.get_registry();
+
+    // Try getting the newton world from the scene root.
+    // It is very unlikely that multiple physics worlds are ever needed
+    auto *worldComp = rReg.try_get<ACompNwtWorld>(rScene.hier_get_root());
```
jonesmz, line 114: pWorldComp

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+    if (worldComp == nullptr)
+    {
+        return; // No physics world component
+    }
+
+    if (worldComp->m_nwtWorld == nullptr)
+    {
+        // Create world if not yet initialized
+        worldComp->m_nwtWorld = NewtonCreate();
```
jonesmz, line 124: Why wouldn't the object creating the ACompNwtWorld be responsible for this?

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+    auto viewTf = rScene.get_registry().view<ACompTransform>();
+
+    // Apply transform changes after the Newton world(s) updates
+    for (ActiveEnt ent : viewBody)
+    {
+        auto &entNwtBody      = viewNwtBody.get<ACompNwtBody>(ent);
+        auto &entTransform    = viewTf.get<ACompTransform>(ent);
+
+        float mass, x, y, z;
+        NewtonBodyGetMass(entNwtBody.body(), &mass, &x, &y, &z);
```
jonesmz, line 173: I have a vague feeling that this should be a different "system", as if there *ARE* more than one physics world, we would want to apply the transformations after all of the world have updated.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+    ActiveReg_t &rReg = rScene.get_registry();
+    ActiveEnt nextChild = ent;
+
+    // iterate through siblings of ent
+    while(nextChild != entt::null)
+    {
+        auto const &childHeir = rScene.reg_get<ACompHierarchy>(nextChild);
+        auto const *childTransform = rScene.get_registry().try_get<ACompTransform>(nextChild);
+
+        if (childTransform != nullptr)
```
jonesmz, line 193: if(auto const *pChildTransform = rScene.get_registry().try_get<ACompTransform>(nextChild);
   pChildTransform != nullptr)

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+            if (rReg.all_of<ACompSolidCollider>(nextChild))
+            {
+                auto* childNwtCollider = rReg.try_get<ACompNwtCollider>(nextChild);
+
+                NewtonCollision const *collision;
+
+                if (nullptr == childNwtCollider)
+                {
+                    // No Newton collider exists yet
+                    collision = NewtonCreateSphere(nwtWorld, 0.5f, 0, NULL);
```
jonesmz, line 206: I get the feeling that allocating / creating these objects "as-needed" when they need to be used, instead of ensuring they are always created at the same time as the component if they are intended to be used later, is a pessimization.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+                    rReg.emplace<ACompNwtCollider>(nextChild, collision);
+                    ACompNwtCollider f(collision);
```
jonesmz, line 208: You construct a new ACompNwtCollider after emplacing one, but then don't use it? (goes out of scope immediately)

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+        {
+            Matrix4 identity{};
+
+            body = NewtonCreateDynamicBody(
+                        nwtWorld, rCompound, identity.data());
+            rReg.emplace<ACompNwtBody>(ent, body);
+        }
+
+        // this doesn't actually destroy, it decrements a reference count
+        NewtonDestroyCollision(rCompound);
```
jonesmz, line 279: Yikes...
Can't this be done using a smart pointer?

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: Hmm, good thought. The code creating the scene also has to specify adding Newton systems, so this isn't much of a problem. Though I guess it would be cleaner if it followed the same scheme as the other Newton components.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: True. Updating the physics world itself also doesn't modify ACompTransform, so separating into its own system is a good idea.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+void SysNewton::create_body(ActiveScene& rScene, ActiveEnt ent,
+                            NewtonWorld const* nwtWorld)
+{
+    ActiveReg_t &rReg = rScene.get_registry();
+
+    auto const& entHier = rReg.get<ACompHierarchy>(ent);
+    auto const& entTransform = rReg.get<ACompTransform>(ent);
+    auto const* entShape = rReg.try_get<ACompShape>(ent);
+
+    if ((! rReg.all_of<ACompSolidCollider>(ent)) || (entShape == nullptr))
```
jonesmz, line 241: if ( ! rReg.all_of<ACompSolidCollider>(ent) || (entShape == nullptr))

the ! doesn't have the same operator precedence issues that i'm normally concerned about

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+void SysNewton::on_shape_destruct(ActiveReg_t& reg, ActiveEnt ent)
+{
+    NewtonCollision const *shape = reg.get<ACompNwtCollider>(ent).collision();
+    NewtonDestroyCollision(shape); // make sure the shape is destroyed
+}
+
+void SysNewton::on_world_destruct(ActiveReg_t& reg, ActiveEnt ent)
+{
+    NewtonWorld const *world = reg.get<ACompNwtWorld>(ent).m_nwtWorld;
```
jonesmz, line 390: pWorld

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
...
+    if (world != nullptr)
+    {
+        // delete all collision shapes first to prevent crash
+        reg.clear<ACompNwtCollider>();
+
+        // delete all rigid bodies too
+        reg.clear<ACompNwtBody>();
+
+        NewtonDestroy(world); // make sure the world is destroyed
```
jonesmz, line 399: Can you use an std::unique_ptr with a custom deleter stored in the component to ensure destruction when the component is deleted?

src/newtondynamics_physics/SysNewton.h
```diff
@@ -0,0 +1,196 @@
...
+struct ACompNwtWorld
+{
+    NewtonWorld *m_nwtWorld{nullptr};
+    float m_timeStep{1.0f / 60.0f};
+};
+
+struct ACompNwtBody
+{
+    constexpr ACompNwtBody(NewtonBody const* body) noexcept
+     : m_body(body) { }
```
jonesmz, line 52: m_pBody

src/newtondynamics_physics/SysNewton.h
```diff
@@ -0,0 +1,196 @@
...
+     * mesh iterators.
+     * @param shape Shape component to store NewtonCollision* into
+     * @param start
+     * @param end
+     */
+    template<class TRIANGLE_IT_T>
+    static void shape_create_tri_mesh_static(
+            osp::active::ActiveScene& rScene, osp::active::ACompShape &rShape,
+            osp::active::ActiveEnt chunkEnt,
+            TRIANGLE_IT_T const& start, TRIANGLE_IT_T const& end);
```
jonesmz, line 97: TRIANGLE_IT_T && ?

src/newtondynamics_physics/SysNewton.h
```diff
@@ -0,0 +1,196 @@
...
+    using osp::Vector3;
+
+    ACompNwtWorld* nwtWorldComp = try_get_physics_world(rScene);
+    NewtonCollision* tree = newton_create_tree_collision(nwtWorldComp->m_nwtWorld, 0);
+
+    newton_tree_collision_begin_build(tree);
+
+    Vector3 triangle[3];
+
+    for (TRIANGLE_IT_T it = start; it != end; it += 3)
```
jonesmz, line 177: std::for_each ?

jonesmz (approved):

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: I'm not quite sure what this is doing here either.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: It at least ensures that only SysNewton is the one that calls Newton functions. Creating components are cheap, and when Newton functions are called together, there's a good chance all the memory they'll touch will be loaded into the cache less often.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: Newton does seem to have support to add custom allocate and free functions, but that's all I can find.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: It's probably more reliable to use entt destructors here, since components can be moved around often and are fully managed by the registry. Components tend to work better (or just 'more correct') when they're POD.

src/newtondynamics_physics/SysNewton.h
Capital-Asterisk: I'd probably just rework this whole thing. 'Triangle iterators' here are just a way to turn mesh data into a list of triangles. This is what Newton wants, but probably not what most other physics engines want. A better way to go upon this is just to pass all the mesh data in, and describe it normally with stride, offset, component counts, etc... I feel that something in Magnum::Trade already does this.

src/newtondynamics_physics/SysNewton.cpp
jonesmz: in terms of performance, i would wager that having it be pre-created, as opposed to checking if it's valid and creating it new if it's not, would be faster. One less branch.

src/newtondynamics_physics/SysNewton.cpp
jonesmz: I propose making helper functions that hide the fact that functions from newton dynamics are being called, but otherwise do all of the necessary initialization work.

That lets the system that my comment here is inside of just blow through all it's work without needing to check if initialization has been done properly.

Consider adding a check / assert for debug builds though.

src/newtondynamics_physics/SysNewton.cpp
jonesmz: My question was more along the line of making rCompound an std::unique_ptr with a custom deleter, or something of that nature.

src/newtondynamics_physics/SysNewton.cpp
jonesmz: std::unique_ptr vs entt destructors don't make a big different to me.

I just hate to see "manual" resource deletion when some form of pre-registered cleanup logic could be triggered as needed instaed.

Capital-Asterisk: RAII and data oriented don't go well together. We're also missing out on potential optimizations when the component isn't POD. Ideally, we shouldn't have strings and vectors in the components either.

The entt registry knows when the components are going to be deleted, and it offers this signal as a feature. However, I'd actually want to move this into its own "delete system" instead, as entt happens to delete components polymorphically.

jonesmz (approved):

PR #113, opened 2021-06-16
Restructure User Controls

Capital-Asterisk:
Depends on #112 

This PR:
* Cleans up UserInputHandler
  * Removed ButtonControlHandle, replaced with ControlSubscriber and EButtonControlIndex
* Changes how MachineUserControl works entirely
* Removed DebugObject
* Refactors DebugCameraController into a system and component

### UserInputHandler

UserInputHandler converts raw user input data into named controls.

Many type names were changed to be more descriptive, and the file's readability was improved.

#### Control handles

ButtonControlHandles and other control handles were kind of ugly to store, as each one has a pointer to their UserInputHandler. This means storing multiple ButtonControlHandles stores multiple pointers to UserInputHandler. This is intended to reference count controls, so they're only tested for when needed. Complexity would also get pretty high as more different kinds of handles are added.

The ControlSubscriber class was added instead, that takes the place of multiple control handles by reference counting multiple controls. ControlSubscriber can be used so subscribe to button controls which returns a EButtonControlIndex: an integer enum class used to get the state of the control.

#### Improved Polling

So far, individual controls are polled. Each one has to be checked each frame regardless of which ones have actually changed.

A bit of code was added to support polling a list of events instead. Each time a control state changes, it is added to a vector inside UserInputHandler. This is more useful for when there are numerous controls that aren't used often, such as in an editor.

### MachineUserControl

This machine was restructured to not use UserInputHandler at all. It now simply reads its members and interfaces them to wiring. It is the new CameraController's job to write new controls into this machine.

### CameraController

DebugObject didn't really have a good reason to exist, nor did it make anything more convenient. DebugCameraController is something new contributors might look at first, and it resembling nothing like anything else in the code is a bad thing.

The new CameraController.h/cpp cleans it up with a component and systems, like everything else should be. Its goals should be much clearer, as each behaviour is separated into its own system:
* update_vehicle: Modify the vehicle
* update_controls: Control the vehicle
* update_area: Move the scene area
* update_view: Position the camera


src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -0,0 +1,428 @@
+/**
```
jonesmz, line 1: Is this PR rebased properly? I thought this file was added in the previous PR?

src/osp/Active/SysPhysics.cpp
```diff
@@ -0,0 +1,281 @@
...
+        if (rootMass != nullptr) { localMass += rootMass->m_mass; }
+    }
+
+    for (ActiveEnt nextChild = rScene.reg_get<ACompHierarchy>(root).m_childFirst;
+        nextChild != entt::null;)
+    {
+        auto const &childHier = rScene.reg_get<ACompHierarchy>(nextChild);
+        auto const *childTransform = rScene.get_registry().try_get<ACompTransform>(nextChild);
+        auto const *massComp = rScene.get_registry().try_get<ACompMass>(nextChild);
+        Matrix4 childMatrix{Magnum::Math::IdentityInit};
```
jonesmz, line 147: Minor, but

```
Matrix4 childMatrix = chioldTransform != nullptr
                                  ? childTransform->m_transform
                                  : Magnum::Math::IdentityInit;
```
should avoid double initializing the childMatrix variable.

Could be the compiler is smart enough to do this regardless, using control flow analysis, but it's never obvious when the compiler is going to actually be that smart.

src/osp/Active/SysPhysics.cpp
```diff
@@ -0,0 +1,281 @@
...
+        }
+        // Else, assume identity transformation relative to parent
+
+        if (massComp != nullptr)
+        {
+            float childMass = rScene.reg_get<ACompMass>(nextChild).m_mass;
+
+            Vector3 offset = childMatrix.translation();
+
+            localCoM += childMass * offset;
```
jonesmz, line 161: i guess i don't quite follow why offset is it's own variable? not a big deal, just not how i would do it.

src/osp/Active/SysPhysics.cpp
```diff
@@ -0,0 +1,281 @@
...
+        // Recursively call this function to include grandchild masses
+        Vector4 subCoM = compute_hier_CoM(rScene, nextChild);
+        Vector3 childCoMOffset = childMatrix.translation() + subCoM.xyz();
+        localCoM += subCoM.w() * childCoMOffset;
+        localMass += subCoM.w();
+
+        nextChild = childHier.m_siblingNext;
+    }
+
+    if (!(localMass > 0.0f))
```
jonesmz, line 174: localMass <=

src/osp/Active/SysPhysics.cpp
```diff
@@ -0,0 +1,281 @@
...
+    {
+        // Massless subhierarchy, return zero contribution
+        return Vector4{0.0f};
```
jonesmz, line 177: omit type for consistancy with other return statements.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: Nope, this isn't properly rebased yet

src/osp/UserInputHandler.cpp
```diff
@@ -241,56 +135,60 @@ bool UserInputHandler::eval_button_expression(
...
 
-    //int index;
-
     // Check if the control was already created before
     if (cfgIt->second.m_enabled)
     {
         // Use existing ButtonControl
-        return ButtonControlHandle(this, cfgIt->second.m_index);
-        //m_controls[index].m_referenceCount ++;
+        size_t index = cfgIt->second.m_index;
```
jonesmz, line 170: size_t const index

src/osp/UserInputHandler.cpp
```diff
@@ -315,40 +213,41 @@ ButtonControlHandle UserInputHandler::config_get(std::string const& name)
...
+        return EButtonControlIndex(m_btnControls.size() - 1);
     }
 }
 
-MouseMovementHandle UserInputHandler::mouse_get()
+void UserInputHandler::button_unsubscribe(EButtonControlIndex index)
 {
-    return MouseMovementHandle(this);
-}
+    uint16_t &rRefCount = m_btnControls.at(size_t(index)).m_referenceCount;
```
jonesmz, line 227: i feel like this should instead be handled with an std::shared_ptr ?

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: Now it is

src/osp/UserInputHandler.h
```diff
@@ -34,168 +34,164 @@
...
-    bool m_triggered, m_held;
-    //std::array<ButtonVar, 10> m_vars;
-    ButtonExpr m_exprPress;
-    ButtonExpr m_exprRelease;
+/**
+ * @brief An invidual term in a boolean expression representing a (raw button)
+ */
+struct ControlTerm
+{
+    ControlTerm(ButtonMap_t::iterator button, EVarTrigger trigger,
```
jonesmz, line 101: For trivial constructors like this, you can just do direct member initialization.

E.g.

ControlTerm{button, trigger, invert, nextop};

jonesmz: This lets you omit the constructor entirely.

src/osp/UserInputHandler.h
```diff
@@ -348,23 +339,63 @@ class UserInputHandler
 
 private:
 
-    std::vector<ButtonMap> m_deviceToButtonRaw;
-    std::map<std::string, ButtonConfig> m_controlConfigs;
-    std::vector<ButtonControl> m_controls;
+    std::vector<ButtonMap_t> m_deviceToButtonRaw;
+    std::map<std::string, ButtonControlConfig, std::less<> > m_btnControlCfg;
```
jonesmz, line 343: Good opportunity for shared string.

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+using testapp::SysCameraController;
+using testapp::ACompCameraController;
+
+using adera::active::machines::MachineUserControl;
+
+using osp::universe::Satellite;
+using osp::universe::Universe;
+using osp::universe::UCompTransformTraj;
+using osp::universe::UCompVehicle;
```
jonesmz, line 44: Are all of these using statements really better than just namespacing things as they're used?

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+          rUni.get_reg().get<UCompTransformTraj>(rCamCtrl.m_selected).m_name);
+        return true;
+    }
+
+    return false;
+}
+
+void SysCameraController::update_vehicle(ActiveScene &rScene)
+{
+    auto const [ent, rCamCtrl] = get_camera_controller(rScene);
```
jonesmz, line 135: const&?

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+    Vector3s translate = Vector3s((camTf.translation() - rCamCtrl.m_orbitPos)
+                            / rCamCtrl.m_originDistanceThreshold)
+                       * rCamCtrl.m_originDistanceThreshold;
+
+    // convert to space int
+    translate *= osp::gc_units_per_meter;
+
+    if (!translate.isZero())
+    {
+      SPDLOG_LOGGER_TRACE(m_scene.get_application().get_logger(),
```
jonesmz, line 302: indentation

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+    {
+        comOset = vehicleTf.transformVector(pCompRb->m_centerOfMassOffset);
+    }
+
+    // Process control inputs
+
+    ControlSubscriber const& controls = rCamCtrl.m_controls;
+
+    // Arrow key rotation
+    float keyRotYaw = (controls.button_held(rCamCtrl.m_rt)
```
jonesmz, line 339: unnecessary parenthesis

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+                       - controls.button_held(rCamCtrl.m_lf));
+    float keyRotPitch = (controls.button_held(rCamCtrl.m_dn)
+                         - controls.button_held(rCamCtrl.m_up));
+
+    // Mouse rotation
+    Quaternion camRotate(Magnum::Math::IdentityInit);
+    if (rCamCtrl.m_controls.button_held(rCamCtrl.m_rmb)
+            || keyRotYaw || keyRotPitch)
```
jonesmz, line 347: indentation

src/test_application/CameraController.cpp
```diff
@@ -0,0 +1,421 @@
...
+    ACompAreaLink const *pAreaLink =  SysAreaAssociate::try_get_area_link(rScene);
+
+    if (pAreaLink == nullptr)
+    {
+        return entt::null; // Scene not connected to universe
+    }
+
+    auto const findIt = pAreaLink->m_inside.find(sat);
+
+    if (findIt != pAreaLink->m_inside.end())
```
jonesmz, line 412: if-initilizer

src/test_application/OSPMagnum.cpp
```diff
@@ -160,102 +171,97 @@ void testapp::config_controls(OSPMagnum& rOspApp)
...
 
-
 //Map for all the keys
 //The tuple is in this order: device, number, and hold/pressed
-using Key_t = OSPMagnum::KeyEvent::Key;
-using Mouse_t = OSPMagnum::MouseEvent::Button;
-using VarOp_t = osp::ButtonVarConfig::VarOperator;
-using VarTrig_t = osp::ButtonVarConfig::VarTrigger;
 
 typedef std::tuple<int, int> button_tuple;
```
jonesmz, line 177: consider using std::pair instead of std::tuple here. Less template machinery for the compiler to deal with.

src/osp/UserInputHandler.cpp
Capital-Asterisk: I guess it's possible, but the elements being reference counted are in a vector. I also don't want there to be a reference counted object for every single control, hence using integer indices and ControlSubscriber instead of the Handle classes.

src/osp/UserInputHandler.cpp
```diff
@@ -391,44 +293,58 @@ void UserInputHandler::event_raw(DeviceId deviceId, int buttonEnum,
 
 void UserInputHandler::update_controls()
 {
+
     // Loop through controls and see which ones are triggered
 
-    for (ButtonControl &control : m_controls)
+    for (auto it = std::begin(m_btnControls); it != std::end(m_btnControls);
+         std::advance(it, 1))
     {
```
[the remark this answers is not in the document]
Capital-Asterisk: It was either C++20 ranges, or the ranges-v3 library

src/osp/UserInputHandler.h
```diff
@@ -34,168 +34,164 @@
...
-struct ButtonVar;
-struct ButtonVarConfig;
-struct ButtonRaw;
+enum class EButtonControlEvent : uint8_t { TRIGGERED = 0, RELEASED = 1 };
 
-using ButtonMap = std::map<int, ButtonRaw>;
-using ButtonExpr = std::vector<ButtonVar>;
+enum class EButtonControlIndex : uint32_t
+{
+    NONE = std::numeric_limits<uint32_t>::max()
```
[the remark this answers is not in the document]
Capital-Asterisk: I'm using enum classes here as an integer type that can't accidentally be implicitly converted, similar to entities. A none/null/invalid is the only sentinel values needed. Would this be better as a separate constant instead?

`osp/Resource/machine.h` has this function:
```cpp
template<class TYPE_T>
constexpr TYPE_T nullvalue()
{
    return TYPE_T(std::numeric_limits<
                  typename std::underlying_type_t<TYPE_T>
                  >::max());
}
```

which may perhaps be moved into types.h, so `nullvalue<EButtonControlIndex>()` can be used instead?

src/osp/UserInputHandler.h
Capital-Asterisk: Not quite sure if this is the case. Strings here are matched with other strings, and no other class would really need to hold references to the strings stored in here.

src/test_application/CameraController.cpp
Capital-Asterisk: The code is much harder to follow if there's `osp::universe` and `osp::active` written everywhere. A few of these are only used in certain functions, I think they can be moved down into just the function scopes where they're used.

src/test_application/CameraController.cpp
```diff
@@ -333,38 +328,39 @@ void SysCameraController::update_view(ActiveScene &rScene)
...
 
-        float yaw = keyRotYaw * float(keyRotDelta);
-        float pitch = keyRotPitch * float(keyRotDelta);
-        if (rCamCtrl.m_controls.button_held(rCamCtrl.m_rmb))
-        {
-            yaw -= controls.get_input_handler()->mouse_state().m_smoothDelta.x();
-            pitch -= controls.get_input_handler()->mouse_state().m_smoothDelta.y();
-        }
+    // 180 degrees per second
+    Rad const keyRotDelta = 180.0_degf * rScene.get_time_delta_fixed();
```
jonesmz, line 341: Nice, using units arithemetic, win win.

jonesmz (approved):

EhWhoAmI (approved):

PR #114, opened 2021-06-16
Remove FunctionOrder

Capital-Asterisk:
Depends on #113 

Function order is incredibly unreliable and we don't actually need it. 

* Updating the scene is done in update_scene in flight.cpp, where all update functions are called. This is actually far easier to understand. Testing real parallelization or straight up implementing taskflow should be a breeze from here.
* The render order used by rendering pipeline is now just a vector of function pointers.


src/osp/Active/scene.h
```diff
@@ -44,6 +44,8 @@ struct ACompTransformControlled { };
 
 struct ACompTransformMutable{ bool m_dirty{false}; };
 
+struct ACompFloatingOrigin { };
```
Capital-Asterisk, line 47: Minor change, moved from activetypes.h

src/osp/Active/ActiveScene.h
```diff
@@ -37,21 +37,20 @@
 namespace osp::active
 {
 
+
```
jonesmz, line 40: Minor: extra newline?

jonesmz (approved):
Wow, this does a *lot* of cleanup and removal.

Nice!!!

z-adams (approved):

PR #122, opened 2021-07-06
Fix file encoding to UTF-8

jonesmz:


Capital-Asterisk (approved):
Something's up with @z-adams 's text editor encoding, just a heads up

PR #124, opened 2021-07-06
Implement Coordinate Spaces

Capital-Asterisk:
This PR adds "Coordinate Spaces" which changes how Satellites are positioned in the universe. The Trajectory class is also removed.

### Coordinate Spaces

Coordinate Spaces store a list of Satellites, and some way of positioning them in space, such as Cartesian XYZ integers.

An N-body system (like the one written by @z-adams ) requires XYZ positions, velocities, and accelerations buffered in a way that each vector component is in its own buffer  ([x, x, x, ...], [y, y, y, ...], [z, z, z, ...]). A patched conics system requires orbital parameters, such as semi-major axis, eccentricity, true anomaly, etc...

Coordinate spaces are free to position Satellites any way they want. Their buffers are exposed through Corrade strided array views to be read externally. These are referred to as "CComp" for coordinate components. All coordinate spaces must output CCompX, CCompY, and CCompZ as a standard interface, which are Cartesian 64-bit signed integers. 

In a way, this can be seen as a way to store ECS components in separate buffers each to be handled by their own systems. This is simply covering a use case that EnTT is not designed for.

Coordinate Spaces don't store the physical properties of the Satellite; only location and velocity. Other properties are handled by normal EnTT components.

### Trajectory

Coordinate Spaces don't modify or operate on their Satellites; they only store them. Modifying Coordinate Spaces are handled by Trajectories. Previously a class, these are planned to now be free functions that do some operation on the Coordinate Space such as step it forward in time.

Summary of old system:

capital-asterisk: _n-body move satellite? patched conics move satellite too? OHHH, Strategy Pattern!!!!! OWO!!!11_

### Other changes

* Active areas are broken, since there isn't yet a way to move satellites within a coordinate space. This will be fixed in subsequent PRs
* Rename Vector3s to Vector3g, since Vector3s refers to a vector with short integers. 
* Rename a few other things to fit conventions


jonesmz:
What does the "g" stand for in vector3g?
Maybe vector3d64_t?

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -0,0 +1,82 @@
...
+    // TODO: handle removal of satellites
+}
+
+void CoordspaceCartesianSimple::update_views(
+        CoordinateSpace& rSpace, CoordspaceCartesianSimple& rData)
+{
+
+    size_t const ccompCount = std::max({
+                ccomp_id<CCompSat>(),
+                ccomp_id<CCompX>(), ccomp_id<CCompY>(), ccomp_id<CCompZ>()});
```
jonesmz, line 61: Could you elaborate on the logic here? CCompSat seems like a different unit than the others.

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -131,9 +133,9 @@ void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3s translate)
     auto &areaPosTraj = pArea->get_universe().get_reg()
             .get<universe::UCompTransformTraj>(pArea->m_areaSat);
 
-    areaPosTraj.m_position += translate;
+//    areaPosTraj.m_position += translate;
```
jonesmz, line 136: commented out?

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -148,12 +150,12 @@ void SysAreaAssociate::sat_transform_set_relative(
             .get<universe::UCompTransformTraj>(relativeSat);
 
     // 1024 units = 1 meter
-    Vector3s posAreaRelative(transform.translation() * gc_units_per_meter);
+    Vector3g posAreaRelative(transform.translation() * universe::gc_units_per_meter);
```
jonesmz, line 153: const

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -0,0 +1,82 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2020 Open Space Program Project
```
jonesmz, line 3: 2021

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+void CoordspaceCartesianSimple::update_exchange(
+        Universe& rUni, CoordinateSpace& rSpace,
+        CoordspaceCartesianSimple& rData)
+{
+    rData.reserve(rData.size() + rSpace.m_toAdd.size());
+
+    auto view = rUni.get_reg().view<UCompCoordspaceIndex>();
+
+    for (auto const &[sat, pos, vel] : rSpace.m_toAdd)
```
jonesmz, line 40:
```
for(auto && [sat, pos, vel] : std::exchange(rSpace.m_toAdd, {}))
{

}
```

src/osp/CoordinateSpaces/CartesianSimple.h
```diff
@@ -0,0 +1,83 @@
...
+     * @brief Update buffer views to expose an CoordspaceCartesianSimple's
+     *        data to its associated CoordinateSpace
+     *
+     * @param rSpace [ref]
+     * @param rData  [ref]
+     */
+    static void update_views(
+            CoordinateSpace& rSpace, CoordspaceCartesianSimple& rData);
+
+    size_t size() const noexcept
```
jonesmz, line 64: constexpr?

src/osp/CoordinateSpaces/CartesianSimple.h
```diff
@@ -0,0 +1,83 @@
...
+        m_satellites.reserve(n);
+        m_positions.reserve(n);
+        m_velocities.reserve(n);
+    }
+
+    // Actual buffer data, these are carefully exposed in CoordinateSpace's
+    // m_components vector
+    std::vector<Satellite> m_satellites;
+    std::vector<Vector3g> m_positions;
+    std::vector<Vector3> m_velocities;
```
jonesmz, line 80: no action needed, but i wonder if we might not benefit in the future from some other data type that supports chunked allocation. I always worry that the total size of a vector will end up being gigantic, and the cost of the operating system finding a contiguous memory area large enough can be quite expensive.

src/osp/Universe.cpp
```diff
@@ -49,27 +53,61 @@ void Universe::sat_remove(Satellite sat)
...
+
+    if (frameInCoord.m_coordSpace == targetInCoord.m_coordSpace)
     {
-        // Same parent, easy calculation
-        return targetPosTraj.m_position - framePosTraj.m_position;
+        // Both Satellites are in the same coordinate system
+
+        std::optional<CoordinateSpace> const &rSpace
+                = m_coordSpaces.at(frameInCoord.m_coordSpace);
+
```
jonesmz, line 78: Need to verify / assert that the optional has a value before you use it in the line below.

src/osp/Universe.cpp
```diff
@@ -49,27 +53,61 @@ void Universe::sat_remove(Satellite sat)
...
+            viewX[targetCoordIndex.m_myIndex] - viewX[frameCoordIndex.m_myIndex],
+            viewY[targetCoordIndex.m_myIndex] - viewY[frameCoordIndex.m_myIndex],
+            viewZ[targetCoordIndex.m_myIndex] - viewZ[frameCoordIndex.m_myIndex],
+        };
+
+        return pos;
     }
-    // TODO: calculation for different parents
+
+    // TODO: calculation for positions across different coordinate spaces
```
jonesmz, line 92: Does this include objects that are transitioning from one coordinate space (which I think i once referred to as a sphere of influence a long time ago) to another, such as leaving orbit of a planet / moon, and thus transitioning to the orbit of the parent body (star / planet) ?

src/osp/Universe.cpp
```diff
@@ -49,27 +53,61 @@ void Universe::sat_remove(Satellite sat)
...
 
     return {0, 0, 0};
 }
 
 Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target) const
 {
-    // 1024 units = 1 meter. this can change
-    return Vector3(sat_calc_pos(referenceFrame, target)) / 1024.0f;
+    // 1024 units = 1 meter. TODO: will change soon
```
jonesmz, line 99: Remove comment and TODO

src/osp/Universe.cpp
```diff
@@ -49,27 +53,61 @@ void Universe::sat_remove(Satellite sat)
...
+    return Vector3(sat_calc_pos(referenceFrame, target)) / gc_units_per_meter;
```
jonesmz, line 100: Is the call to Vector3 constructor needed here?

If so, can this be replaced with {} ?

src/osp/Universe.cpp
```diff
@@ -49,27 +53,61 @@ void Universe::sat_remove(Satellite sat)
...
+}
+
+void Universe::update_sat_coordspace(uint32_t coordSpace)
+{
+    auto view = m_registry.view<UCompInCoordspace>();
+
+    for (auto const &[sat, pos, vel] : m_coordSpaces[coordSpace]->m_toAdd)
```
jonesmz, line 107: Don't need to clear after the loop?

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
-     * @tparam TRAJECTORY_T Type of Trajectory to construct
-     * @return Reference to new trajectory just created
+     * @brief Ressign indices in the UCompInCoordspace components of satellites
+     *        in a CoordinateSpace's m_toAdd queue
+     *
+     * @param coordSpace [in] Index to CoordinateSpace in m_coordSpaces
      */
-    template<typename TRAJECTORY_T, typename ... ARGS_T>
-    TRAJECTORY_T& trajectory_create(ARGS_T&& ... args);
+    void update_sat_coordspace(uint32_t coordSpace);
```
jonesmz, line 118: typedef for coordSpaceIndex

Or use size_t for indicies into std:: containers.

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
 private:
 
     Satellite m_root;
 
-    std::vector<std::unique_ptr<ISystemTrajectory>> m_trajectories;
-
-    Registry_t m_registry;
+    Reg_t m_registry;
 
 };
```
jonesmz, line 132: Not terribly important, but i find it's frequently very useful for other contributors to a codebase if the end of a class / struct has a comment with it's name.

e.g.

```
}; // class Universe
```

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
-    constexpr const Registry_t& get_reg() const noexcept
+    constexpr Reg_t& get_reg() noexcept { return m_registry; }
+    constexpr const Reg_t& get_reg() const noexcept
     { return m_registry; }
 
+    std::vector< std::optional<CoordinateSpace> > m_coordSpaces;
+
 private:
 
     Satellite m_root;
```
jonesmz, line 128: Could you go into a bit more detail, such as a comment here in the code, on the reason why a root satalite is needed, and how it's used in the code?

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
-    TRAJECTORY_T& trajectory_create(ARGS_T&& ... args);
+    void update_sat_coordspace(uint32_t coordSpace);
 
-    constexpr Registry_t& get_reg() noexcept { return m_registry; }
-    constexpr const Registry_t& get_reg() const noexcept
+    constexpr Reg_t& get_reg() noexcept { return m_registry; }
+    constexpr const Reg_t& get_reg() const noexcept
     { return m_registry; }
 
+    std::vector< std::optional<CoordinateSpace> > m_coordSpaces;
```
jonesmz, line 124: Move to private.



Also, storing a vector of std::optionals feels odd to me.  More deets?

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
-    Satellite m_parent; // Set only by trajectory
-
-    ISystemTrajectory *m_trajectory{nullptr}; // get rid of this some day
-    unsigned m_index; // index in trajectory's vector
-
-    // set this to true when you modify something the trajectory might use
-    bool m_dirty{true};
-
     // move this somewhere else eventually
     std::string m_name;
```
jonesmz, line 140: Reason to keep UCompTransformTraj if all it holds is a name, especially one that has a comment saying it's going to be moved eventually?

src/osp/Universe.h
```diff
@@ -67,225 +67,89 @@ class Universe
...
-};
 
-struct UCompVelocity
-{
-    Vector3 m_velocity;
+    Quaternion m_rotation;
```
jonesmz, line 142: Oh. My mistake. It also holds the rotation.

src/osp/universetypes.h
```diff
@@ -0,0 +1,44 @@
...
+namespace osp::universe
+{
+
+enum class Satellite : uint32_t {};
+
+using spaceint_t = int64_t;
+
+// 1024 space units = 1 meter
+// TODO: this should vary by trajectory, but for now it's global
+const float gc_units_per_meter = 1024.0f;
```
jonesmz, line 38: Were you thinking that it would be a template variable, with the template parameter being the trajectory type?

Or were you thinking that the trajectory would be a runtime lookup'ed value?

jonesmz: Could you write a little bit of documentation on how the units per meter varying by trajectory would be used in practice?

src/osp/coordinates.h
```diff
@@ -0,0 +1,144 @@
...
+struct CCompZ { using datatype_t = spaceint_t; };
+struct CCompSat { using datatype_t = Satellite; };
+
+template <typename CCOMP_T>
+using ViewCComp_t = Corrade::Containers::StridedArrayView1D<typename CCOMP_T::datatype_t>;
+
+template<typename CCOMP_T>
+constexpr ccomp_id_t ccomp_id() noexcept
+{
+    return ccomp_family_t::type<CCOMP_T>;
```
jonesmz, line 58: Why not make this a template variable, instead of a template function?
```
template<typename CCOMP_T>
inline constexpr ccomp_id_t ccomp_Id = ccomp_family_t::type<CCOMP_T>;
```

src/osp/coordinates.h
```diff
@@ -0,0 +1,144 @@
...
+}
+
+using CoordinateView_t = std::optional<Corrade::Containers::StridedArrayView1D<void>>;
```
jonesmz, line 61: Personally i don't enjoy when an std::optional is used as a typedef. Makes it hard to know if accessing the contents of the type is safe without checking first.

src/osp/coordinates.h
```diff
@@ -0,0 +1,144 @@
...
+ * kind of coordinate data, such as Cartesian XYZ or orbital parameters.
+ *
+ * As part of ECS, this is a way to store common components in separate buffers.
+ * that can each be managed by a specific system (simultaneously too).
+ * aka: a non-EnTT but still ECS component pool
+ *
+ */
+struct CoordinateSpace
+{
+    using SatToAdd_t = std::tuple<Satellite, Vector3g, Vector3>;
```
jonesmz, line 77: Std::tuple does have a pretty hefty compilation-time cost

not enough to justify changing this line, but I've found myself much more frequently using simple POD structs now that C++ has relatively convienient support for that.

You can even continue using structured bindings with the POD type if you define a specialization of get<> for it. https://devblogs.microsoft.com/oldnewthing/20201015-00/?p=104369

src/osp/coordinates.h
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * @brief TODO - commands to move and accelerate satellites
+     *
+     * @param cmd
+     */
+    void command(Command_t&& cmd)
+    {
+        m_commands.emplace_back(std::move(cmd));
```
jonesmz, line 109: For situations where there's no meaningful difference between the implementations of by const-ref and by-rvalue-ref, i normally just take the variable in question by value.

It has the advantage of, depending on the specific ABI of the platform, providing the compiler more opportunities for passing the data in registers.

src/osp/coordinates.h
```diff
@@ -0,0 +1,144 @@
...
+    ViewCComp_t<CCOMP_T> const ccomp_view() const
+    {
+        return Corrade::Containers::arrayCast<typename CCOMP_T::datatype_t>(
+                    m_components.at(ccomp_id<CCOMP_T>()).value());
+    }
+
+    // Queues for systems to communicate
+    std::vector<SatToAdd_t> m_toAdd;
+    std::vector<Satellite> m_toRemove;
+    std::vector<Command_t> m_commands;
```
jonesmz, line 131: If these were instead std::lists, there would be a smaller cost for adding new items to them.
As well as the toAdd list being splicable into the destination.

src/test_application/universes/planets.cpp
```diff
@@ -43,64 +44,81 @@ using osp::universe::Universe;
...
         Satellite sat = testapp::debug_add_part_vehicle(rUni, rPkg, "Placeholder Mk. I " + std::to_string(i));
 
-        auto &posTraj = rUni.get_reg().get<UCompTransformTraj>(sat);
-
-        posTraj.m_position = osp::Vector3s(i * 1024l * 4l, 0l, 0l);
-        posTraj.m_dirty = true;
+        auto &posTraj = rReg.get<UCompTransformTraj>(sat);
 
-        stationary.add(sat);
+        rSpace->add(sat, {i * 1024l * 4l, 0l, 0l}, {});
```
jonesmz, line 82: magic numbers?

src/test_application/universes/planets.cpp
```diff
@@ -43,64 +44,81 @@ using osp::universe::Universe;
...
-    posTraj.m_position = osp::Vector3s(22 * 1024l * 5l, 0l, 0l);
-    posTraj.m_dirty = true;
-    stationary.add(sat);*/
+    // Add Moon
+    {
+        Satellite sat = rUni.sat_create();
 
+        // Create the real world moon
+        float radius = 1.737E+6;
+        float mass = 7.347673E+22;
```
jonesmz, line 91: i take it this would be replaced by @EhWhoAmI 's solar-system-loader ?

jonesmz (approved):
Approved because I didn't see anything that I consider a problem.

I did leave feedback on several minor items that might inspire large-ish changes to the PR, but not enough to say I don't approve of the code.

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: This is calculating the size needed for the m_components vector, so each ccomp_id is a valid index. I should probably add a comment for this.

src/osp/Active/SysAreaAssociate.cpp
Capital-Asterisk: This part is currently broken, I'll fix it in the next PR

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: We're over half a year overdue for this one. Should these have been changed as each file was touched? Or how about in a separate PR?

src/osp/CoordinateSpaces/CartesianSimple.h
Capital-Asterisk: std::vector::size() isn't constexpr unfortunately

src/osp/CoordinateSpaces/CartesianSimple.h
Capital-Asterisk: In almost all of our cases, we never need to have vectors of large objects. ECS and SoA favours packing the small datatypes together, leading to more numerous but smaller vectors.

src/osp/Universe.cpp
Capital-Asterisk: I'm planning on making it guaranteed for it to never be possible for `frameInCoord.m_coordSpace` here to point to an invalid coordinate space.

src/osp/CoordinateSpaces/CartesianSimple.cpp
jonesmz: There are a few different strategies.

One is to do a once-yearly PR that updates all the timestamps.

The other is to update the files as the files get updated.

I guess just doing a once-yearly PR makes the most sense to me.

src/osp/Universe.cpp
Capital-Asterisk: Satellites in mid-transition (I personally call it a handoff) would probably be in an invalid state.

A Trajectory function would be responsible for determining when it's time to give a Satellite to a new coordinate space. It has to calculate the position of the Satellite relative to the new coordinate space (maybe a different function would be more efficient), then throw it into the new coordinate space's mailbox with add(...).

src/osp/Universe.cpp
Capital-Asterisk: This is a cast. sat_calc_pos returns 64-bit signed int Vector3g, which is casted to float and divided.

src/osp/Universe.cpp
Capital-Asterisk: gc_units_per_meter might not exist for much longer. CoordinateSpace::m_pow2scale is planned to determine units per meter, in a similar way I [described variable precision in the past](https://github.com/TheOpenSpaceProgramArchive/urho-osp/wiki/Dealing-with-a-Large-Universe) but more efficient.

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: Not sure if this is important, but if a file remains unchanged when the year is updated, the file becomes different even though it's functionally the same. This makes me prefer the latter option.

On the meantime, I'll add 2021 to only the newly added files

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: Found a better solution using a template function; will be using this instead.
```cpp
template<typename ... CCOMP_T>
constexpr size_t ccomp_min_size() noexcept
{
    return size_t(std::max({ccomp_id<CCOMP_T>() ...})) + 1;
}
```

src/osp/Universe.cpp
Capital-Asterisk: This is intended to be called before (possibly in parallel with) `CoordspaceCartesianSimple::update_exchange`, such as in simple.cpp line 134. It may be more practical to call clear separately to make this a bit more clear. Most of these should come together after implementing some kind of "universe update" in the following PRs.

src/osp/Universe.h
Capital-Asterisk: Good thought, it actually isn't needed at all and will be removed.

src/osp/Universe.h
Capital-Asterisk: Forgot to move this to private; left it in public just so it's easier to quickly throw the structure together. CoordinateSpaces are going to need to be added and removed, so I figured optionals can be used for deleted ones. Accessing by stable index is required and iteration speed is not important.

src/osp/Universe.h
Capital-Asterisk: It's pretty much dead right now. Name should be separated into a UCompName, and rotation can be stored in the coordinate spaces.

src/osp/universetypes.h
Capital-Asterisk: This is going to be assignable at runtime, configurable in config files. Should be useful consider for @EhWhoAmI too.

Going to link to [this old document again](https://github.com/TheOpenSpaceProgramArchive/urho-osp/wiki/Dealing-with-a-Large-Universe), but in short:
* A large galaxy can use (1 unit = 64 meters) to position stars. Who cares if stars are ± 64 meters away from each other? Milky way is about 10^21m in diameter. 64 bit int will have a max range of (2^64 * 64 = 1.18×10^21 m)
* Each solar system in that galaxy can use (1024 units = 1 meter), about 1 millimeter precision. Our solar system is about 2.9×10^11m in diameter. 64 bit ints will have a max range of (2^64 / 1024 = 1.8×10^16 m), enough for a bit of interstellar travel.

src/osp/CoordinateSpaces/CartesianSimple.cpp
jonesmz: that'd work

you can also make this an inline constexpr template variable, no need for the function syntax.

6-of-one, half-dozen-of-other.

src/osp/CoordinateSpaces/CartesianSimple.cpp
jonesmz: The reason for doing the yearly bump would just to have it done as a separate identifiable commit.

There's not a need to only modify the year when the files have changed, as the intention of modifying the date is to indicate that the over-all work is still being maintained, and not just the one file.

src/osp/coordinates.h
Capital-Asterisk: Small individually allocated POD types in std::lists really doesn't feel right. A reserved std::vector would outperform almost every case.

Considering concurrency, threads would need to push multiple elements into the queues all at once in its critical section; assuming this is why you're suggesting std::list. The best container I can think of is one of those lock-free atomic linked lists where each node is a block of multiple elements, so they aren't individually allocated.

src/osp/universetypes.h
jonesmz: My only concern with using 64bit ints for star positions would be how to handle interstellar travel, but I suppose one solution to that could be to have the ship travel via a series of massless stars that are positioned as a star, and are moved as the ship moves from one star to another.

src/osp/coordinates.h
jonesmz: You're right that an atomic producer / consumer queue wouldn't be a bad call.

However, in this case, i was specifically concerned about the number of items these lists could have at a time.

You seem pretty satisfied that we're not going to get many at a time, so it's a mute point, and i agree that std::vector is a good choice.

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -47,36 +47,30 @@ void CoordspaceCartesianSimple::update_exchange(
         rData.m_velocities.push_back(vel);
     }
 
-    rSpace.m_toAdd.clear();
```
jonesmz: does m_toAdd get cleaned up by something external to this function?

My reason for suggesting the use of std::exchange() was to limit the lifetime of the data held by m_toAdd.

But if the data is going to be cleaned up soon after, it's fine to not immediately clear it.

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -47,36 +47,30 @@ void CoordspaceCartesianSimple::update_exchange(
...
+    rSpace.m_components.resize(
+                ccomp_min_size<CCompSat, CCompX, CCompY, CCompZ>());
 
-    rSpace.m_components[ccomp_id<CCompSat>()] = {
+    rSpace.m_components[ccomp_id<CCompSat>] = {
                 rData.m_satellites, satCount,
                 sizeof(Satellite)};
 
-    rSpace.m_components[ccomp_id<CCompX>()] = {
+    rSpace.m_components[ccomp_id<CCompX>] = {
```
jonesmz, line 65: I almost wonder if we should use the std:: namespace convention of using a _v suffix for template variables like this.

src/osp/Universe.cpp
jonesmz: Ah, i see. I misread the parenthesis.

src/osp/Universe.cpp
```diff
@@ -100,7 +95,15 @@ Vector3 Universe::sat_calc_pos_meters(Satellite referenceFrame, Satellite target
     return Vector3(sat_calc_pos(referenceFrame, target)) / gc_units_per_meter;
 }
 
-void Universe::update_sat_coordspace(uint32_t coordSpace)
+std::pair<coordspace_index_t, CoordinateSpace&> Universe::coordspace_create()
+{
+    coordspace_index_t const index = m_coordSpaces.size();
+    return {index, *m_coordSpaces.emplace_back(CoordinateSpace{})};
+
+    // TODO: find deleted spaces instead of emplacing back each time
```
jonesmz, line 103: Maybe the function that removes items should compact the items in the vector, so that after removal, all the empty spots are at the end?

src/osp/Universe.h
```diff
@@ -69,7 +69,7 @@ class Universe
 
     using Reg_t = entt::basic_registry<Satellite>;
 
-    Universe();
+    Universe() = default;
     Universe(Universe const &copy) = delete;
     Universe(Universe &&move) = delete;
```
jonesmz, line 74: recommend deleting the operator= functions too, for consistency.

jonesmz: I like to define a macro OSP_DISABLE_COPY_AND_MOVE(Universe) that just deletes all of em.

src/osp/coordinates.h
```diff
@@ -52,14 +52,25 @@ struct CCompSat { using datatype_t = Satellite; };
...
+using CoordinateView_t = Corrade::Containers::StridedArrayView1D<void>;
+
+using coordspace_index_t = uint32_t;
+
+/**
+ * Runtime-generated sequential ID for a CComp
+ */
 template<typename CCOMP_T>
-constexpr ccomp_id_t ccomp_id() noexcept
+inline ccomp_id_t ccomp_id = ccomp_family_t::type<CCOMP_T>;
```
jonesmz, line 63: Wouldn't you want this to be constexpr?

Capital-Asterisk: Putting constexpr here caused a compile error. The type variable itself isn't constexpr, and is incremented at runtime.

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: These two lines in [simple.cpp line 133](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/884376500efb7162531c28e609d7c890dc41e155/src/test_application/universes/simple.cpp#L133) act like two different tasks that depend on m_toAdd.
```cpp
rUni.coordspace_update_sats(coordIndex);
CoordspaceCartesianSimple::update_exchange(rUni, rSpace, *pData);
```

 It would be odd if one cleared the queue, and the other didn't, as they don't modify the same data and can be run in parallel eventually. This is soon intended to be part of the 'universe update function' to be added very soon.

src/osp/coordinates.h
jonesmz: Ah. Hrmmm.

That could potentially mean a race-condition across multiple threads. But I haven't thought about it carefully.

src/osp/CoordinateSpaces/CartesianSimple.h
jonesmz: ah... it becomes constexpr inC++20

src/osp/Universe.cpp
jonesmz: Gotcha. Then an assert would be the right call.

src/osp/Universe.cpp
jonesmz: Do you know where you'd like that to happen in the per-frame order of operations?

src/osp/Universe.h
jonesmz: would the deleted coordinate spaces be left as empty optionals until a later point in the frame? When does the vector get compacted, i guess is my question?

EhWhoAmI (approved):

src/osp/Universe.cpp
Capital-Asterisk: This is going to be put in the upcoming universe updates. I guess there are two main steps:
1. Trajectory function moves the Satellites, and also determine which satellites need to be handed off to other coordinate systems using queues
2. Exchange (like CoordspaceCartesianSimple::update_exchange) would add and remove the satellites coordinates from their buffers

src/osp/Universe.cpp
Capital-Asterisk: The the element indices have to be stable however, so compacting cannot be done.

src/osp/coordinates.h
Capital-Asterisk: I think it probably does; EnTT has an option to make families atomic

z-adams (approved):

PR #128, opened 2021-07-24
Universe Updates and improved Active Area

Capital-Asterisk:
This PR progresses towards a more sophisticated Universe required to meet the project's goals.

Interactions between the ActiveScene and Universe were not that well thought out before, and gets rather ugly when considering asynchronous updates and rotating planets among other possible features.

### Universe Updates

The Universe now relies only on queues (just std::vectors for now) for communication with the ActiveScene. 

Updating the universe appears to have two main steps:
* Exchange update: Add, remove, and transfer Satellites between coordinate systems
* Trajectory update: Step through time and move the Satellites (none yet implemented)

The update function is called each frame for now. Since there are many possible test universes (only 'simple' and 'moon' so far), the update function is handled by an std::function that can be a injected into the application.

See [common.cpp](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/universes/common.cpp#L43) and [simple.cpp](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/universes/simple.cpp#L131)

There is also the possibility of some parts of the universe updating more frequently than others. ie. solar system's moves around the galaxy once every day, solar system updates every second, ActiveArea can move every 2 frames.

### ActiveArea Capture and Domain CoordinateSpaces

Imagine there's a solar system in an N-body simulation being explored by an ActiveArea.

The ActiveArea is moving itself; it has no mass, not affected by gravity, and has nothing to do with the N-body simulation whatsoever. This means its position data should NOT be in the N-body's coordinate space. Instead, the ActiveArea can live in its own coordinate space layered over top of the N-body coordinate space. This is implemented as the "Domain CoordinateSpace."

Now a vehicle enters the area, and can get modified by the user. This means the area now has full control over the vehicle Satellite's position, and should no longer be part of the N-body coordinate space. The ActiveArea can steal the vehicle, and store it in its own "Capture CoordinateSpace" instead.

The creation of both of these coordinate spaces can be found in [flight.cpp](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/flight.cpp#L321)

Capturing a Satellite (or just transferring Satellites between CoordinateSpaces) is done by queuing remove() from the source, and add() to the destination.
https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/osp/Satellites/SatActiveArea.cpp#L113-L114

### Coordinate Space Transforms

Satellites need to be handed off from one coordinate space to another. For example:
* between spheres of influence in a patched conics simulation
* from space to a rotating planet and its surface
* multiple coordinate spaces of varying precision for gigantic universe sizes

So far, 1024 space units is 1 meter globally. The upcoming system is similar to [the one planned for urho-osp](https://github.com/TheOpenSpaceProgramArchive/urho-osp/wiki/Dealing-with-a-Large-Universe#precision), but each CoordinateSpace having a precision, instead of individual Satellites.

The transformation required to transfer a Satellite between CoordinateSpaces is a translation, rotation, and scale. This can be represented by a Matrix4, but precision may be an issue due to converting between floating points and 64-bit ints. Instead, this solution takes a functional-style approach that uses only integers:

https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/osp/coordinates.h#L228

[Example usage](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/flight.cpp#L225) for when the scene is closed, and the ActiveArea is about to be deleted. CoordspaceTransform is used to transform positions from the Capture coordinate space back into the 'main' coordinate space.

The idea works for now, but is untested and does not yet support rotation.

Rotation is a rare case, only really useful for planet surfaces. Using floating points to rotate coordinates (such as with Quaterinions) should be acceptable.

### Improved Scene-Universe Synchronization

Before, SysAreaAssociate (system in ActiveScene) was responsible for calculating distances to nearby Satellites. This is impractical, as this task should be handled during the Universe update, which it now is.

[SatActiveArea.cpp](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/osp/Satellites/SatActiveArea.cpp#L55) now does the scanning, and uses queues in [UCompActiveArea](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/osp/Satellites/SatActiveArea.h#L73) to communicate with SysAreaAssociate.

Moving the scene and area is now a more robust 3-step process.
1. [CameraController requests to move](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/CameraController.cpp#L283) when the camera gets too far from the scene origin. this writes into UCompActiveArea's m_requestMove
2. Universe update [moves the ActiveArea in its Domain CoordinateSpace](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/test_application/universes/common.cpp#L87)
3. [SysAreaAssociate notices the area moves, and translates all floating origin entities in the scene](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/f486e6842524b87294e84fa0c7a9b46bb3514167/src/osp/Active/SysAreaAssociate.cpp#L74)

Step 1 can be replaced by other external requests.

### Splitting SysVehicle

SysVehicle.cpp was getting a bit crammed, so most of the ActiveScene-Universe synchronization was moved to SysVehicleSync.h/cpp instead.

### DIFF
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -39,7 +39,7 @@ using osp::universe::UCompActiveArea;
 
 using osp::universe::Vector3g;
 
-void SysAreaAssociate::update_scan(ActiveScene& rScene)
+void SysAreaAssociate::update_clear(ActiveScene& rScene)
 {
     ACompAreaLink *pArea = try_get_area_link(rScene);
 
@@ -54,46 +54,8 @@ void SysAreaAssociate::update_scan(ActiveScene& rScene)
 
     // Clear enter/leave event vectors
 
-    pArea->m_enter.clear();
-    pArea->m_leave.clear();
-
-    // Deal with activating satellites that have a UCompActivationRadius
-    // Satellites that have an Activation Radius have a sphere around them. If
-    // this sphere overlaps the ActiveArea's sphere 'm_areaRadius', then it
-    // will be activated.
-
-    auto viewActRadius = rUni.get_reg().view<
-            UCompActivationRadius, UCompActivatable>();
-
-    for (Satellite sat : viewActRadius)
-    {
-        auto satRadius = viewActRadius.get<UCompActivationRadius>(sat);
-
-        // Check if already activated
-        auto found = pArea->m_inside.find(sat);
-        bool alreadyInside = (found != pArea->m_inside.end());
-
-        // Simple sphere-sphere-intersection check
-        float distanceSquared = rUni.sat_calc_pos_meters(areaSat, sat).dot();
-        float radius = areaUComp.m_areaRadius + satRadius.m_radius;
-
-        if ((radius * radius) > distanceSquared)
-        {
-            if (alreadyInside)
-            {
-                continue; // Satellite already activated
-            }
-            // Satellite is nearby, activate it!
-            auto entered = pArea->m_inside.emplace(sat, entt::null).first;
-            pArea->m_enter.emplace_back(entered);
-        }
-        else if (alreadyInside)
-        {
-            // Satellite exited area
-            pArea->m_leave.emplace_back(found->first, found->second);
-            pArea->m_inside.erase(found);
-        }
-    }
+    areaUComp.m_enter.clear();
+    areaUComp.m_leave.clear();
 }
 
 ACompAreaLink* SysAreaAssociate::try_get_area_link(ActiveScene &rScene)
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -39,23 +39,40 @@ using osp::universe::UCompActiveArea;
 
 using osp::universe::Vector3g;
 
-void SysAreaAssociate::update_clear(ActiveScene& rScene)
+void SysAreaAssociate::update_consume(ActiveScene& rScene)
 {
-    ACompAreaLink *pArea = try_get_area_link(rScene);
+    ACompAreaLink *pLink = try_get_area_link(rScene);
 
-    if (pArea == nullptr)
+    if (pLink == nullptr)
     {
         return;
     }
 
-    Universe &rUni = pArea->m_rUniverse;
-    Satellite areaSat = pArea->m_areaSat;
-    auto &areaUComp = rUni.get_reg().get<UCompActiveArea>(areaSat);
+    Universe &rUni = pLink->m_rUniverse;
+    Satellite areaSat = pLink->m_areaSat;
+    auto &rAreaUComp = rUni.get_reg().get<UCompActiveArea>(areaSat);
 
-    // Clear enter/leave event vectors
+    pLink->m_enter = std::exchange(rAreaUComp.m_enter, {});
+    pLink->m_leave = std::exchange(rAreaUComp.m_leave, {});
 
-    areaUComp.m_enter.clear();
-    areaUComp.m_leave.clear();
+    Vector3g deltaTotal{};
+
+    for (Vector3g const delta : std::exchange(rAreaUComp.m_moved, {}))
+    {
+        deltaTotal += delta;
+    }
+
+    pLink->m_move = Vector3(deltaTotal) / universe::gc_units_per_meter;
+}
+
+void SysAreaAssociate::update_translate(ActiveScene &rScene)
+{
+    ACompAreaLink const *pLink = try_get_area_link(rScene);
+
+    if ( ! pLink->m_move.isZero())
+    {
+        floating_origin_translate(rScene, pLink->m_move * -1.0f);
+    }
 }
 
 ACompAreaLink* SysAreaAssociate::try_get_area_link(ActiveScene &rScene)
@@ -76,13 +93,6 @@ void SysAreaAssociate::connect(ActiveScene& rScene, universe::Universe &rUni,
 
 }
 
-void SysAreaAssociate::disconnect(ActiveScene& rScene)
-{
-    // TODO: Have a vector of function pointers for each satellite type, that
-    //       are called to de-activate them right away.
-    //       Also delete the ACompAreaLink from the scene
-}
-
 void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g translate)
 {
     ACompAreaLink *pArea = try_get_area_link(rScene);
@@ -92,37 +102,14 @@ void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g translate)
         return;
     }
 
-    auto &areaPosTraj = pArea->get_universe().get_reg()
-            .get<universe::UCompTransformTraj>(pArea->m_areaSat);
+    Universe &rUni = pArea->get_universe();
+    auto &rAreaSat = rUni.get_reg().get<UCompActiveArea>(pArea->m_areaSat);
 
-//    areaPosTraj.m_position += translate;
-
-    Vector3 meters = Vector3(translate) / universe::gc_units_per_meter;
-
-    floating_origin_translate(rScene, -meters);
-}
-
-void SysAreaAssociate::sat_transform_set_relative(
-        universe::Universe& rUni, universe::Satellite relativeSat,
-        universe::Satellite tgtSat, Matrix4 transform)
-{
-    auto &satPosTraj = rUni.get_reg().get<universe::UCompTransformTraj>(tgtSat);
-
-    auto const &areaPosTraj = rUni.get_reg()
-            .get<universe::UCompTransformTraj>(relativeSat);
-
-    // 1024 units = 1 meter
-    Vector3g const posAreaRelative(transform.translation()
-                                   * universe::gc_units_per_meter);
-
-//    satPosTraj.m_position = areaPosTraj.m_position + posAreaRelative;
-//    satPosTraj.m_rotation = Quaternion::
-//            fromMatrix(transform.rotationScaling());
-//    satPosTraj.m_dirty = true;
+    rAreaSat.m_requestMove.push_back(translate);
 }
 
 void SysAreaAssociate::floating_origin_translate(
-        ActiveScene& rScene,Vector3 translation)
+        ActiveScene& rScene, Vector3 translation)
 {
     auto &rReg = rScene.get_registry();
 
--- a/src/osp/Active/SysAreaAssociate.cpp
+++ b/src/osp/Active/SysAreaAssociate.cpp
@@ -57,7 +57,7 @@ void SysAreaAssociate::update_consume(ActiveScene& rScene)
 
     Vector3g deltaTotal{};
 
-    for (Vector3g const delta : std::exchange(rAreaUComp.m_moved, {}))
+    for (Vector3g const& delta : std::exchange(rAreaUComp.m_moved, {}))
     {
         deltaTotal += delta;
     }
@@ -93,7 +93,7 @@ void SysAreaAssociate::connect(ActiveScene& rScene, universe::Universe &rUni,
 
 }
 
-void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g translate)
+void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g const& translate)
 {
     ACompAreaLink *pArea = try_get_area_link(rScene);
 
--- a/src/osp/Active/SysAreaAssociate.h
+++ b/src/osp/Active/SysAreaAssociate.h
@@ -120,7 +120,7 @@ public:
      * @param rScene    [ref] Scene assicated with ActiveArea
      * @param translate [in] Amount to translate the area by
      */
-    static void area_move(ActiveScene& rScene, universe::Vector3g translate);
+    static void area_move(ActiveScene& rScene, universe::Vector3g const& translate);
 
 private:
 
--- a/src/osp/Active/SysAreaAssociate.h
+++ b/src/osp/Active/SysAreaAssociate.h
@@ -52,6 +52,11 @@ struct ACompAreaLink
 
     std::reference_wrapper<universe::Universe> m_rUniverse;
 
+    std::vector<universe::Satellite> m_enter;
+    std::vector<universe::Satellite> m_leave;
+
+    Vector3 m_move;
+
 };
 
 struct ACompActivatedSat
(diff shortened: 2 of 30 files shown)


src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -39,60 +39,39 @@ using osp::universe::UCompActiveArea;
...
-    // will be activated.
+    pLink->m_enter = std::exchange(rAreaUComp.m_enter, {});
+    pLink->m_leave = std::exchange(rAreaUComp.m_leave, {});
 
-    auto viewActRadius = rUni.get_reg().view<
-            UCompActivationRadius, UCompActivatable>();
+    Vector3g deltaTotal{};
 
-    for (Satellite sat : viewActRadius)
+    for (Vector3g const delta : std::exchange(rAreaUComp.m_moved, {}))
```
jonesmz, line 60: Vector3g has s no move semantics right? If so, const& would be slightly better on low optimization levels.

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -114,13 +93,6 @@ void SysAreaAssociate::connect(ActiveScene& rScene, universe::Universe &rUni,
...
 
-void SysAreaAssociate::disconnect(ActiveScene& rScene)
-{
-    // TODO: Have a vector of function pointers for each satellite type, that
-    //       are called to de-activate them right away.
-    //       Also delete the ACompAreaLink from the scene
-}
-
 void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g translate)
 {
```
jonesmz, line 97: Const on parameter

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -130,37 +102,14 @@ void SysAreaAssociate::area_move(ActiveScene& rScene, Vector3g translate)
...
-//    satPosTraj.m_position = areaPosTraj.m_position + posAreaRelative;
-//    satPosTraj.m_rotation = Quaternion::
-//            fromMatrix(transform.rotationScaling());
-//    satPosTraj.m_dirty = true;
+    rAreaSat.m_requestMove.push_back(translate);
 }
 
 void SysAreaAssociate::floating_origin_translate(
-        ActiveScene& rScene,Vector3 translation)
+        ActiveScene& rScene, Vector3 translation)
```
jonesmz, line 112: Const on parameter

src/osp/Active/SysAreaAssociate.h
```diff
@@ -36,9 +36,7 @@ namespace osp::active
 
 
 struct ACompAreaLink
-{
-    using MapSatToEnt_t = std::unordered_map<universe::Satellite, ActiveEnt>;
-    
+{
```
jonesmz, line 39: Std::move satalite on the AConpAreaLubj constructor?

jonesmz: Is a constructor actually needed for thia type? What about aggregate initialization?

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
+/**
```
jonesmz, line 1: No change needed for this PR. But wanted to ask your opinion on using SPDX copyright headers instead of full license text in very file?

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+ActiveEnt SysVehicleSync::activate(ActiveScene &rScene, Universe &rUni,
+                                   Satellite areaSat, Satellite tgtSat)
+{
+    SPDLOG_LOGGER_INFO(rScene.get_application().get_logger(),
+                      "Loading a vehicle");
+
+    auto &loadMeVehicle = rUni.get_reg().get<universe::UCompVehicle>(tgtSat);
+    auto &tgtPosTraj = rUni.get_reg().get<universe::UCompTransformTraj>(tgtSat);
+
+    // make sure there is vehicle data to load
```
jonesmz, line 60: Thoughts on prefixing std::optional with m_oBlueprint, similar to m_rBlueprint for references or m_pBlueprint?

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    rScene.reg_emplace<ACompActivatedSat>(vehicleEnt, tgtSat);
+
+    ACompVehicle& vehicleComp = rScene.reg_emplace<ACompVehicle>(vehicleEnt);
+    rScene.reg_emplace<ACompDrawTransform>(vehicleEnt);
+    rScene.reg_emplace<ACompVehicleInConstruction>(
+                    vehicleEnt, loadMeVehicle.m_blueprint);
+
+    // Convert position of the satellite to position in scene
+    Vector3 positionInScene = rUni.sat_calc_pos_meters(areaSat, tgtSat).value();
```
jonesmz, line 82: Const&

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+    rScene.reg_emplace<ACompFloatingOrigin>(vehicleEnt);
+    //vehicleTransform.m_enableFloatingOrigin = true;
+
+    // Create the parts
+
+    // Unique part prototypes used in the vehicle
+
+    // Access with [blueprintParts.m_partIndex]
+
+    std::vector<DependRes<PrototypePart> >& partsUsed = vehicleData.m_prototypes;
```
jonesmz, line 97: Tempted to recommend using auto& instead of explicitly stating the data type.

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+    rScene.get_registry().reserve(rScene.get_registry().capacity()
+                                   + vehicleData.m_machines.size());
+
+    // Keep track of parts
+    //std::vector<ActiveEnt> newEntities;
+    vehicleComp.m_parts.reserve(blueprintParts.size());
+
+    // Loop through list of blueprint parts, and initialize each of them into
+    // the ActiveScene
+    for (BlueprintPart& partBp : blueprintParts)
```
jonesmz, line 113: Const& ?

src/osp/Active/SysVehicleSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    auto viewVehicles = rScene.get_registry().view<ACompVehicle,
+                                                   ACompActivatedSat>();
+
+    for (ActiveEnt vehicleEnt : viewVehicles)
+    {        
+        auto const &vehicleTf = viewTf.get<ACompTransform>(vehicleEnt);
+        auto const &vehicleSat = viewVehicles.get<ACompActivatedSat>(vehicleEnt);
+
+        universe::Vector3g pos{vehicleTf.m_transform.translation() * universe::gc_units_per_meter};
```
jonesmz, line 219: Const

src/osp/CoordinateSpaces/CartesianSimple.cpp
```diff
@@ -33,21 +33,47 @@ void CoordspaceCartesianSimple::update_exchange(
...
 {
-    rData.reserve(rData.size() + rSpace.m_toAdd.size());
+    rData.reserve(rData.size() + rSpace.m_toAdd.size() - rSpace.m_toRemove.size());
 
-    auto view = rUni.get_reg().view<UCompCoordspaceIndex>();
+    auto coordIndexView = rUni.get_reg().view<UCompCoordspaceIndex>();
 
+    // Sort in descending order. If any of these are the last element, then
+    // then they may get invalidated in the following swaps.
+    sort(std::begin(rSpace.m_toRemove), std::end(rSpace.m_toRemove),
```
jonesmz, line 42: Are you using ADL here?

Or some other implementation of sort?

src/osp/Satellites/SatActiveArea.cpp
```diff
@@ -25,10 +25,119 @@
...
+
+    CoordinateSpace& rDomainCoord = rUni.coordspace_get(rInCoord.m_coordSpace);
+    auto *pDomainData = entt::any_cast<CoordspaceCartesianSimple>(&rDomainCoord.m_data);
+
+    Vector3g &rPos = pDomainData->m_positions[rInCoordIndex.m_myIndex];
 
-UCompActiveArea& SatActiveArea::add_active_area(Universe &rUni, Satellite sat)
+    std::vector<Vector3g> const toMove = std::exchange(rArea.m_requestMove, {});
+
+    for (Vector3g delta: toMove)
```
jonesmz, line 48: Const&

src/osp/Satellites/SatActiveArea.cpp
```diff
@@ -25,10 +25,119 @@
...
+    {
+        rPos += delta;
+        rArea.m_moved.push_back(delta);
+    }
+}
+
+void SatActiveArea::scan_radius(Universe& rUni, Satellite areaSat)
```
jonesmz, line 55: Const

src/osp/Satellites/SatActiveArea.cpp
```diff
@@ -25,10 +25,119 @@
...
+
+    // Deal with activating satellites that have a UCompActivationRadius
+    // Satellites that have an Activation Radius have a sphere around them. If
+    // this sphere overlaps the ActiveArea's sphere 'm_areaRadius', then it
+    // will be activated.
+
+    auto viewActRadius = rUni.get_reg().view<UCompActivationRadius,
+                                             UCompActivatable>();
+
+    for (Satellite sat : viewActRadius)
```
jonesmz, line 67: Const all the things

src/osp/Universe.cpp
```diff
@@ -64,45 +63,110 @@ Vector3g Universe::sat_calc_pos(Satellite referenceFrame, Satellite target) cons
...
+    };
+
+    CoordspaceTransform fromTransform;
+    CoordspaceTransform toTransform;
+
+    CoordinateSpace const *currentFrom = &fromCoord;
+    CoordinateSpace const *currentTo = &toCoord;
+
+    // Current depth to check for a common satellite
+    int16_t testDepth = std::min({fromCoord.m_depth, toCoord.m_depth});
```
jonesmz, line 134: Don't need initializer list

src/osp/Universe.cpp
```diff
@@ -64,45 +63,110 @@ Vector3g Universe::sat_calc_pos(Satellite referenceFrame, Satellite target) cons
...
+
+    while (true)
```
jonesmz, line 136: While true always make me nervous

src/osp/Satellites/SatVehicle.h
```diff
@@ -56,6 +56,9 @@ class SatVehicle
     static UCompVehicle& add_vehicle(
         osp::universe::Universe& rUni, osp::universe::Satellite sat,
         DependRes<BlueprintVehicle> blueprint);
+
```
jonesmz, line 59: Unnecessary newlines

src/osp/coordinates.h
```diff
@@ -71,6 +71,24 @@ constexpr size_t ccomp_min_size() noexcept
     return size_t(std::max({ccomp_id<CCOMP_T> ...})) + 1;
 }
 
+template <typename ... CCOMP_T>
+struct TupleCComp
+{
```
jonesmz, line 76: Consider inheriting from std::tuple instead of aggregating.

src/osp/coordinates.h
```diff
@@ -71,6 +71,24 @@ constexpr size_t ccomp_min_size() noexcept
...
+    constexpr TupleCComp(ViewCComp_t<CCOMP_T> ... ccomps) noexcept
+     : m_data{std::forward< ViewCComp_t<CCOMP_T> >(ccomps)...}
```
jonesmz, line 78: Inheriting from tuple allows you to replace this with using std::tuple::tuple

src/osp/coordinates.h
```diff
@@ -136,21 +172,170 @@ struct CoordinateSpace
...
 
+    /**
+     * @brief Access multiple components using a TupleCComp
+     *
+     * TupleCComp can be used to conveniently pass CComps as arguments
+     *
+     * @tparam COMP_T... Coordinate components to view
+     */
+    template<typename ... CCOMP_T>
+    constexpr std::optional< TupleCComp<CCOMP_T ...> const>ccomp_view_tuple() const noexcept
```
jonesmz, line 183: I'm suspicious of an optional that holds a const type. This seems weird

src/osp/coordinates.h
```diff
@@ -136,21 +172,170 @@ struct CoordinateSpace
...
+    int16_t m_depth{0};
+    int16_t m_pow2scale{10};
+
+}; // struct CoordinateSpace
+
+
+template <typename T>
+constexpr T mulpow2(T in, int16_t pow2) noexcept
+{
+    return (pow2 > 0) ? (in * (1L << pow2)) : (in / (1L << (-pow2)));
```
jonesmz, line 223: Write comment about how the math here works. And break down the three cases. Less than 0. 0, and greater than 0

src/osp/Active/SysAreaAssociate.h
Capital-Asterisk: Satellite is just an integer, and doesn't need move. `m_enter`, `m_leave`, and `m_move` are always going to be default initialized; the constructor is just for convenience.

src/osp/Active/SysVehicleSync.cpp
Capital-Asterisk: SPDX? sounds like some kind of pushbutton switch. I'd have to read up.

Replacing the header entirely doesn't appear to be the right way to go:

> When a license defines a recommended notice to attach to files under that license (sometimes called a “standard header”), the SPDX project recommends that the standard header be included in the files, in addition to an SPDX ID.
> 
> Additionally, when a file already contains a standard header or other license notice, the SPDX project recommends that those existing notices should not be removed. The SPDX ID is recommended to be used to supplement, not replace, existing notices in files.

https://spdx.dev/ids/#how

src/osp/Active/SysVehicleSync.cpp
Capital-Asterisk: I'm cool with that; guessing every kind of single-element 'container' should do this? It doesn't seem that useful for local variables however, since an explicit std::optional<...> type declaration would already be present and easy to spot.

This case with m_blueprint is actually a DependRes and not an optional though.

src/osp/CoordinateSpaces/CartesianSimple.cpp
Capital-Asterisk: simply forgot an `std::`

src/osp/coordinates.h
```diff
@@ -72,21 +73,18 @@ constexpr size_t ccomp_min_size() noexcept
...
 
 template <typename ... CCOMP_T>
-struct TupleCComp
+struct TupleCComp : std::tuple<ViewCComp_t<CCOMP_T> ...>
 {
-    constexpr TupleCComp(ViewCComp_t<CCOMP_T> ... ccomps) noexcept
-     : m_data{std::forward< ViewCComp_t<CCOMP_T> >(ccomps)...}
-    { }
+    using base_t = std::tuple<ViewCComp_t<CCOMP_T> ...>;
+    using base_t::tuple;
```
jonesmz, line 79: You can do "using base_t::base_t" but it doesn't really matter.

jonesmz (approved):

z-adams (approved):
A cursory skim of the code didn't reveal anything glaring, although I've never been super familiar with the ActiveArea stuff so I can't really make any code-specific architectural comments.

However, on the topic of

> Imagine there's a solar system in an N-body simulation being explored by an ActiveArea.

> The ActiveArea is moving itself; it has no mass, not affected by gravity, and has nothing to do with the N-body simulation whatsoever. This means its position data should NOT be in the N-body's coordinate space. Instead, the ActiveArea can live in its own coordinate space layered over top of the N-body coordinate space. This is implemented as the "Domain CoordinateSpace."

I wonder if the ActiveArea *shouldn't* be subject to gravity in some way. Since the acceleration of an object or frame of reference is independent of its mass, an ActiveArea centered around a vessel travelling on an inertial trajectory would follow the exact same trajectory as the vessel, which would be quite convenient. I suppose you might mean that this behavior is handled implicitly, though? When I hear "ActiveArea [...] is not affected by gravity" I imagine the ActiveArea constantly following a tangent and needing to be corrected, and that correction is what I would call "gravity" :) Let me know if you've already thought through this though (lmao english is hilarious).

Capital-Asterisk:
> I wonder if the ActiveArea _shouldn't_ be subject to gravity in some way...

I'm planning on making the ActiveArea only be able to handle constant velocities. The scene origin would be moving around as the area follows it, but this isn't a problem. Over the span of seconds and minutes, objects in space would have appeared to travel in straight lines, with little deviation. 

Needing to account for accelerations of the entire reference frame sounds like a preventable headache. Rotating planets is (hopefully) the only accelerating reference frame we'd have to deal with.

jonesmz:
Merging.

It's been approved by two reviewers, and no new commits for a month.

Issue #127, opened 2021-07-22
Change the starting scene so that the rockets that are spawned are all in a somewhat stable orbit


jonesmz:
It's OK if they eventually fall, as long as they can make 5-10 passes around the planet.

Capital-Asterisk:
This sets a few goals. Right now, there's no "set velocity" for the physics engine; this needs to be implemented. Storing velocity data in the Universe is [already possible](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/fbf0a19c2a01fbc0c302a9031fce01ec6b16fc80/src/osp/CoordinateSpaces/CartesianSimple.h#L80) but all are to 0 and are unused.

Capital-Asterisk:
This test scene doesn't exist anymore

PR #132, opened 2021-08-06
Improve Physics interface, Newton, and Vehicle separation

Capital-Asterisk:
Physics is quite lacking in functionality. For example, there's no way to set velocity (as shown in #127), not to mention how we'd even handle detecting collision events. This PR is a step towards a more solid physics interface.

[This blog post](https://www.gamasutra.com/blogs/NiklasGray/20190611/344437/Syncing_a_dataoriented_ECS_with_a_stateful_external_system.php) talks about the same problem. I've leaned towards the 'list of changes' as the most practical approach for our case.

## Physics Interface

See `src/osp/Active/physics.h`. Note that none of these are specifically related to Newton. I've also been thinking of adding PhysX support some time later.

### ACtxPhysics

Additional communication with the physics system is now handled through ACtxPhysics, the 'context' struct, which is stored in the registry. It is obtainable through `registry.ctx<ACtxPhysics>()`. This EnTT feature of contexts kind of went unnoticed for most of the project's duration.

### Rigid body component?

The rigid body component (ACompRigidBody) is now split into more components:
* ACompPhysBody
* ACompPhysDynamic
* ACompPhysLinearVel
* ACompPhysAngularVel

An entity with only ACompPhysBody and without ACompPhysDynamic will be static, and does not need mass, motion, or inertia data that dynamic bodies need.

Forces and torques can be applied by adding ACompPhysNetForce and ACompPhysNetTorque components.

Dirty flags for updating colliders and inertia are replaced with queues (vectors) stored in ACtxPhysics.

### Setting velocity

Setting velocity is handled through queues alone, without directly modifying ACompPhysLinearVel. There can be cases where multiple systems would want to read the current velocity. If ACompPhysLinearVel was intentionally writable, then that would introduce unnecessary dependency.

Dirty flags are okay for colliders and inertia, because they don't change often and imply a change in other components. Colliders and any other way to describe physics objects ends up being an INPUT read by the physics system. Velocity is an OUTPUT written by the physics engine. Changing it should be handled on some separate channel.

### Origin translation

This is previously handled by SysAreaAssociate alone, which simply translates every entity. The transformation of entities with physics should be fully controlled by the physics system, where the need to modify transformation should be a rare case. Origin translation is a bit of a special case, and there are potential optimizations for moving multiple objects vs individually moving each one (I'm mostly concerned about some broadphase structure getting messed up on the inside). PhysX has origin shifting as a built-in function.

Origin translations for physics can be handled by adding to m_originTranslate in ACtxPhysics.

## Newton improvements

Improved the general 'sturdiness' of the code, and things to comply with the new physics interface changes. There's less reliance on brute force, and better use of callbacks that in theory should improve performance.

## Vehicle separation

Most of the vehicle separation code is still sort of temporary, as a separate 'vehicle structure' system should be the way to go in the future. Aside from changes to how rigid bodies are created, the new set velocity features were used to properly explode the vehicle (not yet for angular velocity however). Separated sections of the vehicle now have the same velocity as the original. 

There was a bug where separated parts are placed without accounting for the vehicle's rotation. This was fixed by removing some outdated center-of-mass calculations. It was simply conflicting with z-adams' inertia calculations.

### DIFF
--- a/src/newtondynamics_physics/SysNewton.cpp
+++ b/src/newtondynamics_physics/SysNewton.cpp
@@ -24,6 +24,7 @@
  */
 
 #include "SysNewton.h"               // IWYU pragma: associated
+#include "ospnewton.h"
 
 #include <osp/Active/scene.h>        // for ACompHierarchy
 #include <osp/Active/physics.h>      // for ACompRigidBody
@@ -49,9 +50,7 @@
 
 #include <new>
 
-using ospnewton::SysNewton;
-using ospnewton::ACompNwtBody;
-using ospnewton::ACompNwtWorld;
+using namespace ospnewton;
 
 using osp::active::ActiveScene;
 using osp::active::ActiveReg_t;
@@ -75,30 +74,6 @@ using osp::active::ACompTransformMutable;
 
 using osp::Matrix4;
 using osp::Vector3;
-
-// TODO: not compiling, std::hardware_destructive_interference_size not found?
-template<typename T>
-struct alignas(64) CachePadded : T { };
-
-struct ACtxNwtWorld
-{
-    ACtxNwtWorld(ActiveScene &rScene, int threadCount)
-     : m_pScene(&rScene)
-     , m_viewForce(rScene.get_registry().view<ACompPhysNetForce>())
-     , m_viewTorque(rScene.get_registry().view<ACompPhysNetTorque>())
-     , m_setTf(threadCount)
-    { }
-
-    ActiveScene *m_pScene;
-
-
-    entt::basic_view<ActiveEnt, entt::exclude_t<>, ACompPhysNetForce> m_viewForce;
-    entt::basic_view<ActiveEnt, entt::exclude_t<>, ACompPhysNetTorque> m_viewTorque;
-
-    // transformations to set recorded in cb_set_transform
-    using SetTfPair_t = std::pair<ActiveEnt, NewtonBody const* const>;
-    std::vector< CachePadded< std::vector<SetTfPair_t> > > m_setTf;
-};
 
 // Callback called for dynamic rigid bodies for applying force and torque
 void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
@@ -137,51 +112,63 @@ void cb_set_transform(NewtonBody const* const pBody,
     uintptr_t userData = reinterpret_cast<uintptr_t>(NewtonBodyGetUserData(pBody));
     ActiveEnt ent = static_cast<ActiveEnt>(userData);
 
-    pWorldCtx->m_setTf[threadIndex].emplace_back(ent, pBody);
+    pWorldCtx->m_perThread[threadIndex].m_setTf.emplace_back(ent, pBody);
 }
 
 void SysNewton::setup(ActiveScene &rScene)
 {
-    // Connect signal handlers to destruct Newton objects when components are
-    // deleted.
+    int threadCount = 1; // temporary
 
-    rScene.get_registry().on_destroy<ACompNwtBody>()
-                    .connect<&SysNewton::on_body_destruct>();
+    ActiveReg_t &rReg = rScene.get_registry();
 
-    rScene.get_registry().on_destroy<ACompNwtCollider>()
-                    .connect<&SysNewton::on_shape_destruct>();
+    rReg.set<ACtxNwtWorld>(rScene, threadCount);
+}
 
-    rScene.get_registry().on_destroy<ACompNwtWorld>()
-                    .connect<&SysNewton::on_world_destruct>();
+void SysNewton::destroy(ActiveScene &rScene)
+{
+    ActiveReg_t &rReg = rScene.get_registry();
+
+    // delete collision shapes and bodies before deleting the world
+    rReg.clear<ACompNwtCollider>();
+    rReg.clear<ACompNwtBody>();
+
+    // delete world
+    rReg.unset<ACtxNwtWorld>();
+}
+
+void SysNewton::update_translate(ActiveScene& rScene)
+{
+    ActiveReg_t &rReg = rScene.get_registry();
+
+    auto &rPhysCtx = rReg.ctx<ACtxPhysics>();
+    auto &rWorldCtx = rReg.ctx<ACtxNwtWorld>();
+
+    NewtonWorld const* pNwtWorld = rWorldCtx.m_nwtWorld.get();
+
+    // Origin translation
+    if (Vector3 const translate = std::exchange(rPhysCtx.m_originTranslate, {});
+        ! translate.isZero())
+    {
+        // Translate every newton body
+        for (NewtonBody const* pBody = NewtonWorldGetFirstBody(pNwtWorld);
+             pBody != nullptr; pBody = NewtonWorldGetNextBody(pNwtWorld, pBody))
+        {
+            Matrix4 matrix;
+            NewtonBodyGetMatrix(pBody, matrix.data());
+            matrix.translation() += translate;
+            NewtonBodySetMatrix(pBody, matrix.data());
+        }
+    }
 }
 
 void SysNewton::update_world(ActiveScene& rScene)
 {
     ActiveReg_t &rReg = rScene.get_registry();
 
-    // Try getting the newton world from the scene root.
-    // It is very unlikely that multiple physics worlds are ever needed
-    auto *pWorldComp = rReg.try_get<ACompNwtWorld>(rScene.hier_get_root());
-
-    if (pWorldComp == nullptr)
-    {
-        return; // No physics world component
-    }
-
-    if (pWorldComp->m_nwtWorld == nullptr)
-    {
-        // Create world if not yet initialized
-        pWorldComp->m_nwtWorld = NewtonCreate();
-        auto& rWorldCtx = rReg.set<ACtxNwtWorld>(
-                rScene, NewtonGetThreadsCount(pWorldComp->m_nwtWorld));
-
-        NewtonWorldSetUserData(pWorldComp->m_nwtWorld, &rWorldCtx);
-    }
-
     auto &rPhysCtx = rReg.ctx<ACtxPhysics>();
     auto &rWorldCtx = rReg.ctx<ACtxNwtWorld>();
 
-    NewtonWorld const* pNwtWorld = pWorldComp->m_nwtWorld;
+    NewtonWorld const* pNwtWorld = rWorldCtx.m_nwtWorld.get();
 
     // temporary: just delete the Newton Body if colliders change, so it can be
     //            reinitialized with new colliders
@@ -217,20 +204,6 @@ void SysNewton::update_world(ActiveScene& rScene)
                               vel.data());
     }
 
-    // Origin translation
-    if (Vector3 const translate = std::exchange(rPhysCtx.m_originTranslate, {});
-        ! translate.isZero())
-    {
-        // Translate every newton body
-        for (NewtonBody const* pBody = NewtonWorldGetFirstBody(pNwtWorld);
-             pBody != nullptr; pBody = NewtonWorldGetNextBody(pNwtWorld, pBody))
-        {
-            Matrix4 matrix;
-            NewtonBodyGetMatrix(pBody, matrix.data());
-            matrix.translation() += translate;
-            NewtonBodySetMatrix(pBody, matrix.data());
-        }
-    }
 
     // Update the world
     NewtonUpdate(pNwtWorld, rScene.get_time_delta_fixed());
@@ -243,9 +216,9 @@ void SysNewton::update_world(ActiveScene& rScene)
 
 
     // Apply transforms and also velocity
-    for (auto &vec : rWorldCtx.m_setTf)
+    for (auto &rPerThread : rWorldCtx.m_perThread)
     {
-        for (auto const& [ent, pBody] : std::exchange(vec, {}))
+        for (auto const& [ent, pBody] : std::exchange(rPerThread.m_setTf, {}))
         {
             NewtonBodyGetMatrix(
                     pBody, viewTf.get<ACompTransform>(ent).m_transform.data());
@@ -264,7 +237,7 @@ void SysNewton::update_world(ActiveScene& rScene)
 
 void SysNewton::find_colliders_recurse(
         ActiveScene& rScene, ActiveEnt ent, Matrix4 const &transform,
-        NewtonWorld const* pNwtWorld, NewtonCollision *rCompound)
+        NewtonWorld const* pNwtWorld, NewtonCollision *pCompound)
 {
     ActiveReg_t &rReg = rScene.get_registry();
     ActiveEnt nextChild = ent;
@@ -272,8 +245,7 @@ void SysNewton::find_colliders_recurse(
     // iterate through siblings of ent
     while(nextChild != entt::null)
     {
-        auto const &childHeir = rScene.reg_get<ACompHierarchy>(nextChild);
-
+        auto const &rChildHeir = rScene.reg_get<ACompHierarchy>(nextChild);
 
         if (auto const *pChildTransform
                     = rScene.get_registry().try_get<ACompTransform>(nextChild);
@@ -292,7 +264,6 @@ void SysNewton::find_colliders_recurse(
                     // No Newton collider exists yet
                     collision = NewtonCreateSphere(pNwtWorld, 0.5f, 0, nullptr);
                     rReg.emplace<ACompNwtCollider>(nextChild, collision);
-                    ACompNwtCollider f(collision);
                 }
                 else
                 {
@@ -305,14 +276,14 @@ void SysNewton::find_colliders_recurse(
                 NewtonCollisionSetMatrix(collision, f.data());
 
                 // Add body to compound collision
-                NewtonCompoundCollisionAddSubCollision(rCompound, collision);
+                NewtonCompoundCollisionAddSubCollision(pCompound, collision);
             }
 
-            find_colliders_recurse(rScene, childHeir.m_childFirst, childMatrix,
-                                   pNwtWorld, rCompound);
+            find_colliders_recurse(rScene, rChildHeir.m_childFirst, childMatrix,
+                                   pNwtWorld, pCompound);
         }
 
-        nextChild = childHeir.m_siblingNext;
+        nextChild = rChildHeir.m_siblingNext;
     }
 }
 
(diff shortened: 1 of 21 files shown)


z-adams (approved):
I don't have the experience to comment on the overall design but it seems reasonable and the code looks good to me. Merge when you see fit

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -51,139 +56,177 @@ using osp::active::ACompTransformMutable;
...
 
-// Callback called for every Rigid Body (even static ones) on NewtonUpdate
+// Callback called for dynamic rigid bodies for applying force and torque
 void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
 {
-    // Get Scene from Newton World
-    ActiveScene* scene = static_cast<ActiveScene*>(
-                NewtonWorldGetUserData(NewtonBodyGetWorld(pBody)));
+    // Get context from Newton World
+    ACtxNwtWorld* pWorldCtx = static_cast<ACtxNwtWorld*>(
```
jonesmz, line 63: could do "auto * const" without losing clarity here, if you wanted to.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -51,139 +56,177 @@ using osp::active::ACompTransformMutable;
...
+            NewtonWorldGetUserData(NewtonBodyGetWorld(pBody)));
 
     // Get Associated entity
     uintptr_t userData = reinterpret_cast<uintptr_t>(NewtonBodyGetUserData(pBody));
     ActiveEnt ent = static_cast<ActiveEnt>(userData);
```
jonesmz, line 68: weak recommendation:
Variables should be const whenever practical to do so.

src/newtondynamics_physics/SysNewton.cpp
jonesmz: it also looks like you only use this in a read-only context, so "auto const*const"

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -51,139 +56,177 @@ using osp::active::ACompTransformMutable;
...
-        pWorldComp->m_nwtWorld = NewtonCreate();
-        NewtonWorldSetUserData(pWorldComp->m_nwtWorld, &rScene);
+        // Translate every newton body
+        for (NewtonBody const* pBody = NewtonWorldGetFirstBody(pNwtWorld);
+             pBody != nullptr; pBody = NewtonWorldGetNextBody(pNwtWorld, pBody))
+        {
+            Matrix4 matrix;
+            NewtonBodyGetMatrix(pBody, matrix.data());
+            matrix.translation() += translate;
+            NewtonBodySetMatrix(pBody, matrix.data());
```
jonesmz, line 140: Odd that there's no built in matrix translate function. Oh well.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -308,9 +349,10 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt ent,
...
             pBody = nullptr;
         }
+        break;
     }
     default:
-        break;
+        return;
     }
 
     // by now, the Newton rigid body components now exist
```
jonesmz, line 358: by now, the Newton rigid body components exist

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -328,8 +370,9 @@ void SysNewton::create_body(ActiveScene& rScene, ActiveEnt ent,
...
     NewtonBodySetAngularDamping(pBody, Vector3(1.0f, 1.0f, 1.0f).data());
 
-    // Set callback for updating position of entity and everything else
-    NewtonBodySetForceAndTorqueCallback(pBody, cb_force_torque);
+    // Set callback for applying force and setting transforms
+    NewtonBodySetForceAndTorqueCallback(pBody, &cb_force_torque);
+    NewtonBodySetTransformCallback(pBody, &cb_set_transform);
 
     // Set user data to entity
     // WARNING: this will cause issues if 64-bit entities are used on a
```
jonesmz, line 378: in my opinion, we should explicitly trigger compiler errors if compiling for 32bit.

I don't think there is any reason to be supporting a 32bit platform in a post 2020 world. 64bit has been ubiquitous for basically a decade now.

jonesmz: Basically something like

#if 32bit
#error "Get a better computer"
#endif

src/newtondynamics_physics/ospnewton.h
```diff
@@ -0,0 +1,124 @@
...
+namespace ospnewton
+{
+
+/**
+ * @brief Associates an entity with a NewtonBody and manages its lifetime
+ */
+struct ACompNwtBody
+{
+
+    ACompNwtBody(NewtonBody const* pBody) noexcept
```
jonesmz, line 41: for such a simple class, is it necessary to use a private member variable?

you could use aggregate initialization, and completely skip the need for a constructor, and get auto-generated move constructors at the same time.

src/newtondynamics_physics/ospnewton.h
```diff
@@ -0,0 +1,124 @@
...
+     : m_pBody(pBody) { }
+
+    NewtonBody const* body() const noexcept { return m_pBody.get(); }
+
+private:
+    struct Deleter
+    {
+        void operator() (NewtonBody const* pCollision)
```
jonesmz, line 49: const

src/newtondynamics_physics/ospnewton.h
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * @brief Represents an instance of a Newton physics world in the scane
+ */
+struct ACtxNwtWorld
+{
+
+    // TODO: not compiling, std::hardware_destructive_interference_size not found?
+    struct alignas(64) PerThread
+    {
+        using SetTfPair_t = std::pair<osp::active::ActiveEnt,
```
jonesmz, line 87: instead of std::pair, consider just a trivial POD struct, which involves way way less compiler timer than std::pair.

If you still need structured bindings, use https://devblogs.microsoft.com/oldnewthing/20201015-00/?p=104369


e.g.

```cpp
struct Person
{
  std::string name;
  int age;

  template<std::size_t Index>
  auto&& get()       &  { return get_helper<Index>(*this); }

  template<std::size_t Index>
  auto&& get()       && { return get_helper<Index>(*this); }

  template<std::size_t Index>
  auto&& get() const &  { return get_helper<Index>(*this); }

  template<std::size_t Index>
  auto&& get() const && { return get_helper<Index>(*this); }

private:
  template<std::size_t Index, typename T>
  auto&& get_helper(T&& t)
  {
    static_assert(Index < 2,
      "Index out of bounds for Custom::Person");
    if constexpr (Index == 0) return std::forward<T>(t).name;
    if constexpr (Index == 1) return std::forward<T>(t).age;
  }
};
```

src/newtondynamics_physics/ospnewton.h
```diff
@@ -0,0 +1,124 @@
...
+};
+
+/**
+ * @brief Represents an instance of a Newton physics world in the scane
+ */
+struct ACtxNwtWorld
+{
+
+    // TODO: not compiling, std::hardware_destructive_interference_size not found?
+    struct alignas(64) PerThread
```
jonesmz, line 85: You said in discord that this alignas(64) was supposed to represent the size of the CPUs cache line.

Recommend documenting that in the code.

src/osp/Active/SysAreaAssociate.cpp
```diff
@@ -184,6 +186,9 @@ void SysAreaAssociate::floating_origin_translate(
 
         entTransform.m_transform.translation() += translation;
     }
+
+    // Tell physics engine to translate too
```
jonesmz, line 190: too? also?

src/osp/Active/SysForceFields.cpp
```diff
@@ -51,17 +51,19 @@ void SysFFGravity::update_force(ActiveScene& rScene)
...
             auto &massTransform = viewMasses.get<ACompTransform>(massEnt);
 
             Vector3 relativePos = fieldTransform.m_transform.translation()
                                 - massTransform.m_transform.translation();
             float r = relativePos.length();
-            Vector3 accel = relativePos * fieldFFGrav.m_Gmass / (r * r * r);
-            //               massBody.m_bodyData.m_mass / r;
-            // gm1/r = a
+            Vector3 force = relativePos * fieldFFGrav.m_Gmass
+                          * massBody.m_totalMass / (r * r * r);
```
jonesmz, line 61: parenthesis to make it very clear what the intended order of operations is.

src/osp/Active/SysVehicle.cpp
```diff
@@ -107,24 +103,25 @@ ActiveEnt SysVehicle::activate(ActiveScene &rScene, universe::Universe &rUni,
     rScene.get_registry().reserve(rScene.get_registry().capacity()
                                    + vehicleData.m_machines.size());
 
-    // Keep track of parts
+    // Keep track of parts                rScene.reg_emplace<ACompFloatingOrigin>(islandEnt);
```
jonesmz, line 106: rScene.reg_emplace<ACompFloatingOrigin>(isLandEnt);
is commented out.

If that was intentional, please change the comment to make it more obvious that it was intentional.

src/osp/Active/SysVehicle.cpp
```diff
@@ -439,26 +432,33 @@ void SysVehicle::update_vehicle_modification(ActiveScene& rScene)
 
             islands[0] = vehicleEnt;
 
+            Vector3 const velocity = rReg.get<ACompPhysLinearVel>(vehicleEnt);
+
             // NOTE: vehicleVehicle and vehicleTransform become invalid
             //       when emplacing new ones.
             for (size_t i = 1; i < islands.size(); i ++)
```
jonesmz, line 439: change this to

for(auto & currIsland : islands)
{

....

currIsland = islandEnt;
}

src/osp/Active/SysVehicle.cpp
```diff
@@ -500,65 +499,21 @@ void SysVehicle::update_vehicle_modification(ActiveScene& rScene)
...
             for (ActiveEnt islandEnt : islands)
             {
-                auto &islandVehicle = view.get<ACompVehicle>(islandEnt);
-                auto &islandTransform
-                        = rScene.reg_get<ACompTransform>(islandEnt);
-
-                Vector3 comOffset;
-                //float totalMass;
+                auto &rIslandVehicle = view.get<ACompVehicle>(islandEnt);
+                auto &rIslandTf = rScene.reg_get<ACompTransform>(islandEnt);
```
jonesmz, line 513: This whole forloop is just two getters now?

that doesn't seem like what you intended?

src/osp/Active/physics.h
```diff
@@ -31,26 +31,57 @@ namespace osp::active
...
+ * @brief The ACtxPhysics struct
  */
-struct ACompRigidBody
+struct ACtxPhysics
 {
-    // Modify these
-    Vector3 m_inertia{1, 1, 1};
-    Vector3 m_netForce{0, 0, 0};
-    Vector3 m_netTorque{0, 0, 0};
+    Vector3 m_originTranslate;
```
jonesmz, line 38: no default value?

src/osp/Active/physics.h
```diff
@@ -31,26 +31,57 @@ namespace osp::active
...
-    bool m_colliderDirty{false}; // set true if collider is modified
-    bool m_inertiaDirty{false}; // set true if rigidbody is modified
+/**
+ * @brief Adds rigid body dynamics to entities with ACompPhysBody
+ */
+struct ACompPhysDynamic
+{
+    Vector3 m_centerOfMassOffset{0, 0, 0};
+    Vector3 m_inertia{1, 1, 1};
+    float m_totalMass;
```
jonesmz, line 59: default value?

src/osp/Resource/AssetImporter.cpp
```diff
@@ -469,8 +469,8 @@ void AssetImporter::proto_add_obj_recurse(
         // TODO: Add support different collider shapes here!
         std::string const& shapeName = extras.Get("shape").Get<std::string>();
         // change this to some map too
-        const phys::ECollisionShape shape = (shapeName == "cylinder")
-            ? phys::ECollisionShape::CYLINDER : phys::ECollisionShape::BOX;
+        const phys::EShape shape = (shapeName == "cylinder")
+            ? phys::EShape::Cylinder : phys::EShape::Box;
```
jonesmz, line 473: personally i always put the ? and : on different lines, but it's not a big deal.

src/planet-a/Active/temporarynewtonmesh.h
```diff
@@ -0,0 +1,68 @@
...
+template<class TRIANGLE_IT_T>
+void debug_create_tri_mesh_static(
+        osp::active::ActiveScene& rScene, osp::active::ACompShape& rShape,
+        osp::active::ActiveEnt chunkEnt,
+        TRIANGLE_IT_T const& start, TRIANGLE_IT_T const& end)
+{
+    // TODO: this is actually horrendously slow and WILL cause issues later on.
+    //       Tree collisions aren't made for real-time loading. Consider
+    //       manually hacking up serialized data instead of add face, or use
+    //       Newton's dgAABBPolygonSoup stuff directly
```
jonesmz, line 39: good note

src/planet-a/Active/temporarynewtonmesh.h
```diff
@@ -0,0 +1,68 @@
...
+
+    auto &rWorldCtx = rReg.ctx<ospnewton::ACtxNwtWorld>();
+    NewtonWorld const* pNwtWorld = rWorldCtx.m_nwtWorld.get();
+    NewtonCollision const* pTree = NewtonCreateTreeCollision(pNwtWorld, 0);
+
+    NewtonTreeCollisionBeginBuild(pTree);
+
+    Vector3 triangle[3];
+
+    for (TRIANGLE_IT_T it = start; it != end; it += 3)
```
jonesmz, line 53: Oh boy. Not a fan of iteration requiring +3.

I mean, you clearly have it working the way you want, so it's fine. But this kinda thing always scares me off a bit.

jonesmz (approved):
Approved, as I didn't find anything super terrible, but there are a few items that I pointed out that I think need to be looked at just to make sure.

src/newtondynamics_physics/ospnewton.h
z-adams: What exactly is the behavior of `alignas()`? Do all instances of this struct get a bunch of extra padding and are shifted within their memory footprint to align them? I'm only familiar with struct packing and allocating aligned dynamic memory, I've never heard of this macro before

src/osp/Active/SysVehicle.cpp
Capital-Asterisk: Yep, this is excess code from removed center of mass calculations.

src/osp/Active/physics.h
Capital-Asterisk: Vector3 default constructor initializes to (0, 0, 0)

src/newtondynamics_physics/ospnewton.h
Capital-Asterisk: These structs are stored contiguously in `std::vector<PerThread> m_perThread;`, and are accessed among different threads simultaneously. Without proper alignment, they're small enough to fall into the same cache line. This can lead to serious performance degradation from False Sharing, as the cpu cores will end up invalidating each other's caches when they write a new value.

`alignas` specifies alignment requirements, though I'm not entirely familiar with what this really affects. If I'm doing this correctly, then the struct's size will be a multiple of 64, and the standard C++17 allocators will do aligned allocations.

src/newtondynamics_physics/ospnewton.h
Capital-Asterisk: Hmm, I guess so. Raw pointers were used before, and accidentally changing it is no longer an issue with unique_ptr. How about just making this entire struct a `using ACompNwtBody_t = std::unique_ptr<NewtonBody const, Deleter> m_pBody;`?

src/newtondynamics_physics/ospnewton.h
z-adams: Ok, interesting. I wouldn't have expected this to be supported by MSVC since they don't support `std::aligned_alloc` yet but perhaps since they get control over both ends it works out. The build seems to pass, anyway.

src/newtondynamics_physics/ospnewton.h
jonesmz: Id' be perfectly happy with that.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: Just realized a 32-bit computer wouldn't be able to support over 2^32 entities in the first place

src/newtondynamics_physics/ospnewton.h
Capital-Asterisk: Doing this in many parts of the code just doesn't seem practical. How about just implementing our own lighter pair?

src/osp/Active/SysVehicle.cpp
Capital-Asterisk: Going to use `for (ActiveEnt& rIslands : Corrade::Containers::arrayView(islands).suffix(1))` instead. Note that `i` is initialized to 1, skipping the first element.

src/planet-a/Active/temporarynewtonmesh.h
Capital-Asterisk: These triangle iterators are gonna go, along with this entire file. A replacement would be some better way to represent meshes, and possibly [NewtonUserCollisions](http://www.newtondynamics.com/wiki/index.php/NewtonCreateUserMeshCollision), which still gets me scratching my head.

src/newtondynamics_physics/ospnewton.h
jonesmz: I mean, you still get 70% of the benefits i described by switching to POD structs.

The structured binding support isn't required.

jonesmz: Hell, half the reason to support structured bindings is to get better names associated with first and second. Or with tuples, to avoid using std::get<>.

With a POD type, that's not even an issue.

src/osp/Active/SysVehicle.cpp
jonesmz: :+1:

src/planet-a/Active/temporarynewtonmesh.h
jonesmz: :+1:

EhWhoAmI (approved):

Issue #107, opened 2021-04-18
Enums for stock shaders


z-adams:
Right now, shaders are described by an `ACompShader` which stores a function pointer that is used to invoke the shader on some entity. The downside to this arrangement is that it leaks graphics API dependence slightly into the rest of the codebase. For now, the default Phong shader is fallen back on when the function pointer value is `nullptr`, but a better approach described by @Capital-Asterisk would be to create an enumeration of stock shader types like
```cpp
enum class EShader
{
    Phong = 0,
    PBR = 1,
    Planet = 2,
    ChemRocketExhaust = 3,
    JetEngineExhaust = 4,
    Skybox =  5,
    // ...
};
```
and provide an interface that accepts the enum when the user attaches a shader:
```cpp
reg.emplace<ACompVisible>(ent);
reg.emplace<ACompShader>(ent, EShader::Phong);
```
which may be achieved by an `ACompShader` constructor like
```cpp
ACompShader::ACompShader(EShader shader)
{
    switch (shader)
    {
    case EShader::Phong:
        m_drawFnc = osp::shader::Phong::draw_entity;
        return;
    case EShader::PBR:
        // ...
    }
}
```
or some similar approach.

Specialized shaders whose function need not be exposed to the user (e.g. compute shaders) do not need to be included in the `EShader` enum.

Capital-Asterisk:
ACompShader is dead, replaced by RenderGroups and materials.

PR #134, opened 2021-08-16
Render Groups, Draw Assigners, and Materials

Capital-Asterisk:
Got a headache after trying to set the shader of the Planet, calling for more changes.

The main purpose of this PR is to move the responsibility of assigning shaders to the renderer.

Systems creating drawable entities will no longer need to locate shaders. This was previously problematic as which shaders to use are dependent on whichever rendering techniques (Forward, Deferred, etc...) are in use. This means creating shaders in the same systems that create drawable entities (like creating Parts and Planet terrain) couples them to a specific rendering technique.

Also, only one shader was allowed per entity, which is also a problem as multiple render passes may simultaneously need different shaders to draw the same entity.

## New shader assignment process

For setting up forward rendering, two new named "RenderGroups" are added in [SysRender.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/osp/Active/SysRender.cpp#L109-L111) as part of ACtxRenderGroups stored in the ActiveScene. This is called from [flight.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/test_application/flight.cpp#L187).

RenderGroups track a set of entities and their assigned drawing functions using entt::storage, essentially a runtime component.

"Draw Assigners" are then injected into ACtxRenderGroups in [flight.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/test_application/flight.cpp#L321-L340). These are closures(?) that are each associated with a "Material" to assign a draw function (and a shader) to entities for each RenderGroup.

Material is simply an empty struct that acts as a runtime enum using entt::family. This works as a better alternative to #107.  

Systems creating drawables will now need to assign a Material to the entities they create, by adding to a queue in ACtxRenderGroups.

[Parts example](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/osp/Active/SysVehicle.cpp#L279), [Planet terrain example](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/planet-a/Active/SysPlanetA.cpp#L221)

Entities in the queue are then passed to the DrawAssigners of every RenderGroup, where they can read the entity's components and determine which draw functions and shaders they should be assigned. See [SysRender.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/osp/Active/SysRender.cpp#L223-L246), called from [flight.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/test_application/flight.cpp#L255).

The Phong shader, for example, chooses a textured or untextured shader depending on if the entity has ACompDiffuseTex: [Phong.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/osp/Shaders/Phong.cpp#L87-L94).

Entities are then drawn in render passes by accessing the named RenderGroups, see [SysRender.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/74bc4b6d0f643de794afe5e9a86092de43b70db2/src/osp/Active/SysRender.cpp#L137-L144).


src/adera/Shaders/PlumeShader.cpp
```diff
@@ -73,8 +78,29 @@ void PlumeShader::draw_plume(ActiveEnt e,
...
+
+    // Draw front face
+    Renderer::setFaceCullingMode(Renderer::PolygonFacing::Back);
+    rShader.draw(rMesh);
+}
+
+PlumeShader::RenderGroup::DrawAssigner_t PlumeShader::gen_assign_plume(
+        PlumeShader *pShader)
+{
+    return [pShader]
```
jonesmz, line 95: my data-ownership senses are tingling here.

If you know that it's 100% safe to be passing around raw-pointers like this, please carefully document the function to indicate as such.

Otherwise, this would be a good opportunity to use a std::weak_ptr if pShader can die, or even an std::shared_ptr, if it can't.

src/adera/Shaders/PlumeShader.h
```diff
@@ -58,12 +63,13 @@ class PlumeShader : public Magnum::GL::AbstractShaderProgram
 
     PlumeShader();
 
-    static void draw_plume(osp::active::ActiveEnt e,
-        osp::active::ActiveScene& rScene,
-        osp::active::ACompCamera const& camera) noexcept;
+    static void draw_plume(
+            osp::active::ActiveEnt ent, osp::active::ActiveScene& rScene,
+            osp::active::ACompCamera const& camera, void* pUserData) noexcept;
```
jonesmz, line 68: Can't say i'm a fan of passing around void*.

Is this absolutely unavoidable?

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+        static constexpr std::string_view namePrefix = "fx_plume_";
+        auto const* name = rScene.reg_try_get<ACompName>(ent);
 
-    auto const& plume = rScene.reg_get<ACompExhaustPlume>(node);
+        if (nullptr == name)
+        {
+            return EHierarchyTraverseStatus::Continue;
+        }
+
+        if (0 == name->m_name.compare(0, namePrefix.size(), namePrefix))
```
jonesmz, line 77: https://en.cppreference.com/w/cpp/string/basic_string/starts_with
https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with ?

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+    if (plumeNode == entt::null)
     {
-        return;  // plume already initialized
+        SPDLOG_LOGGER_ERROR(rScene.get_application().get_logger(),
+                          "ERROR: could not find plume anchor for MachineRocket {}", part);
+        return;
     }
 
-    DependRes<PlumeEffectData> plumeEffect = plume.m_effect;
+    SPDLOG_LOGGER_INFO(rScene.get_application().get_logger(), "MachineRocket {}\'s associated plume: {}",
```
jonesmz, line 94: the ' needs escaping? That's really odd...?

jonesmz: Unrelated to the ', i feel like we eventually want to have per-component logging, not application wide logging. But that doesn't need anything to happen here.,

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+        part, plumeNode);
+
+    // Get plume effect
+
+    using osp::DependRes;
+    using osp::Package;
 
     Package& pkg = rScene.get_application().debug_find_package("lzdb");
+    std::string_view plumeAnchorName = rScene.reg_get<ACompName>(plumeNode).m_name;
+    std::string_view effectName = plumeAnchorName.substr(3, plumeAnchorName.length() - 3);
```
jonesmz, line 104: plumeAnchorName.remove_prefix(3);

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+    SPDLOG_LOGGER_INFO(rScene.get_application().get_logger(), "MachineRocket {}\'s associated plume: {}",
+        part, plumeNode);
+
+    // Get plume effect
+
+    using osp::DependRes;
+    using osp::Package;
 
     Package& pkg = rScene.get_application().debug_find_package("lzdb");
+    std::string_view plumeAnchorName = rScene.reg_get<ACompName>(plumeNode).m_name;
```
jonesmz, line 103: const

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+    // TODO: this is kind of a hacky function, Plumes should be made into
+    //       their own Machines
+
+    auto view = rScene.get_registry()
+            .view<osp::active::ACompVehicle,
+                  osp::active::ACompVehicleInConstruction>();
+
+    osp::machine_id_t const id = osp::mach_id<machines::MachineRocket>();
+
+    for (auto [vehEnt, rVeh, rVehConstr] : view.each())
```
jonesmz, line 150: auto&

jonesmz: or const&

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,135 @@
...
+
+        // Get all MachineRockets in the vehicle
+        for (osp::BlueprintMachine const &mach : vehBp.m_machines[id])
+        {
+            // Get part
+            ActiveEnt partEnt = rVeh.m_parts[mach.m_partIndex];
+
+            // Get machine entity previously reserved by SysVehicle
+            auto const& machines = rScene.reg_get<osp::active::ACompMachines>(partEnt);
+            ActiveEnt machEnt = machines.m_machines[mach.m_protoMachineIndex];
```
jonesmz, line 168: const

src/osp/Active/SysRender.cpp
```diff
@@ -77,57 +70,78 @@ void SysRender::setup(ActiveScene& rScene)
...
+void SysRender::setup_forward_renderer(ActiveScene &rScene)
+{
+    Package &rCtxResources = rScene.get_context_resources();
+
+    Vector2i viewSize = Magnum::GL::defaultFramebuffer.viewport().size();
 
-    // Create the default render pipeline
-    resources.add<RenderPipeline>("default", create_forward_renderer());
+    rScene.reg_emplace<ACompRenderTarget>(
+            rScene.hier_get_root(), viewSize,
```
jonesmz, line 102: minor:
prefer arguments all on same line, or all on new line, not mixed.

src/osp/Active/SysRender.cpp
```diff
@@ -77,57 +70,78 @@ void SysRender::setup(ActiveScene& rScene)
...
-
-            // Fetch opaque objects
-            auto opaqueView = reg.view<ACompOpaque, ACompVisible, ACompShader>();
-
-            SysRender::draw_group(rScene, opaqueView, camera);
+            Renderer::setDepthMask(true);
+
+            ActiveReg_t &rReg = rScene.get_registry();
+            auto const viewVisible = rReg.view<const ACompVisible>();
+            auto &rGroups = rReg.ctx<ACtxRenderGroups>();
```
jonesmz, line 135: const& ?

src/osp/Active/SysRender.cpp
```diff
@@ -77,57 +70,78 @@ void SysRender::setup(ActiveScene& rScene)
...
+
+            for (auto const& [ent, toDraw]
+                 : rGroups.m_groups["fwd_opaque"].view().each())
```
jonesmz, line 138: is your intention to create-if-not-exists ?

Imho, that should be happening before you get to the render pipeline.

Since the render pipeline is going to execute the same instructions over and over, no sense in paying the cost of checking if the thing is already allocated and allocating if not, over and over.

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+     *
+     * @param ActiveEnt    [in] The entity being drawn; used to fetch component data
+     * @param ActiveScene& [ref] The scene containing the entity's component data
+     * @param ACompCamera  [in] Camera used to draw the scene
+     * @param void*        [in] User data
+     */
+    using ShaderDrawFnc_t = void (*)(
+            ActiveEnt, ActiveScene&, ACompCamera const&, void*) noexcept;
+
+    ShaderDrawFnc_t m_draw;
```
jonesmz, line 83: The whole EntityToDraw type should have an
```

void operator()(ActiveEnt a, ActiveScene& b, ACompCamera const& c)
{
    m_draw(a,b,c,m_data);
}
```

src/osp/Active/SysRender.cpp
```diff
@@ -187,10 +206,91 @@ void SysRender::display_default_rendertarget(ActiveScene& rScene)
...
+                    = rGroup.m_assigners.at(i);
+
+            if ( ! bool(assigner) )
+            {
+                continue; // Assigner is null, not set to a function
+            }
+
+            // Pass all new drawables (rGroups.m_added[i]) to the assigner.
+            // This adds accepted entities to rGroup.m_entities
+            assigner(rScene, rGroup.m_entities, rGroups.m_added.at(i));
```
jonesmz, line 244: looks like m_added and m_assigners are required to be exactly the same size, and that you always iterate them in lock step.

Recommend adding to this rGroup object an iterator type that implements the structured binding protocol. (https://devblogs.microsoft.com/oldnewthing/20201015-00/?p=104369)

Then you can iterate like
```
for(auto & [added, assigner] : rGroup.get_thing_range())
{
...
}
```


which'll make it substantially harder to screw up the lock-step access.

src/osp/Active/SysRender.cpp
```diff
@@ -187,10 +206,91 @@ void SysRender::display_default_rendertarget(ActiveScene& rScene)
...
+        toDelete.push_back(ent);
+    }
+
+    if (toDelete.empty())
+    {
+        return;
+    }
+
+    // Delete from each draw group
+    for (auto& [name, rGroup] : rGroups.m_groups)
```
jonesmz, line 280: [[maybe_unused]]

to surpress "unused variable 'name'"

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+#include <functional>
+#include <unordered_map>
+
+namespace osp::active
+{
+    struct EntityToDraw;
+}
+
+template<>
+struct entt::storage_traits<osp::active::ActiveEnt, osp::active::EntityToDraw> {
```
jonesmz, line 42: newline?

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+
+    /**
+     * @brief Assigns drawing functions and user data (like shaders) to entities
+     *        of a certain material type for each RenderGroup
+     *
+     * @param ActiveScene& [ref] Scene the entities are part of
+     * @param Storage_t&   [ref] Storage to add accepted entities and data to
+     * @param ArrayView_t  [in] Entities to consider
+     */
+    using DrawAssigner_t = std::function<
```
jonesmz, line 115: needs to be std::function? or can be function-pointer?

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+            void(ActiveScene&, Storage_t&, ArrayView_t) >;
+
+    /**
+     * @brief Set the DrawAssigner function for a certain material type
+     */
+    template<typename MATERIAL_T>
+    void set_assigner(DrawAssigner_t assigner)
+    {
+        material_id_t id = material_family_t::type<MATERIAL_T>;
+        m_assigners.resize(std::max(m_assigners.size(), size_t(id + 1)));
```
jonesmz, line 125: skeptical of this.

If the type of MATERIAL_T is known at compile time (which it must be, given this is a template)

then unless you're planning to have dlls with previously unknown materials loaded at runtime, you should be able to determine the full list of possible materials at compile time, or at the very worst, at runtime, but during the static initalization phase prior to main() running, in the case of DLL injection, or at the very very very worst, after the DLL loader finishes, but before the game engine actually starts.

As a result, m_assigners should either be a std::array<> with that compile-time size, or it should be a std::unique_ptr<DrawAssigner_t[]> with the run-time determined size (either static initialization phase, or after all DLL loading completes)

jonesmz: Also, are you checking to ensure that any particular item in the m_assigners vector is an initialized function prior to calling?

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+
+    /**
+     * @brief Adds a single entity to be considered by RenderGroups and
+     *        assigned draw functions
+     *
+     * @tparam MATERIAL_T Material of newly added entity
+     */
+    template<typename MATERIAL_T>
+    void add(ActiveEnt ent)
+    {
```
jonesmz, line 170: same argument about pre-initializing data prior to use, as above.

src/osp/Active/drawing.h
```diff
@@ -24,67 +24,202 @@
...
+        m_added[material_family_t::type<MATERIAL_T>].push_back(ent);
+    }
+
+    std::unordered_map< std::string, RenderGroup > m_groups;
```
jonesmz, line 174: consider an std::unordered_set with a custom search predicate and storing the name inside of the RenderGroup

This cuts down on the number of distinct types the compiler has to deal with, and keeps the name of the object local to the object, making it easier on other parts of the code needing to co-locate the data.

jonesmz: Also: shared_string is a good choice for write-once, read-many string types.

src/osp/Resource/AssetImporter.cpp
```diff
@@ -231,14 +231,14 @@ void osp::AssetImporter::load_plume(TinyGltfImporter& gltfImporter,
     float zMax = extras.Get("zMax").Get<double>();
     float zMin = extras.Get("zMin").Get<double>();
 
-    PlumeEffectData plumeData;
+    adera::active::PlumeEffectData plumeData;
     plumeData.m_meshName = meshName;
```
jonesmz, line 235: std::move(meshName)

src/osp/Shaders/MeshVisualizer.cpp
```diff
@@ -0,0 +1,63 @@
...
+        const ACompCamera &camera, void *pUserData) noexcept
+{
+    using Magnum::Shaders::MeshVisualizerGL3D;
+
+    auto &rShader = *reinterpret_cast<MeshVisualizerGL3D*>(pUserData);
+
+    auto &rMesh = rScene.reg_get<ACompMesh>(ent);
+    auto const& drawTf = rScene.reg_get<ACompDrawTransform>(ent);
+
+    Matrix4 entRelative = camera.m_inverse * drawTf.m_transformWorld;
```
jonesmz, line 51: const

jonesmz (approved):
Like with most PRs, i don't see anything terrible here.

I did point out some architectural considerations that I hope you address, but it's not gonna bug me if you submit this.

src/adera/Shaders/PlumeShader.cpp
Capital-Asterisk: Shaders are stored in context resources, which shares its lifetime with the GL context. Most of the time™ the shaders will outlive the draw assignment functions, but it's still possible for the shader to be unloaded if something like that is ever implemented.

Similar to a smart pointer, DependRes can be used here, stored in the lambda somehow, where it can also access the "is loaded" functionality. Now that I think of it, DependRes can possibly be replaced with, or rely upon a shared pointer.

src/adera/Shaders/PlumeShader.h
Capital-Asterisk: Some kind of non-owning `any` can help add some type checking. This part of the code may be hot, but a bit of extra overhead probably wouldn't be the bottleneck.

src/osp/Active/SysRender.cpp
Capital-Asterisk: The intention is to always get, where the key always exists. This will be changed to at()

src/osp/Active/SysRender.cpp
Capital-Asterisk: I feel that the index has significance in this context, where they are sequential IDs. It's useful to know the index when debugging, and new features may be added that rely on it.

src/osp/Active/drawing.h
Capital-Asterisk: Draw assigners rely on lambda captures. They're only called once per frame at max, so I don't believe any overhead will be an issue here.

src/osp/Active/drawing.h
Capital-Asterisk: It's possible to (easily?) modify entt::family to register new types at runtime from a config file or scripting engine. It would also be sane to just set to some large std::array (~256) since it will take a long time to reach that many material types. This function is just convenient, and it's performance isn't important since it's called only during setup.

They are checked in SysRender.cpp line 237:
```cpp
if ( ! bool(assigner) )
{
    continue; // Assigner is null, not set to a function
}
```

Capital-Asterisk: Some details on how these materials will work: they are very broad 'suggestions' that are supported by components. They are suppose to be generic enough to never require specialization for individual objects.

ie. Instead of having ClearGlass, RedGlass, GreenGlass, and BlueGlass materials, there will only be one Glass material, and a color component. The assigner for glass will be responsible for reading the color, and setting the shader.

src/adera/SysExhaustPlume.cpp
Capital-Asterisk: These are new in C++20 only

src/adera/SysExhaustPlume.cpp
Capital-Asterisk: Systems are simple static functions and loggers are objects, which makes it 'difficult' to store the loggers. I'm thinking it's better for the loggers to be passed as arguments eventually. Or heck, maybe even a globally accessible per-thread 'currently selected logger.'

src/adera/SysExhaustPlume.cpp
jonesmz: really? that's obnoxious. My bad.

src/adera/SysExhaustPlume.cpp
jonesmz: that's fair too. could have a `thread_local std::reference_wrapper<Logger>` for the currently active logger (reference to indicate "not null").

How would we swap the logger though, if the different systems are being executed by a thread pool?

Logger can't be stored in entt? Certainly we could store std::reference_wrapper<Logger> in entt ?

src/osp/Active/drawing.h
Capital-Asterisk: I'll leave this as-is due to the 'flat' nature of how this project is structured. The RenderGroup isn't really responsible for anything outside of storing its own data. It's unlikely for other code to need to store references to render groups for long term.

I remember a fixed version of shared_string being somewhere, but as of now it still doesn't compile.

jonesmz: What...? I fixed it like a month ago.

What errors did you get?

EhWhoAmI (approved):

src/adera/Shaders/PlumeShader.h
z-adams: Shouldn't be that hot, since it'll only be called at most (number of thrusters onscreen) times per frame.

Actually, what am I saying. If KSP is any indication, that number might be pretty large.

In any case, at that stage it's entirely possible that much of this framework will have been rewritten for higher performance, so I wouldn't worry about it too much. If you've ever seen one of those "frame breakdown" videos from graphics-hungry games, they're always absurdly complicated, so it's safe to assume our renderer will undergo a lot of evolution over time, and the notion of drawing individual objects via a shader will likely no longer exist.

src/adera/SysExhaustPlume.cpp
```diff
@@ -39,52 +41,142 @@
...
-    // Get plume mesh
-    Package& glResources = rScene.get_context_resources();
-    DependRes<Mesh> plumeMesh = glResources.get<Mesh>(plumeEffect->m_meshName);
+    // Compile the plume mesh if not done so yet
     if (plumeMesh.empty())
     {
-        plumeMesh = AssetImporter::compile_mesh(plumeEffect->m_meshName, pkg, glResources);
+        plumeMesh = AssetImporter::compile_mesh(
+                        plumeEffect->m_meshName, rPkg, rGlResources);
     }
```
z-adams, line 138: I seem to remember writing `compile_mesh()` for the specific purpose of simplifying this process, but since this get-or-compile process will probably be used every time a resource is fetched like this, it might be worth writing a specific function called `get_or_load` or similar that basically just does


```cpp
template<typename T>
DependRes<T> get_or_load(std::string_view resName)
{
    Package &rPkg = rScene.get_application().debug_find_package("lzdb");
    Package& rGlResources = rScene.get_context_resources();
    DependRes<T> res = rGlResources.get<T>(resName);
    if (res.empty())
    {
        res = AssetImporter::compile_resource<T>(resName, rPkg, rGlResources);
    }
    return res;
}
```

Not strictly pertinent to this PR, but could be a good time to do it.

Also, we should probably move away from the name `rGlResources` and use `rGfxResources` or similar since presumably we will replace the GL backend in the future or offer multiple choices, so we should aim to be as API agnostic as possible (although it's just a name so obviously it's not a big deal).

src/osp/Active/SysRender.cpp
```diff
@@ -138,24 +153,29 @@ RenderPipeline SysRender::create_forward_renderer()
...
-
-            // Draw frontfaces
-            Renderer::setFaceCullingMode(Renderer::PolygonFacing::Back);
-            SysRender::draw_group(rScene, transparentView, camera);
+                    Renderer::BlendFunction::SourceAlpha,
+                    Renderer::BlendFunction::OneMinusSourceAlpha);
+
+            // temporary: disabled depth writing makes the plumes look nice, but
+            //            can mess up other transparent objects once added
+            Renderer::setDepthMask(false);
```
z-adams, line 165: Not sure if I understand how this works without doing several passes?

z-adams (approved):
See above comments. Otherwise looks good.

src/adera/SysExhaustPlume.cpp
z-adams: I see that you used `rCtxResources` elsewhere in which case carry on

src/osp/Active/SysRender.cpp
Capital-Asterisk: This makes the plumes just not write any new depth data, so they render on top of each other without care.

src/adera/SysExhaustPlume.cpp
Capital-Asterisk: This is likely to be  moved into a dedicated system to decouple gl-specific functions from systems that create meshes, and better control which thread they are run on.

src/osp/Active/SysRender.cpp
z-adams: Ah I see. It would be better to just sort the objects by depth and use the method I had above. The two-pass transparency is among the highest performance transparency rendering algorithms there is, the only issue is that it doesn't do per-pixel sorting for overlapping objects but that's an unsolved problem in computer graphics as far as I'm aware.

PR #153, opened 2021-09-11
Implement Thread-local logging

Capital-Asterisk:
Step 2 to removing OSPApplication. Important bits are in `src/osp/logging.h` and `src/test_application/main.cpp`.

Storing loggers in thread-local storage avoids global state, without the inconvenience of having to locate them somewhere or passing through function arguments. 

Loggers can be switched between using `osp::set_thread_logger` on the application-level.

```cpp
set_thread_logger(systemALogger);

system_a();

set_thread_logger(systemBLogger);

system_b();
```

Due to the nature of this project, its unlikely that systems (or just most functions in general) would directly call functions into other 'services' (synchronous message passing). This makes thread-local loggers a practical choice.

Contrary what's written above, systems like SysVehicle directly calls AssetImporter, but this will likely not be the case on first release.

Loggers are all managed in main.cpp. This can be moved elsewhere in the future, as long as they're all created or configured in one place.

### DIFF
--- a/src/osp/logging.h
+++ b/src/osp/logging.h
@@ -1,6 +1,6 @@
 /**
  * Open Space Program
- * Copyright © 2019-2020 Open Space Program Project
+ * Copyright © 2019-2021 Open Space Program Project
  *
  * MIT License
  *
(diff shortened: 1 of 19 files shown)


src/osp/logging.h
```diff
@@ -0,0 +1,48 @@
...
+namespace osp
+{
+
+using logger_t = std::shared_ptr<spdlog::logger>;
+
+inline thread_local logger_t t_currentLogger;
+
+inline void set_thread_logger(logger_t const logger)
+{
+    t_currentLogger = logger;
```
jonesmz, line 38: since logger_t is a std::shared_ptr, take the argument by value, and std::move() it.

jonesmz (approved):

jonesmz:
Great replacement. Very clean and drop-in.

EhWhoAmI (approved):
This change looks good, this is certainly a more elegant way of implementing a logger than the way I implemented loggers before.

z-adams (approved):
Slick

PR #156, opened 2021-09-24
Refactor OSPApplication and OSPMagnum, and Improve Package

Capital-Asterisk:
Depends on #153

Final step to removing OSPApplication.

* OSPApplication only stores packages, so it's renamed to PackageRegistry and moved to `src/osp/Resource`.
* OSPMagnum is renamed to ActiveApplication. This better hints at its purpose in the program along with relevant docs.
* Packages don't need to store their own name prefix; these members were removed. the `group_get()` function was made const-friendly.
* Variable names and docs in many other places were changed accordingly

Important changes in files:
* `src/test_application/ActiveApplication.cpp`
* `src/osp/Resource/PackageRegistry.cpp`
* `src/osp/Resource/Package.h`

### DIFF
--- a/src/osp/Active/ActiveScene.h
+++ b/src/osp/Active/ActiveScene.h
@@ -148,7 +148,7 @@ public:
     constexpr PackageRegistry& get_packages() const { return m_rPackages; };
 
     constexpr Package& get_context_resources() { return m_rContext; }
-    constexpr Package& get_context_resources() const { return m_rContext; }
+    constexpr Package const& get_context_resources() const { return m_rContext; }
 
 private:
 
--- a/src/osp/Active/ActiveScene.h
+++ b/src/osp/Active/ActiveScene.h
@@ -53,9 +53,7 @@ class ActiveScene
 
 public:
 
-    ActiveScene(PackageRegistry& rApp, Package& rContext);
-
-    PackageRegistry& get_application() { return m_app; };
+    ActiveScene(PackageRegistry& rPkgs, Package& rContext);
 
     /**
      * @return Root entity of the entire scene graph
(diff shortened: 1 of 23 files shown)


src/osp/Active/ActiveScene.h
```diff
@@ -135,18 +133,27 @@ class ActiveScene
...
     // TODO
     constexpr float get_time_delta_fixed() const { return 1.0f / 60.0f; }
 
-    Package& get_context_resources() { return m_context; }
+
+    constexpr PackageRegistry& get_packages() { return m_rPackages; };
+    constexpr PackageRegistry& get_packages() const { return m_rPackages; };
+
+    constexpr Package& get_context_resources() { return m_rContext; }
+    constexpr Package& get_context_resources() const { return m_rContext; }
```
jonesmz, line 151: If they both return a non-const refernece to Package, why bother defining the non-const version of the function?

src/osp/Resource/PackageRegistry.h
```diff
@@ -24,49 +24,58 @@
...
-#include "Universe.h"
-#include "Resource/Package.h"
+#include <map>
 
 namespace osp
 {
 
+// Prefix for resource paths to indicate which package they're from
+// Ideally, this would be some kind of short string
+using ResPrefix_t = std::string;
```
jonesmz, line 36: boost::static_string, or std::array.

jonesmz (approved):
Everything looks ok to me.

Mostly you just renamed a bunch of stuff.

src/osp/Resource/PackageRegistry.h
z-adams: The prefixes are typically short, so small string optimization will probably be able to contain the whole thing in 99% of cases. `std::basic_string` has a buffer size of 16 chars in the STL implementation I've got for MSVC, and I think the gcc equivalent is even longer. The prefixes were originally going to be only 4 characters but 16 is a ton so I think it's probably fine. If we get to the point where `std::string` would really be a bad idea here, we'd be moving away from string-based keys anyways.

z-adams (approved):

src/osp/Active/ActiveScene.h
Capital-Asterisk: This is probably a mistake

src/osp/Resource/PackageRegistry.h
Capital-Asterisk: std::string here works for now; it's not in a performance critical path either. Regardless of whatever implementation is used, stuff is still going to run a couple hundred times faster than a python program. I'd leave this as is, but open to a drive-by PR.

PR #157, opened 2021-11-05
Add new WIP Planet system, Id Registries, and Remove old planet system

Capital-Asterisk:
> Capital-Asteirk:  August 20, 2021
> most of the everything-breaking major changes are done™️

The planet system is potentially the most convoluted and messed up part of the code. While trying to cleaning it up, I soon realized that its incredible complexity was due to a fundamental flaw: I focused too much on the triangles and chunks that make up the icosahedron planet, but haven't considered the importance of the vertices shared between them.

Additional considerations on making more permanent structures for planets calls upon a few more major changes needed to the engine. It's rather uncertain whether data should be stored as an ECS component, stored in the registry's context, or somewhere else. There is also no go-to place to store data between scenes.

There is quite a bit of new code, but I'm fairly confident that they are well isolated and easy to understand.

## IdRegistry - The important part

See unit tests: `test/id_registry/main.cpp`

Many parts of the planet code (both old and new), CoordinateSpaces, Wiring, and ECS in general, rely on throwing around and managing ownership of array indices. Array indices have many benefits over pointers: allowing structures of arrays, not needing to know the type they refer to, no problems with unstable memory, etc... 

The IdRegistry's job is to easily generate and manage sequential IDs to be used as array indices starting from zero. When IDs are deleted, they are re-used when new IDs are created. Enum classes can be used as IDs to prevent them from getting mixed up with other integer types.

EnTT registry does a very similar thing to manage entities, but the feature is glued into the registry and not available on its own.

### Hierarchical Bitset

Unit test also available: `test/hierarchical_bitset/main.cpp`

On the inside, the IdRegistry uses an obscure datastructure called the Hierarchical Bitset. There doesn't seem to be very many resources on this.

I first read about "Hierarchical dirty flags" from this post from a developer of the Our Machinery engine:
https://www.gamedeveloper.com/programming/syncing-a-data-oriented-ecs-with-a-stateful-external-system

This structure can very quickly iterate all bits that are set to 1, which I find perfect for the IdRegistry for detecting which IDs are deleted/available.

A bit more digging led to hbitset used in the Amethyst engine:
https://github.com/amethyst/hibitset

I couldn't find a C++ implementation, so I decided to write one. 

My implementation uses a single allocation of 64-bit integers to store rows of bits. This takes advantage of 64-bit registers and instructions, most notably the 'count trailing zeros' instruction. In theory this is really fast, though never benchmarked.

### Safety / Reference counting

[...]

Issue #155, opened 2021-09-18
Beef up main.cpp, flight.cpp, and Test Application in general


Capital-Asterisk:
Files in question are `src/test_application/main.cpp` and `src/test_application/flight.cpp`

Some problems:
* It's getting pretty ugly, such as lots of repeated code in main.cpp
* The window needs to be closed to change scenes. This is annoying and not final.

New things to consider
* Unified commands for REPL and command line arguments.
* Support for switching scenes for a menu and a vehicle editor.
* Separate universe and scene updates. As of now, the universe is updated each frame by the flight scene.

The way to go about this would be to have some commands to better manage the Magnum thread. The Magnum thread can then receive messages (or just mutexed) from the main thread to setup and switch scenes.

Most of the infrastructure is already present, and most of the changes are needed in main.cpp and flight.cpp. If done well, no changes need to be made to the OSPMagnum class (soon to be renamed to ActiveApplication) or anything else.

Capital-Asterisk:
Fixed in #158

PR #158, opened 2021-11-16
Restructure to registry-less ECS

### LINKED ISSUE #74: Abstract GL resource compilation away from systems
Right now, `SysVehicle::part_instantiate` has code that attempts to get the resources required by the vehicle, and if they don't yet exist, attempts to compile them from image/mesh data in resource packs. This pattern is repeated consistently in several places in the repo, and will likely continue to be common. We may therefore want to consider abstracting this resource check away, so that systems can call a single function to fetch the required data without having to manually check for and/or compile the GL resources every time.

Capital-Asterisk:
There are many casualties, but lots of cool things as a result.

Depends on #157 

**Almost everything is disabled and many parts of the code remain unused. Only a few files are important to review.**

Many files have an `#if 0` block covering the whole thing. Please ignore those files.

## Problem

The entt::registry is designed for the common case, but had started to become a burden as the project's architecture evolved.

Consider that the game scene (ActiveScene) needs to be synchronized to a Universe, associated with resources tied to the GPU, and access data shared between scenes. It may make sense to store references to each of these external dependencies, but instead, it's far more practical (in the case of ECS) to pass them all as function parameters. The application that calls all the functions/systems will have easy access to each of the dependencies. The case right now is that the systems themselves service-locate, which unneeded complexity and is harder to control from outside.

For example, if we want to manage thread-local instances of a component pool, dirty flags, or queues, it would be really easy to handle them at the application-level and pass them to system functions as parameters. The systems themselves needing to locate the thread-local instances would be a headache in comparison.

The best system function is one that has very little to no side effects, and only modifies whatever mutable parameters its given. This also means it's best to pass individual component pools and views as parameters, instead of the entire scene each time. This makes it obvious which components a system modifies, as they can be marked const too, and makes it completely independent of the entt::registry. In my opinion, passing around a huge centralized class is an anti-pattern nonetheless.

Other problems with the registry:
* It's used to store _everything_ because it's more convenient to access
* Things that didn't need to be components were made into components anyways
* Many parts of the project are already designed to work around the registry's shortcomings. ie. Coordinate spaces

I'm confident that the root cause to many of these problems are because of the registry.

## Changes

Many parts of the code were touched, mostly to replace ActiveScene from their arguments to individual views and storages. There were changes to things related to universe, machines, and vehicles, but none of them are used yet.

All the broken systems will be reworked in following PRs.

### No more centralized scene state

RIP ActiveScene

"Scene" now refers to any structure that stores the state of a game world (with entities and components).

Instead of using the entt::registry, components are stored as entt::basic_storages. For example, see [basic.h](https://github.com/Capital-Asterisk/osp-magnum/blob/no-registry/src/osp/Active/basic.h); acomp_storage_t is typedefed in [activetypes.h](https://github.com/Capital-Asterisk/osp-magnum/blob/no-registry/src/osp/Active/activetypes.h).

System functions can accept views to read or modify components, or take a reference to the storage to be able to create or destroy components.

Collecting together all the component storages into some structure (such as a struct) is up to the application. The system functions won't be aware of how the component storages are stored. IdRegistry can be used to generate entity IDs.

The Universe class is not dead, since it's responsible for storing CoordinateSpaces.

### Simplified Render system

Lots of complexity was turned into simple functions. Many structures didn't need to be tied to the scene, or didn't need to exist at all. The GPU functions that get called are still the same, but are less tangled and far easier to follow. Additionally, OpenGL-specific code from SysRender was split into SysRenderGL, cleanly separating it from any scene logic.

This fixes #74

### Revamped Test Application - The important part

A small self-contained 'Engine Test' is added to demonstrate osp-magnum's use as a game engine. See [scenarios_enginetest.cpp](https://github.com/Capital-Asterisk/osp-magnum/blob/no-registry/src/test_application/activescenes/scenarios_enginetest.cpp). All it does right now is display a spinning cube.

The renderer state is now entirely separated from the scene logic. The Magnum application (ActiveApplication) can be closed and re-opened without needing to destruct the scene. This will also make it incredibly easy to run a scene in the background.

I'd highly recommend reading through `scenarios_enginetest.cpp` more thoroughly, as it should give a good understanding of the new API.

### main.cpp rework

This partially addresses #155

Lots of things were removed; again, no universe stuff is functional. A map of commands for loading scenes was added. Scenes are now created separately from the Magnum thread, as they are no longer coupled to gpu resources.


src/test_application/activescenes/scenarios_physicstest.cpp
```diff
@@ -0,0 +1,358 @@
...
+    rScene.m_drawing.m_diffuseDirty.assign(std::begin(rMeshSet), std::end(rMeshSet));
+
+    return [&rScene, pRenderer = std::move(pRenderer)] (ActiveApplication& rApp)
+    {
+        update_test_scene(rScene);
+        render_test_scene(rApp, rScene, *pRenderer);
+    };
+}
+
+} // namespace testapp::physicstest
```
Capital-Asterisk, line 358: This physics test scene is currently incomplete; ignore this entire file.

src/test_application/activescenes/scenarios_enginetest.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+constexpr int const gc_maxMaterials = 2;
+
+/**
+ * @brief State of the entire engine test scene
+ */
+struct EngineTestScene
+{
+    // ID registry generates entity IDs, and keeps track of which ones exist
+    osp::IdRegistry<osp::active::ActiveEnt> m_activeIds;
```
z-adams, line 72: Does this actually completely replace the EnTT entity system? I suppose most of the ent-related functions we're used to are members of the component pools. Does this change preclude us from using functions like `reg.all_of<T...>(entity)` to check for component existence, or `reg.valid(entity)`? Are there replacements for these functionalities?

z-adams (approved):
Much cleaner, I like it.

src/test_application/activescenes/scenarios_enginetest.cpp
Capital-Asterisk: As far as I'm concerned, this works as a more capable replacement. Memory usage is lower, more efficient recycling of deleted entity IDs, but the main benefits come from the flexibility of manually managing component pools. The only thing we lose is the existing automation previously provided by the registry (and also entity versions, which we can implement if we really want to). From working on other branches, passing component pools as function parameters eliminates the need for many of the convenient functions in the first place.

`reg.all_of<>` on the inside is calling contains() on each sparse set in the registry. There aren't any instances of all_of being called with any more than a single component in the codebase. `reg.valid(ent)` is roughly equivalent to `m_activeIds.exists(ent)`.

PR #159, opened 2021-12-17
Combine all builds into one file, and enable skipping of non code files

### LINKED ISSUE #150: Make the continuous integration ignore documentation changes
There should be some way to force it to not run if the only changes are to markdown files or something.

EhWhoAmI:
This PR adds a workflow step that checks if the build has been built before, or if the files edited in the current commit are markdown files, and skips if any of these conditions are met.

Closes #150


z-adams (approved):

jonesmz (changes requested):
I'm a big fan of skipping the build if none of the cpp files have changed, but could i ask you to preserve the one file per platform aspect of how the continuous integration is setup? I think that's an important separation.

.github/workflows/linux.yml
```diff
@@ -7,7 +7,19 @@ on:
     types: published
 
 jobs:
+  skip-duplicates:
```
jonesmz, line 10: You shouldn't need to have this be its own job.

I'd prefer just adding the step to the existing job(s) before the checkout step.

Is that possible?

jonesmz:
Can you also make the analyzers skip? 

There are two jobs in the analyzers file. You'll have to add the check to both jobs.

Capital-Asterisk (approved):

jonesmz:
Isn't there a way to abort the whole job right away if the skip checker says to skip?

My reason for not wanting to do a second "job" per platform is because the startup cost of github launching another runner can take a long time.

But if the only way to add this without adding a whole new job is to add if checks to every step, then I'd rather pay the cost of doing a new job per platform.

Could you let me know what the capabilities are?

.github/workflows/analyzers.yml
```diff
@@ -38,6 +48,12 @@ jobs:
       run: |
         cmake -B build -DCMAKE_BUILD_TYPE=${{ matrix.config }}
 
+    - name: Download Clang Analysis Driver Scripts
+      run: |
+        wget https://raw.githubusercontent.com/llvm/llvm-project/llvmorg-11.0.1/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py
```
jonesmz, line 53: Looks like you're out if date with master. These lines shouldn't be added back.

EhWhoAmI:
I don't think it's possible to exit a workflow prematurely, so the current method would have to do.

EhWhoAmI:
Is this ready for merging?

jonesmz:
Had git merge problems. Merged via https://github.com/TheOpenSpaceProgram/osp-magnum/pull/167

PR #161, opened 2021-12-18
Port physics & Newton Dynamics to new architecture, Add physics test, Add delete systems

Capital-Asterisk:
This PR ports SysNewton to the new architecture, adds a new self-contained physics test in `src/test_application/activescenes/scenarios_physicstest.cpp`, along with other changes to support them.

### Physics & Newton Dynamics

Not much has really changed besides no longer relying on the entt registry. All components are passed as function parameters, as well as improved multithreading considerations. Many of these changes were applied in the previous PR #157.

SysNewton is no longer responsible for calculating inertia. Inertia calculations don't rely on the physics engine, hence were separated out. Rigid bodies can now be assigned any inertia value.

The inertia calculations have been separated out into ACtxHierBody, generic physics components in `physics.h` that describes the mass and inertia of entities in the hierarchy. A future system can use these components to calculate values that can be passed to any physics engine.

### Camera Controller

The new camera controller actually only controls the camera and isn't a god class anymore. The camera orbits around an imaginary 'target' point, similar to blender and a few other programs.

Default controls, assignable in settings.toml:
* W/S - move camera forward and back
* A/D - move camera left and right
* Q/E - move camera up and down
* RMB - orbit around with mouse
* MouseWheel - Zoom in/out

Zooming out makes the camera move faster

### Physics test scene

The physics test consists of a 64x64 floor; a box and a cylinder are created every 2 seconds. The renderer uses the new Camera Controller and its controls. Pressing space throws a sphere.
image

### Deleter systems

For the physics test scene, entities that are going to be deleted are added to vectors. Near the end of the update function, these vectors are passed to functions to delete their corresponding components. Once all the components are deleted, the entity ID can be deleted and reused. Since these functions use iterators, containers other than vectors can be used eventually, such as the Hierarchial bitset, which btw now have iterators in the new [Longeron++ library](https://github.com/Capital-Asterisk/longeronpp).

### Other changes:

* Fixed shaders
* Added Cube, Cylinder, Sphere, and Grid primitives added in main.cpp
* Added delta time to scene updates

Still no vehicles, flight, or anything too fancy.

### DIFF
--- a/bin/settings.toml
+++ b/bin/settings.toml
@@ -103,8 +103,38 @@ primary = "Right"
 secondary = "None"
 holdable = true
 
-[ui_rmb]
+[cam_orbit]
 primary = "RMouse"
+secondary = "None"
+holdable = true
+
+[cam_fd]
+primary = "W"
+secondary = "None"
+holdable = true
+
+[cam_bk]
+primary = "S"
+secondary = "None"
+holdable = true
+
+[cam_lf]
+primary = "A"
+secondary = "None"
+holdable = true
+
+[cam_rt]
+primary = "D"
+secondary = "None"
+holdable = true
+
+[cam_up]
+primary = "Q"
+secondary = "None"
+holdable = true
+
+[cam_dn]
+primary = "E"
 secondary = "None"
 holdable = true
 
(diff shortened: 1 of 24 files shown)


z-adams (approved):

jonesmz:
Go ahead and merge this as soon as you've resolved the conflict.

Stop waiting so long for other people to review :-)

PR #174, opened 2022-01-28
Add new Resources class to eventually replace Package

Capital-Asterisk:
While considering better ways to associate vehicle blueprints with glTF and its images, textures, and mesh data, it became clear that the Package class is actually rather incapable. Its heavy reliance on string identifiers and resources being coupled to a single type makes them difficult and hacky to store. This calls upon the need for a successor.

This pull requests adds the new 'Resources' class, a centralized class intended to store data from loaded files. Resources is not responsible for any file loading, it only stores and organizes data.

See unit test in `test/resources/main.cpp` for basic usage.

Insides of the Resources class can be visualized like this (simplified):
ospresources

*Example with 2 different resource types (ResTypeId), each with 4 resources coming from 2 different packages.*

Resources can be obtained by providing their Type (ResTypeId), Package Id (PkgId), and name identifier.

Similar to all the new code I've been writing, integer IDs are heavily used. Ids are easier to compare and pass around (compared to the old DependRes). They can be stored without needing to know what data type(s) it's associated with it, which should improve compile times and reduce coupling previously required by DependRes.

Number of resource types and data per resource type is set at runtime during setup. Since there aren't going to be many data types associated with each resource type, a simple linear search is used to find the right data container.

The old Package class used std::optionals to reserve resource. The equivalent to reserving is to create a resource Id, but with no data assigned to it.

Reference counting is handled using Longeron IdStorage, see unit tests. Reference counts don't do anything yet. Like before, there is no way to remove resources, too lazy atm :p

edit: updated diagram fixing 'huge mistake'


test/resources/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+##
+PROJECT(test_resources CXX)
+ADD_TEST_DIRECTORY(${PROJECT_NAME})
+
+# workaround?
+TARGET_LINK_LIBRARIES(test_resources longeron EnTT::EnTT)
+TARGET_SOURCES(test_resources PRIVATE "${CMAKE_SOURCE_DIR}/src/osp/Resource/resources.cpp")
```
Capital-Asterisk, line 30: unsure about this part

Capital-Asterisk:
Huge mistake here: resource names are suppose to be stored per-package to prevent name conflicts.

z-adams (approved):
I haven't fully inspected the inner workings of the index stuff and refcounting, but the interface and behavior of the resources stuff is cool, merge when you think it's ready

jonesmz (approved):

PR #176, opened 2022-02-05
Replace 'GL Resources' Package with RenderGL struct

Capital-Asterisk:
Depends on #171 

Most of the action happens in `src/osp/Active/opengl/SysRenderGL.h`.

The GL Resources Package was simply an osp::Package stored in ActiveApplication that stores instances of meshes, texture, and shaders on the renderer-side. Its reliance on string names and general inflexibility makes it hard to manage and is currently misused in a few places. This is now removed.

The new RenderGL struct is more specialized and serves a similar purpose, also stored in ActiveApplication. It's responsible for storing essential rendering resources that are accessible across multiple scenes. Renderer-side meshes and textures are each assigned integer IDs: MeshGlId and TexGlId. This makes them addressable regardless of their source, either from a framebuffer or loaded from a resource.

This is of course part of a gradual phase-out of the Package class. The old mesh and texture components are still used (ACompMesh and ACompTexture) which rely on Package resources. The string names of the resources are temporarily mapped to the new mesh and texture Ids, noting m_oldResToTex and m_oldResToMesh.

### DIFF
--- a/src/adera/Shaders/PlumeShader.cpp
+++ b/src/adera/Shaders/PlumeShader.cpp
@@ -27,16 +27,10 @@
 #include <adera/Plume.h>                  // for PlumeEffectData
 #include <adera/SysExhaustPlume.h>        // for ACompExhaustPlume
 
-#include <osp/Resource/Package.h>         // for Package
-#include <osp/Resource/Resource.h>        // for DependRes
-#include <osp/Resource/AssetImporter.h>   // for AssetImporter
-
 #include <osp/Active/basic.h>             // for ACompCamera
 #include <osp/Active/drawing.h>           // for ACompDrawTransform
 #include <osp/Active/SysRender.h>         // for ACompMesh
 #include <osp/Active/activetypes.h>       // for basic_sparse_set, ActiveEnt
-
-#include <osp/Resource/PackageRegistry.h> // for PackageRegistry
 
 #include <Magnum/GL/Shader.h>             // for Shader, Shader::Type
 #include <Magnum/GL/Texture.h>            // for Texture2D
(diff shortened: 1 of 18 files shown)


src/adera/Shaders/PlumeShader.cpp
```diff
@@ -80,9 +80,9 @@ void PlumeShader::draw_plume(
     ACompDrawTransform const &drawTf    = rData.m_pDrawTf->get(ent);
     ACompExhaustPlume const &comp       = rData.m_pExaustPlumes->get(ent);
     PlumeEffectData const &effect       = *comp.m_effect;
-    Magnum::GL::Mesh &rMesh             = *rData.m_pMesh->get(ent).m_mesh;
+    Magnum::GL::Mesh &rMesh             = rData.m_pMeshGl->get(rData.m_pMeshId->get(ent));
```
jonesmz, line 83: Not sure how exited I am with a "get()" function returning a reference.

Maybe do "at()" ?

src/osp/Active/SysRender.h
```diff
@@ -128,8 +162,45 @@ class SysRender
...
+    {
+        return; // Stop recursing when about to reach root
+    }
+
+    // Tail recurse into parent
+    add_draw_transforms_recurse(hier, rDrawTf, currHier.m_parent);
+}
+
+template<typename IT_T>
+void SysRender::assure_draw_transforms(
```
jonesmz, line 193: consider taking two template args to support C++20's sentinal interators

jonesmz (approved):

src/osp/Active/opengl/SysRenderGL.cpp
```diff
@@ -182,16 +187,18 @@ DependRes<Texture2D> try_compile_texture(
             .setStorage(1, textureFormat((*texData).format()), (*texData).size())
             .setSubImage(0, {}, view);
 
+        return newId;
+
```
jonesmz, line 191: extra space?

src/osp/Shaders/Phong.cpp
```diff
@@ -57,14 +57,17 @@ void shader::draw_ent_phong(
...
+        {
+            rShader.bindAmbientTexture(rTexture);
+        }
 
-    if (rShader.flags() & (Flag::DiffuseTexture | Flag::AmbientTexture | Flag::AlphaMask))
-    {
-        rShader.bindAmbientTexture(*rData.m_pDiffuseTexGl->get(ent).m_tex);
     }
 
+
```
jonesmz, line 70: extra space?

src/osp/Shaders/Phong.cpp
```diff
@@ -81,7 +84,7 @@ void shader::draw_ent_phong(
         .setTransformationMatrix(entRelative)
         .setProjectionMatrix(viewProj.m_proj)
         .setNormalMatrix(Matrix3{drawTf.m_transformWorld})
-        .draw(*rData.m_pMeshGl->get(ent).m_mesh);
+        .draw(rData.m_pMeshGl->get(rData.m_pMeshId->get(ent)));
```
jonesmz, line 87: little awkward to need to do 2 lookups here

src/adera/Shaders/PlumeShader.cpp
Capital-Asterisk: This is an entt::storage, it doesn't use exceptions and the `at` function does something entirely different. The entities there should always be valid unless something goes wrong; there's asserts in the get function for that.

jonesmz (approved):

src/adera/Shaders/PlumeShader.cpp
jonesmz: I wasn't really commenting on the possibility for exceptions.

I'm just used to thinking of "get()" always returning a pointer, that i find it odd that we're getting a reference back.

But i guess when in rome...

src/osp/Active/SysRender.h
Capital-Asterisk: Hmm, just looked this up and I'm already a fan of them. Unfortunately, the subsequent functions used in many other places need to support them too. I'll consider adding this to Longeron++.

z-adams (approved):
Looks good, merge it. The GL context resources are pleasantly simple

PR #177, opened 2022-02-06
Try fixing macos

jonesmz:


jonesmz:
Turning on GNU compiler extensions until SDL2 fixes their build in a later release

See https://github.com/libsdl-org/SDL/issues/5319

PR #179, opened 2022-03-22
Switch to new resource system

Capital-Asterisk:
Details on how the `osp::Resources` class works is here: #174

This PR switches the test scenes away from the old resource system to the new one, most notable the mesh and texture components.

### Scene Mesh and Textures

see `src/osp/Active/drawing.h`

Instead of indirectly storing a pointer to a Texture resource through DependRes<TextureData>, the scene itself now keeps its own set of Texture Ids (TexId). Some of these 'scene-space' texture IDs can be associated with a Resource Id (ResId).

This also allows scenes to have textures that aren't associated with a resource, ie. in-game screens that require render-to-texture. All this applies to meshes too, ie. planet terrain mesh. The task of associating with a resource naturally decoupled from the scene itself; consequently, this also entirely decouples the scene's dependency on any resource system >:)

All of the meshes and texture resources used by the scene are easily accessible, stored in a fast-to-iterate entt dense hash map.

Copying a mesh in the scene, does not need to increment the reference count of the Resources class. The scene itself holds only a single 'owner' instance of each unique mesh/texture it uses. The scene instead keeps its own reference counts for its mesh and texture Ids.

### Reference counting and owners

A bit of a refresher: the Longeron IdStorage class is a move-only wrapped integer Id, that can only be modified by a single friend class. This will soon be renamed to IdOwner to better describe its intentions.

Containers can use IdOwners as safe reference-counted integer Ids that does not require pointers to the container. A container can have functions to create and destroy owners. Owners cannot be copied (unless unsafe code is used), and will assert when destructed while still holding a value. A container can keep track of how many owners it creates and destroys to work as a reference count.

Owner cleanup needs to be done manually; but as an upside, they have little side effects compared to typical RAII like in shared_ptr. This makes them potentially free of overhead on release builds, and just compile down to a single int.

see `test/resources/main.cpp`

Owners are used for resources, as ResIdOwner_t, as well as meshes and textures: TexIdOwner_t and MeshIdOwner_t.

### DIFF
--- a/src/osp/Active/SysRender.h
+++ b/src/osp/Active/SysRender.h
@@ -26,6 +26,8 @@
 
 #include "drawing.h"
 
+#include <unordered_map>
+
 namespace osp::active
 {
 
--- a/src/osp/Active/SysRender.h
+++ b/src/osp/Active/SysRender.h
@@ -119,11 +119,38 @@ class SysRender
 {
 public:
 
-    static MeshId own_mesh_resource(ACtxDrawing& rCtxDrawing, ACtxDrawingRes& rCtxDrawingRes, Resources &rResources, ResId resId);
+    /**
+     * @brief Attempt to create a scene mesh associated with a resource
+     *
+     * @param rCtxDrawing       [ref] Drawing data
+     * @param rCtxDrawingRes    [ref] Resource drawing data
+     * @param rResources        [ref] Application Resources containing meshes
+     * @param resId             [in] Mesh Resource Id
+     *
+     * @return Id of new mesh, existing mesh Id if it already exists.
+     */
+    static MeshId own_mesh_resource(
+            ACtxDrawing& rCtxDrawing,
+            ACtxDrawingRes& rCtxDrawingRes,
+            Resources &rResources,
+            ResId resId);
 
+    /**
+     * @brief Remove all mesh and texture components, aware of refcounts
+     *
+     * @param rCtxDrawing       [ref] Drawing data
+     */
     static void clear_owners(ACtxDrawing& rCtxDrawing);
 
-    static void clear_resource_owners(ACtxDrawing& rCtxDrawing, ACtxDrawingRes& rCtxDrawingRes, Resources &rResources);
+    /**
+     * @brief Dissociate resources from the scene's meshes and textures
+     *
+     * @param rCtxDrawingRes    [ref] Resource drawing data
+     * @param rResources        [ref] Application Resources
+     */
+    static void clear_resource_owners(
+            ACtxDrawingRes& rCtxDrawingRes,
+            Resources &rResources);
 
     inline static void add_draw_transforms_recurse(
             acomp_storage_t<ACompHierarchy> const& hier,
--- a/src/osp/Active/SysRender.h
+++ b/src/osp/Active/SysRender.h
@@ -119,6 +119,12 @@ class SysRender
 {
 public:
 
+    static MeshId own_mesh_resource(ACtxDrawing& rCtxDrawing, ACtxDrawingRes& rCtxDrawingRes, Resources &rResources, ResId resId);
+
+    static void clear_owners(ACtxDrawing& rCtxDrawing);
+
+    static void clear_resource_owners(ACtxDrawing& rCtxDrawing, ACtxDrawingRes& rCtxDrawingRes, Resources &rResources);
+
     inline static void add_draw_transforms_recurse(
             acomp_storage_t<ACompHierarchy> const& hier,
             acomp_storage_t<ACompDrawTransform>& rDrawTf,
--- a/src/osp/Active/SysRender.h
+++ b/src/osp/Active/SysRender.h
@@ -208,19 +208,42 @@ void SysRender::assure_draw_transforms(
     }
 }
 
+template<typename STORAGE_T, typename REFCOUNT_T>
+void remove_refcounted(
+        ActiveEnt ent, STORAGE_T &rStorage, REFCOUNT_T &rRefcount)
+{
+    if (rStorage.contains(ent))
+    {
+        auto &rOwner = rStorage.get(ent);
+        if (rOwner.has_value())
+        {
+            rRefcount.ref_release(rOwner);
+        }
+        rStorage.erase(ent);
+    }
+}
+
 template<typename IT_T>
 void SysRender::update_delete_drawing(
         ACtxDrawing &rCtxDraw, IT_T first, IT_T last)
 {
-    rCtxDraw.m_opaque           .remove(first, last);
-    rCtxDraw.m_transparent      .remove(first, last);
-    rCtxDraw.m_visible          .remove(first, last);
-    rCtxDraw.m_mesh             .remove(first, last);
-    rCtxDraw.m_diffuseTex       .remove(first, last);
-
-    for (MaterialData& rMat : rCtxDraw.m_materials)
+    while (first != last)
     {
-        rMat.m_comp.remove(first, last);
+        ActiveEnt const ent = *first;
+        rCtxDraw.m_opaque       .remove(ent);
+        rCtxDraw.m_transparent  .remove(ent);
+        rCtxDraw.m_visible      .remove(ent);
+
+        // Textures and meshes are reference counted
+        remove_refcounted(ent, rCtxDraw.m_diffuseTex, rCtxDraw.m_texRefCounts);
+        remove_refcounted(ent, rCtxDraw.m_mesh, rCtxDraw.m_meshRefCounts);
+
+        for (MaterialData& rMat : rCtxDraw.m_materials)
+        {
+            rMat.m_comp.remove(ent);
+        }
+
+        std::advance(first, 1);
     }
 }
 
--- a/src/osp/Active/SysRender.h
+++ b/src/osp/Active/SysRender.h
@@ -134,7 +134,7 @@ public:
     static MeshId own_mesh_resource(
             ACtxDrawing& rCtxDrawing,
             ACtxDrawingRes& rCtxDrawingRes,
-            Resources &rResources,
+            Resources& rResources,
             ResId resId);
 
     /**
@@ -152,7 +152,7 @@ public:
      */
     static void clear_resource_owners(
             ACtxDrawingRes& rCtxDrawingRes,
-            Resources &rResources);
+            Resources& rResources);
 
     inline static void add_draw_transforms_recurse(
             acomp_storage_t<ACompHierarchy> const& hier,
@@ -191,11 +191,11 @@ public:
 
     template<typename IT_T>
     static void update_delete_drawing(
-            ACtxDrawing &rCtxDraw, IT_T first, IT_T last);
+            ACtxDrawing& rCtxDraw, IT_T first, IT_T last);
 
     template<typename IT_T>
     static void update_delete_groups(
-            ACtxRenderGroups &rCtxGroups, IT_T first, IT_T last);
+            ACtxRenderGroups& rCtxGroups, IT_T first, IT_T last);
 
 private:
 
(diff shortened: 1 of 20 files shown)


src/osp/Active/drawing.h
```diff
@@ -99,25 +88,59 @@ struct MaterialData
...
+    IdMap_t<ResId, TexId>                   m_resToTex;
+    IdMap_t<TexId, ResIdOwner_t>            m_texToRes;
+
+    // Associate Mesh Ids with resources
+    IdMap_t<ResId, MeshId>                  m_resToMesh;
+    IdMap_t<MeshId, ResIdOwner_t>           m_meshToRes;
+};
+
+//
+
```
Capital-Asterisk, line 144: Not necessary, but I'm considering moving these to a new header. This will allow ACtxDrawing to be included without needing include resourcetypes.h.

src/osp/Active/SysRender.h
```diff
@@ -119,6 +121,39 @@ class SysRender
...
+     * @param rCtxDrawingRes    [ref] Resource drawing data
+     * @param rResources        [ref] Application Resources containing meshes
+     * @param resId             [in] Mesh Resource Id
+     *
+     * @return Id of new mesh, existing mesh Id if it already exists.
+     */
+    static MeshId own_mesh_resource(
+            ACtxDrawing& rCtxDrawing,
+            ACtxDrawingRes& rCtxDrawingRes,
+            Resources &rResources,
```
jonesmz, line 137: spacing around the&

src/osp/Active/SysRender.h
```diff
@@ -119,6 +121,39 @@ class SysRender
...
+
+    /**
+     * @brief Dissociate resources from the scene's meshes and textures
+     *
+     * @param rCtxDrawingRes    [ref] Resource drawing data
+     * @param rResources        [ref] Application Resources
+     */
+    static void clear_resource_owners(
+            ACtxDrawingRes& rCtxDrawingRes,
+            Resources &rResources);
```
jonesmz, line 155: spacing around the &

src/osp/Active/SysRender.h
```diff
@@ -202,19 +237,42 @@ void SysRender::assure_draw_transforms(
     }
 }
 
+template<typename STORAGE_T, typename REFCOUNT_T>
+void remove_refcounted(
+        ActiveEnt ent, STORAGE_T &rStorage, REFCOUNT_T &rRefcount)
+{
+    if (rStorage.contains(ent))
+    {
+        auto &rOwner = rStorage.get(ent);
```
jonesmz, line 246: doesn't this require two lookups?

src/osp/Active/SysRender.h
```diff
@@ -202,19 +237,42 @@ void SysRender::assure_draw_transforms(
...
+        // Textures and meshes are reference counted
+        remove_refcounted(ent, rCtxDraw.m_diffuseTex, rCtxDraw.m_texRefCounts);
+        remove_refcounted(ent, rCtxDraw.m_mesh, rCtxDraw.m_meshRefCounts);
+
+        for (MaterialData& rMat : rCtxDraw.m_materials)
+        {
+            rMat.m_comp.remove(ent);
+        }
+
+        std::advance(first, 1);
```
jonesmz, line 275: I think the second argument defaults to 1.

src/osp/Active/opengl/SysRenderGL.cpp
```diff
@@ -27,6 +27,10 @@
 
 #include "../../Shaders/FullscreenTriShader.h"
 
+#include "../../Resource/resources.h"
```
jonesmz, line 30: Food for thought for a later change.

Instead of using relative paths like this, we should have a folder for header files so that they can be included like

<osp/component/header.h>

which would allow, in the future, for the code of osp-magnum to be further decomposed into libraries that can be compiled against package-manager installed headers from the OS, without needing to do any extra work.

src/osp/Active/opengl/SysRenderGL.cpp
```diff
@@ -101,134 +105,112 @@ void SysRenderGL::setup_context(RenderGL& rCtxGl)
...
 }
 
-void SysRenderGL::compile_textures(
-        acomp_storage_t<ACompTexture> const& textures,
-        std::vector<ActiveEnt> const& dirty,
-        acomp_storage_t<TexGlId>& rTexGl,
+void SysRenderGL::assign_meshes(
+        acomp_storage_t<MeshIdOwner_t> const& cmpMeshIds,
+        IdMap_t<MeshId, ResIdOwner_t> const& meshToRes,
+        std::vector<ActiveEnt> const& entsDirty,
```
jonesmz, line 170: I think we can safely start using C++20, all the compilers should be caught up for most C++20 features.

So you could, if you wanted to, use std::span<> for this.

It's basically the sting-view of the non-char world.

src/osp/Resource/resourcetypes.h
```diff
@@ -93,19 +94,40 @@ struct res_container
...
 
-using ResIdStorage_t = lgrn::IdStorage<ResId, Resources>;
+using ResIdOwner_t = lgrn::IdStorage<ResId, Resources>;
+
+
+// Resource Type Ids
+
+struct GenResTypeId
+{
+    friend inline ResTypeId resource_type_next() noexcept;
```
jonesmz, line 108: What's up with "friend inline" here?

jonesmz: Oh. You're making these global, instead of static member functions.

src/osp/Resource/resourcetypes.h
```diff
@@ -93,19 +94,40 @@ struct res_container
...
+    friend inline int resource_type_count() noexcept;
+private:
+    static inline int count{0};
+};
+
+inline ResTypeId resource_type_next() noexcept
+{
+    return ResTypeId(GenResTypeId::count++);
```
jonesmz, line 116: starting at zero is desired?

Means you can't detect if you've overflowed and rolled over.

src/osp/Resource/resourcetypes.h
```diff
@@ -93,19 +94,40 @@ struct res_container
...
+
+
+// Resource Type Ids
+
+struct GenResTypeId
+{
+    friend inline ResTypeId resource_type_next() noexcept;
+    friend inline int resource_type_count() noexcept;
+private:
+    static inline int count{0};
```
jonesmz, line 111: Is int really the right choice?

At the least, shouldn't it be ResTypeId ?

src/osp/Shaders/Flat.cpp
```diff
@@ -89,12 +86,12 @@ void shader::assign_flat(
             if (diffuse.contains(ent))
             {
                 pStorageOpaque->emplace(
-                        ent, EntityToDraw{&draw_ent_flat, {&rData, &(*rData.m_shaderDiffuse)} });
+                        ent, EntityToDraw{&draw_ent_flat, {&rData, &rData.m_shaderDiffuse} });
```
jonesmz, line 89: consider std::addressof() for more genericness.

src/osp/id_map.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace osp
+{
+
+// Preferred container to map int/enum Ids with other Ids or small datatypes
+template <typename KEY_T, typename VALUE_T>
+using IdMap_t = entt::dense_hash_map<KEY_T, VALUE_T>;
+
+// TODO: Consider writing a wrapper that specializes the map with the integer
+//       underlying type of KEY_T. This will these maps easier to manage, and
```
jonesmz, line 37: Grammar.

src/osp/id_map.h
```diff
@@ -0,0 +1,40 @@
...
+//       may improve compile times as only one specialization is actually made.
```
jonesmz, line 38: You'd want to use an extern template to force only one instantiation to ever happen.

src/test_application/activescenes/scenarios_enginetest.cpp
```diff
@@ -69,27 +69,47 @@ constexpr int const gc_maxMaterials = 2;
...
     // Components and supporting data structures
     osp::active::ACtxBasic          m_basic;    
     osp::active::ACtxDrawing        m_drawing;
+    osp::active::ACtxDrawingRes     m_drawingRes;
 
     // Hierarchy root, needs to exist so all hierarchy entities are connected
     osp::active::ActiveEnt          m_hierRoot;
 
     // The rotating cube
     osp::active::ActiveEnt          m_cube;
```
jonesmz, line 93: Code scanning
Code scanning

Variable 'testapp::enginetest::EngineTestScene::m_cube' is uninitialized. Always initialize a member variable (type.6).

Warning
Variable 'testapp::enginetest::EngineTestScene::m_cube' is uninitialized. Always initialize a member variable (type.6).

src/test_application/activescenes/scenarios_physicstest.cpp
```diff
@@ -108,9 +121,8 @@ struct PhysicsTestScene
     osp::active::ActiveEnt          m_hierRoot;
```
jonesmz, line 121: Code scanning

Variable 'testapp::physicstest::PhysicsTestScene::m_hierRoot' is uninitialized. Always initialize a member variable (type.6).

Warning
Variable 'testapp::physicstest::PhysicsTestScene::m_hierRoot' is uninitialized. Always initialize a member variable (type.6).

jonesmz (approved):

src/osp/Active/SysRender.h
Capital-Asterisk: This doesn't seem to be the case

src/osp/Active/SysRender.h
Capital-Asterisk: Assuming this looks similar to the stuff in the shaders? The shaders needed to go from ActiveEnt -> MeshGlId -> Mesh; here, it's just ActiveEnt -> Owner

src/osp/Active/opengl/SysRenderGL.cpp
Capital-Asterisk: This is already used in the other files that aren't in src/osp, they should work here too.

src/osp/Active/opengl/SysRenderGL.cpp
Capital-Asterisk: In this case, Corrade also provides ArrayView equivalent to std::span and is also stl-compatible.

It shouldn't take that much effort to edit the line in the CMakeLists, but switching to C++20 will also affect other people who might want to use the project in the future. Also noting the possibility of using that 'C++ as a scripting language' thing or wasm.

Though I have started using [[likely]] and [[unlikely]] in other projects.

src/osp/Resource/resourcetypes.h
Capital-Asterisk: Yes, zero should be the first value returned. It's incredibly unlikely this is going to go any higher than 20 or so; overflow isn't a problem.

src/osp/Resource/resourcetypes.h
Capital-Asterisk: Mentioned above, this number isn't going to go that high. It would still need to be casted to an integer type to increment. I could use the underlying type, but that excess complication.

src/osp/Shaders/Flat.cpp
Capital-Asterisk: Will this function benefit from the genericness? This just seems like extra clutter that makes it more difficult to read in this case.

src/osp/Active/SysRender.h
jonesmz: Ugh. Sorry. I got it confused with std::next.
https://en.cppreference.com/w/cpp/iterator/next

src/osp/Active/opengl/SysRenderGL.cpp
jonesmz: the cling interpreter should be able to be built as a shared library for whatever version of c++ it needs, and then it would consume only the public headers of OSP.

If it doesn't support C++20, we can restrict the public headers to C++17, but it's based on clang, so i would hope it's not *too* terribly out of date...

ArrayView is fine, if you'd rather use that.

PR #183, opened 2022-03-25
Add Common Test Scene

Capital-Asterisk:
Common test scenes aim to reduce the absurd amount of duplicate code that would have been necessary when creating new test scenes.

The notable classes added are CommonTestScene and CommonSceneRendererGL, mostly consisting of code moved out of `scenarios_physicstest.cpp`.

`scenarios_enginetest.cpp` is designed to be self contained, so it does not use the new common test scene.

### CommonTestScene

CommonTestScene contains common data that every test scene will probably use: IdRegistry for entities, basic components such as hierarchy, drawing, etc.. Taking a composition approach, additional arbitrary data can be added using `emplace<T>()` and `get<T>()`. Besides cleanup functions, CommonTestScene is mostly pure data, and does not describe any procedures for updating.

### CommonSceneRendererGL

CommonSceneRendererGL contains common data needed for forward rendering a single scene. It comes with material IDs, shaders, and features the same composition approach as CommonTestScene. It has a few important member functions, but it's also mostly just data.

### Draw function and test scene 'class'

The primary way to draw and update the world, is to inject an `on_draw_t` std::function into ActiveApplication. The std::function stores renderer data using lambda capture.

The enginetest has `testapp::enginetest::generate_draw_func` for creating an `on_draw_t`, storing an EngineTestRenderer into it.

Common test scenes have `testapp::generate_common_draw` in `common_renderer_gl.cpp` to create an `on_draw_t`. Custom renderer setup code can be injected as a parameter.

CommonTestScene does not depend on CommonSceneRendererGL. ActiveApplication is made for Magnum and OpenGL. It shouldn't be too difficult to make a text mode renderer, vulkan renderer, RTX raytracer, or whatever.

### DIFF
--- a/src/test_application/activescenes/common_renderer_gl.cpp
+++ b/src/test_application/activescenes/common_renderer_gl.cpp
@@ -31,7 +31,7 @@
 namespace testapp
 {
 
-void CommonSceneRendererGL::setup(ActiveApplication& rApp, CommonTestScene& rScene)
+void CommonSceneRendererGL::setup(ActiveApplication& rApp, CommonTestScene const& rScene)
 {
     using namespace osp::active;
     using namespace osp::shader;
@@ -56,13 +56,10 @@ void CommonSceneRendererGL::setup(ActiveApplication& rApp, CommonTestScene& rSce
 
 }
 
-void CommonSceneRendererGL::render(ActiveApplication& rApp, CommonTestScene& rScene)
+void CommonSceneRendererGL::sync(ActiveApplication& rApp, CommonTestScene const& rScene)
 {
     using namespace osp::active;
     using namespace osp::shader;
-    using Magnum::GL::Framebuffer;
-    using Magnum::GL::FramebufferClear;
-    using Magnum::GL::Texture2D;
 
     RenderGroup &rGroupFwdOpaque
             = m_renderGroups.m_groups["fwd_opaque"];
@@ -115,6 +112,14 @@ void CommonSceneRendererGL::render(ActiveApplication& rApp, CommonTestScene& rSc
             rScene.m_basic.m_hierarchy,
             rScene.m_basic.m_transform,
             m_renderGl.m_drawTransform);
+}
+
+void CommonSceneRendererGL::render(ActiveApplication& rApp, CommonTestScene const& rScene)
+{
+    using namespace osp::active;
+    using Magnum::GL::Framebuffer;
+    using Magnum::GL::FramebufferClear;
+    using Magnum::GL::Texture2D;
 
     // Get camera to calculate view and projection matrix
     ACompCamera const &rCamera = rScene.m_basic.m_camera.get(m_camera);
@@ -149,5 +154,35 @@ void CommonSceneRendererGL::update_delete(std::vector<osp::active::ActiveEnt> co
     osp::active::SysRender::update_delete_groups(m_renderGroups, first, last);
     osp::active::SysRenderGL::update_delete(m_renderGl, first, last);
 }
+
+on_draw_t generate_common_draw(CommonTestScene& rScene, ActiveApplication& rApp, setup_renderer_t setup_scene)
+{
+    std::shared_ptr<CommonSceneRendererGL> pRenderer
+            = std::make_shared<CommonSceneRendererGL>();
+
+    // Setup default resources
+    pRenderer->setup(rApp, rScene);
+
+    // Setup scene-specifc stuff
+    setup_scene(*pRenderer, rScene, rApp);
+
+    // Set all drawing stuff dirty then sync with renderer.
+    // This allows clean re-openning of the scene
+    osp::active::SysRender::set_dirty_all(rScene.m_drawing);
+    pRenderer->sync(rApp, rScene);
+
+    return [&rScene, pRenderer = std::move(pRenderer)] (ActiveApplication& rApp, float delta)
+    {
+        pRenderer->m_onCustomDraw(*pRenderer, rScene, rApp, delta);
+
+        // Delete components of deleted entities on renderer's side
+        pRenderer->update_delete(rScene.m_deleteTotal);
+
+        pRenderer->sync(rApp, rScene);
+
+        // Render to screen
+        pRenderer->render(rApp, rScene);
+    };
+};
 
 } // namespace testapp
--- a/src/test_application/activescenes/common_renderer_gl.cpp
+++ b/src/test_application/activescenes/common_renderer_gl.cpp
@@ -31,7 +31,7 @@
 namespace testapp
 {
 
-void CommonRendererGL::setup(ActiveApplication& rApp, CommonTestScene& rScene)
+void CommonSceneRendererGL::setup(ActiveApplication& rApp, CommonTestScene& rScene)
 {
     using namespace osp::active;
     using namespace osp::shader;
@@ -56,7 +56,7 @@ void CommonRendererGL::setup(ActiveApplication& rApp, CommonTestScene& rScene)
 
 }
 
-void CommonRendererGL::render(ActiveApplication& rApp, CommonTestScene& rScene)
+void CommonSceneRendererGL::render(ActiveApplication& rApp, CommonTestScene& rScene)
 {
     using namespace osp::active;
     using namespace osp::shader;
@@ -142,7 +142,7 @@ void CommonRendererGL::render(ActiveApplication& rApp, CommonTestScene& rScene)
     SysRenderGL::display_texture(rApp.get_render_gl(), rFboColor);
 }
 
-void CommonRendererGL::update_delete(std::vector<osp::active::ActiveEnt> const& toDelete)
+void CommonSceneRendererGL::update_delete(std::vector<osp::active::ActiveEnt> const& toDelete)
 {
     auto first = std::cbegin(toDelete);
     auto last = std::cend(toDelete);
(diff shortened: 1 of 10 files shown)


src/test_application/activescenes/scenarios_physicstest.cpp
```diff
@@ -207,40 +186,60 @@ ActiveEnt add_quick_shape(
...
 
-    entt::any sceneAny = entt::make_any<PhysicsTestScene>();
-    PhysicsTestScene &rScene = entt::any_cast<PhysicsTestScene&>(sceneAny);
+    // It should be clear that the physics test scene is a composition of:
+    // * PhysicsData:       Generic physics data
+    // * ACtxNwtWorld:      Newton Dynamics Physics engine
+    // * PhysicsTestData:   Additional scene-specific data, ie. dropping blocks
+    auto &rScnPhys  = rScene.emplace<PhysicsData>();
+    auto &rScnNwt   = rScene.emplace<ospnewton::ACtxNwtWorld>(gc_threadCount);
+    auto &rScnTest  = rScene.emplace<PhysicsTestData>();
```
Capital-Asterisk, line 224: I'd point out this part in particular for what I mean by composition.

src/test_application/activescenes/common_renderer_gl.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <Magnum/GL/DefaultFramebuffer.h>
+
+#include <Corrade/Containers/ArrayViewStl.h>
+
+namespace testapp
+{
+
+void CommonSceneRendererGL::setup(ActiveApplication& rApp, CommonTestScene const& rScene)
+{
+    using namespace osp::active;
```
jonesmz, line 36: You might as well just do this at the top of the file.

src/test_application/activescenes/common_renderer_gl.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto first = std::cbegin(toDelete);
+    auto last = std::cend(toDelete);
+    osp::active::SysRender::update_delete_groups(m_renderGroups, first, last);
+    osp::active::SysRenderGL::update_delete(m_renderGl, first, last);
+}
+
+on_draw_t generate_common_draw(CommonTestScene& rScene, ActiveApplication& rApp, setup_renderer_t setup_scene)
+{
+    std::shared_ptr<CommonSceneRendererGL> pRenderer
+            = std::make_shared<CommonSceneRendererGL>();
```
jonesmz, line 161: auto pRenderer = std::make_shared<CommonSceneRendererGL>();

src/test_application/activescenes/common_renderer_gl.h
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @brief Delete components of entities to delete
+     *
+     * @param toDelete  [in] Vector of deleted entities
+     */
+    void update_delete(std::vector<osp::active::ActiveEnt> const& toDelete);
+};
+
+
+using setup_renderer_t = void(*)(CommonSceneRendererGL&, CommonTestScene&, ActiveApplication&);
```
jonesmz, line 98: noexcept

src/test_application/activescenes/common_renderer_gl.h
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * @brief Common data needed to render a scene
+ *
+ * Note: GPU resources and application-level rendering data can be found in
+ *       osp::active::RenderGL
+ *
+ */
+struct CommonSceneRendererGL : MultiAny
+{
+    using on_custom_draw_t = void(*)(CommonSceneRendererGL&, CommonTestScene&,
```
jonesmz, line 48: noexcept

src/test_application/activescenes/common_scene.cpp
```diff
@@ -0,0 +1,83 @@
...
+            m_basic.m_hierarchy,
+            std::cbegin(m_delete), std::cend(m_delete));
+
+    // Create a new delete vector that includes the descendents of our current
+    // vector of entities to delete
+    m_deleteTotal = m_delete;
+    SysHierarchy::update_delete_descendents(
+            m_basic.m_hierarchy,
+            std::cbegin(m_delete), std::cend(m_delete),
+            [=] (ActiveEnt ent)
```
jonesmz, line 60: [=] -> [this]

jonesmz: alternatively, you can do

`std::back_inserter(m_deleteTotal)

src/test_application/activescenes/common_scene.cpp
```diff
@@ -0,0 +1,83 @@
...
+    auto last = std::cend(m_deleteTotal);
+    osp::active::update_delete_basic                (m_basic,   first, last);
+    osp::active::SysRender::update_delete_drawing   (m_drawing, first, last);
+
+    // Delete entity IDs
+    for (ActiveEnt const ent : m_deleteTotal)
+    {
+        if (m_activeIds.exists(ent))
+        {
+            m_activeIds.remove(ent);
```
jonesmz, line 78: doesnt .remove() have to do the same search logic as .exists() ?

shouldn't you just call remove() and let it be a no-op if the ent doesn't exist?

src/test_application/activescenes/common_scene.h
```diff
@@ -0,0 +1,140 @@
...
+{
+    Array_t::iterator found = std::find_if(std::begin(m_data), std::end(m_data),
+                              [] (entt::any const &any)
+    {
+        return ! bool(any);
+    });
+
+    assert (found != std::end(m_data));
+
+    found->emplace<T>(std::forward<ARGS_T>(args) ...);
```
jonesmz, line 64: You probably want std::in_place, if entt::any supports it.

src/test_application/activescenes/common_scene.h
```diff
@@ -0,0 +1,140 @@
...
+
+    return entt::any_cast<T&>(*found);
```
jonesmz, line 66: I feel like you would want to mimic the api of std::set or std::map, with this function returning a std::pair<iterator, bool>

src/test_application/activescenes/common_scene.h
```diff
@@ -0,0 +1,140 @@
...
+}
+
+template <typename T>
+T& MultiAny::get()
```
jonesmz, line 70: and if you have two of the same type?

src/test_application/activescenes/scenarios_physicstest.cpp
```diff
@@ -207,40 +186,60 @@ ActiveEnt add_quick_shape(
...
+
+    // Add cleanup function to deal with reference counts
+    // Note: The problem with destructors, is that REQUIRE storing pointers in
```
jonesmz, line 227: ??

destructors for std::unique_ptr ?

if you make the destructor a stateless functor, you don't need to store anything extra.
```cpp
struct Bob
{
    void operator()(whatever)
    {
    }
}
std::unique_ptr<whatever, Bob> ...
```

jonesmz (approved):

src/test_application/activescenes/common_scene.h
```diff
@@ -0,0 +1,140 @@
...
+    T& emplace(ARGS_T&& ... args);
+
+    template <typename T>
+    T& get();
+};
+
+template <typename T, typename ... ARGS_T>
+T& MultiAny::emplace(ARGS_T&& ... args)
+{
+    Array_t::iterator found = std::find_if(std::begin(m_data), std::end(m_data),
```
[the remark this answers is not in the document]
Capital-Asterisk: Here, it's looking for an empty spot in the array; will add a comment for this.

src/test_application/activescenes/common_scene.h
Capital-Asterisk: Won't that only be needed for the any's constructor? This is a dedicated emplace function.

src/test_application/activescenes/common_scene.h
Capital-Asterisk: Hmm, it should also detect if the type already exists, and return false there.

src/test_application/activescenes/common_scene.h
Capital-Asterisk: As implied above, it's only intended for unique types.

src/test_application/activescenes/scenarios_physicstest.cpp
Capital-Asterisk: The problem here is that PhysicsTestData needs a way to access the CommonTestScene it is stored inside when destructed. I'm not sure how a unique ptr can be used to fix this.

src/test_application/activescenes/common_scene.cpp
Capital-Asterisk: IdRegistry::remove throws an exception because I didn't know any better. 

Removing an invalid ID should be considered an error most of the time; the user should never be keeping them around in the first place. I might have two versions of this function instead.

The only reason why a check is needed here, is because an entity can be deleted twice, which adds it twice to the m_deleteTotal vector. If m_deleteTotal is changed to a hierarchical bitset, then this wouldn't be possible

jonesmz: If it is considered a "you did something really stupid", then an assert might be better than an exception.

But I suppose that doesn't really answer my concern about the double search.

What architectural challenges would you have to introduce the hierarchical bitset?

Capital-Asterisk: Double search? It's just accessing a bitset, no searching involved.

"Architectural challenges" would mostly involve adding a function to allocate the bitset, as well as changing a push_back(...) to a set(...).

PR #184, opened 2022-03-25
Add new glTF Loader

Capital-Asterisk:
This adds a new general purpose glTF loader utilizing the new resource system. It's far less janky, and isn't specifically made for in-game parts compared to the old one. As of now, it's able to load all the existing gltf files without crashing 

The main loading code simply uses Magnum TinyGltfLoader to populate a "ImporterData" struct. ImporterData is independent of whatever importer was used, and is associated with a Resource Id.

In the future, code that loads OSP-specific data can simply read the ImporterData, and add new types to the Resource Id.

### DIFF
--- a/src/osp/Resource/load_tinygltf.cpp
+++ b/src/osp/Resource/load_tinygltf.cpp
@@ -1,0 +1,195 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2022 Open Space Program Project
+ *
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "load_tinygltf.h"
+
+#include "ImporterData.h"
+#include "resources.h"
+
+#include "../logging.h"
+#include "../string_concat.h"
+
+#include <MagnumPlugins/TinyGltfImporter/TinyGltfImporter.h>
+#include <MagnumExternal/TinyGltf/tiny_gltf.h>
+
+#include <Magnum/Trade/ImageData.h>
+#include <Magnum/Trade/TextureData.h>
+#include <Magnum/Trade/MeshData.h>
+#include <Magnum/Trade/ObjectData3D.h>
+
+#include <Corrade/PluginManager/Manager.h>
+
+using namespace osp;
+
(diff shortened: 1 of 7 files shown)


src/osp/Resource/load_tinygltf.cpp
```diff
@@ -0,0 +1,284 @@
...
+        {
+            string_append(tempStr, name, ":", resName);
+        }
+        else
+        {
+            // i don't like std::to_string but screw it, fix it later
+            string_append(tempStr, name, ":unnamed-", std::to_string(id));
+        }
+
+        return tempStr;
```
jonesmz, line 96: I don't get why you are storing a "tempStr" and then .clear()ing it every time.

```cpp
if ( ! resName.empty())
        {
            return string_concat(name, ":", resName);
        }
        else
        {
            // i don't like std::to_string but screw it, fix it later
            return string_concat(name, ":unnamed-", std::to_string(id));
        }
```

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -0,0 +1,284 @@
...
+            (std::string_view resName, UnsignedInt id) mutable -> std::string const&
+    {
+        tempStr.clear();
+        if ( ! resName.empty())
+        {
+            string_append(tempStr, name, ":", resName);
+        }
+        else
+        {
+            // i don't like std::to_string but screw it, fix it later
```
jonesmz, line 92: Same.

But I haven't gotten around to attempting to implement the in-place conversion function described in this paper: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1228r1.html

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -0,0 +1,284 @@
...
+        Pointer<ObjectData3D const> obj         = rImporter.object3D(i);
+        int const instance                      = obj->instance();
+        ObjectInstanceType3D const instanceType = obj->instanceType();
+
+        if ( obj == nullptr )
+        {
+            continue;
+        }
+
+        rImportData.m_objNames[i] = rImporter.object3DName(i);
```
jonesmz, line 198: Code scanning

The pointer is dangling because it points at a temporary instance which was destroyed.

Warning
The pointer is dangling because it points at a temporary instance which was destroyed.
Show more details

Check warning on line 198 in src/osp/Resource/load_tinygltf.cpp

Code scanning

gsl::span or std::string_view created from a temporary will be invalid when the temporary is invalidated (gsl.view).

Warning
gsl::span or std::string_view created from a temporary will be invalid when the temporary is invalidated (gsl.view).
Show more details

jonesmz (approved):

src/osp/Resource/load_tinygltf.cpp
Capital-Asterisk: Its a minor optimization to potentially avoid allocating a string each time. Im not sure if the standard guarantees it, but clear should just set the length to zero.

src/osp/Resource/load_tinygltf.cpp
Capital-Asterisk: yep, this is a catastrophic error. Originally intended to keep the importer around, but that isn't the case anymore.

Capital-Asterisk:
Don't merge this. `load_tinygltf.cpp` uses lots of deprecated Magnum features that will eventually stop working. I'd suggest updating Magnum to a certain commit

src/osp/Resource/load_tinygltf.cpp
jonesmz: you're inhibiting return value optimization.

I'm not sure the benefits are worth the costs

src/test_application/main.cpp
```diff
@@ -389,6 +410,63 @@ void load_a_bunch_of_stuff()
     OSP_LOG_INFO("Resource loading complete");
 }
 
+static void resource_for_each_type(osp::ResTypeId type, void(*do_thing)(osp::ResId))
```
jonesmz, line 413: I recommend making a typedef for your pointer-to-function type.

src/test_application/main.cpp
```diff
@@ -389,6 +410,63 @@ void load_a_bunch_of_stuff()
...
+
+    // Texture resources contain osp::TextureImgSource, which refererence counts
+    // their associated image data
+    resource_for_each_type(gc_texture, [] (osp::ResId id)
+    {
+        auto *pData = g_resources
+                .data_try_get<osp::TextureImgSource>(gc_texture, id);
+        if (pData == nullptr)
+        {
+            return;
```
jonesmz, line 437: minor: I recommend doing this in reverse.

if(pData != nullptr)
{
    call the function
}
// else nothing

src/osp/Resource/load_tinygltf.cpp
Capital-Asterisk: The return value is only being passed to `Resources::create` which only accepts a string view and allocates internally.

PR #186, opened 2022-03-29
Fix and clean up some build system stuff

Capital-Asterisk:
Addressing various issues:
* Linux workflow artifacts don't contain executables, and overwrite each other
* Newton dynamics shared libraries not building in the correct directory on unix
* Inconsistent build directories between operating systems: ie. `bin/Debug` and `Debug/bin`
* Confusing CMake comments, likely referring to lines that have entirely changed

The biggest change here is that executables are now built in `build/Debug` or `build/Release`. Staying with the windows crap fix, libraries are built into the same directory as the executables. On Unix, libraries are built in a `lib` subdirectory.


src/CMakeLists.txt
```diff
@@ -96,46 +96,45 @@ target_link_libraries(osp-magnum-deps PUBLIC
...
-
-# Enforce conformance mode for osp-magnum
-target_compile_options(osp-magnum PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
-
 # Segregate headers, shaders into filters
 source_group("Shader Files" FILES ${SHADER_FILES})
 source_group("Header Files" FILES ${H_FILES})
 
-# Put executable in the bin folder
-set_target_properties(osp-magnum PROPERTIES EXPORT_COMPILE_COMMANDS TRUE)
```
jonesmz, line 135: The comment here is wrong, but the EXPORT_COMPILE_COMMANDS property still needs to be set to true in order to get include what you use and clang tidy to work.

jonesmz: Oh. Nevermind, you set it below.

CMakeLists.txt
```diff
@@ -119,11 +122,17 @@ ELSE()
...
 IF(MSVC)
-  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/$<CONFIG> CACHE PATH "")
-  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/$<CONFIG> CACHE PATH "")
+  # Put shared libraries in the same directory as the executable on windows
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIG> CACHE PATH "")
+  # Force multiprocessor compilation for all projects
   add_compile_options($<$<CXX_COMPILER_ID:MSVC>:/MP>)
+ELSEIF(UNIX)
+  # Put shared libraries in ./lib, and use RPATH
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIG>/lib CACHE PATH "")
```
jonesmz, line 134: i'm skeptical about this.
Why bother on linux?

CMakeLists.txt
```diff
@@ -73,6 +73,9 @@ IF(OSP_ENABLE_COMPILER_WARNINGS)
   STRING(APPEND GCC_COMPILE_FLAGS " -Wno-psabi")
 ENDIF() # OSP_ENABLE_COMPILER_EARNINGS
 
+# Set OSP as default startup project in Visual Studio
+set_property(DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT OSP-MAGNUM)
```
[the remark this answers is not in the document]
jonesmz: You don't *HAVE* to set it in the top level. They're just saying that it's an example of where you might find the `project` function called.

CMakeLists.txt
Capital-Asterisk: I'm thinking it would just be cleaner to not have the directory cluttered with shared libraries.

PR #191, opened 2022-04-29
Add Vehicle Test Scene, Prefabs, Parts, and Vehicle Builder

Capital-Asterisk:
## Prefabs

The new glTF loader added in #184 doesn't care about any OSP-specific data, and just loads plain glTF. The new `assigns_prefabs_tinygltf()` function in `load_tinygltf.cpp` will read loaded glTFs to assign 'prefabs' as well as colliders and mass.

Prefabs simply describe an instantiable hierarchy of entities, a term used in many other game engines. This is different from a 'Part', which is described as "a more complex physical 'thing' in a scene, such as a rocket engine, fuel tank, or a capsule." Parts consist of a prefab, but prefabs are general-purpose and don't have to be parts.

Prefabs can be addressed with the integer Resource ID of the gltf data (importer data) and another integer ID for the prefab index. eg. PrefabPair stored in resourcetypes.h:
```cpp
struct PrefabPair
{
    ResIdOwner_t m_importer;
    int m_prefabId;
};
```
Because of the flexible resource system, importers and its prefabs can be easily assigned additional data in the future.

Carrying over from the existing '.sturdy.gltf' scheme, any glTF node with a name prefixed with "part_" will be considered a 'prefab'. However, machine data stored in the .sturdy.gltf files ([example](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/edc5dec6d906d8912546b87cb326930b4e2ddd56/bin/OSPData/adera/ph_rcs.sturdy.gltf#L56-L76)) **will be ignored (forever).**

It simply isn't practical to store machine data in the glTF file. Needing to modify a model file just to change some in-game properties is a pain. This is even worse when considering how to make identical parts that have different stats.

Machines are not yet implemented, but is instead planned to be stored inside of vehicle data, not in parts (if storing parts anywhere is even practical in the first place). A library of parts a player can select from and build spacecraft is something that should really be handled at the game-level, not at this early engine stage.

About parts, there's a bit of an explanation in `parts.h`

## Prefab Initialization

Prefabs can be initialized in a parallel fashion. see `SysPrefabInit.h` and `scenarios_vehicletest.cpp`.

The vehicle test scene keeps a vector of which prefabs to initialize on the current frame. All at once, it creates a bunch of entity IDs, then calls system functions to add components to those entities.

## Vehicle Builder

Interface shouldn't be too difficult to understand:
```cpp
    VehicleBuilder &rBuilder = rScnTest.m_builder.emplace(rScene.m_pResources);

    auto const [ capsule, fueltank, engine ] = rBuilder.create_parts<3>();
    rBuilder.set_prefabs({
        { capsule,  "phCapsule" },
        { fueltank, "phFuselage" },
        { engine,   "phEngine" },
    });

    rBuilder.set_transform({
       { capsule,  Matrix4::translation({0, 0, 3}) },
       { fueltank, Matrix4::translation({0, 0, 0}) },
       { engine,   Matrix4::translation({0, 0, -3}) },
    });
```

Vehicle data is stored in a bunch of arrays. It takes a bit of manual work to copy the data into a scene but is mostly just remapping IDs and may be abstracted later.

Eventually, machines, wiring, and other data will be added here too.

## Vehicle Test Scene

it looks like this:
image

The scene is setup in a way that makes vehicles a single rigid body. I'm too lazy to add a deleter system for the vehicles; they can't be deleted yet.


Capital-Asterisk:
VehicleBuilder diff is screwed up, but is readable in https://github.com/TheOpenSpaceProgram/osp-magnum/pull/191/commits/39b1119e6b0ee95fc04faf074021201f6ed98fef

jonesmz:
> Carrying over from the existing '.sturdy.gltf' scheme, any glTF node with a name prefixed with "part_" will be considered a 'prefab'. However, machine data stored in the .sturdy.gltf files ([example](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/edc5dec6d906d8912546b87cb326930b4e2ddd56/bin/OSPData/adera/ph_rcs.sturdy.gltf#L56-L76)) will be ignored (forever).

Shouldn't we have a "prefab_" prefix for things that are prefabs, but not parts?

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -0,0 +1,201 @@
...
+                gc_importer, rPrefab.m_importerRes);
+        auto &rPrefabData = rResources.data_get<osp::Prefabs const>(
+                gc_importer, rPrefab.m_importerRes);
+
+        auto const objects = rPrefabData.m_prefabs[rPrefab.m_prefabId];
+        auto const parents = rPrefabData.m_prefabParents[rPrefab.m_prefabId];
+
+        for (int i = 0; i < objects.size(); ++i)
+        {
+            int const prefabParent = parents[i];
```
jonesmz, line 61: int, instead of ssize_t ?

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -0,0 +1,201 @@
...
+        acomp_storage_t<ACompHierarchy>&    rHier) noexcept
+{
+    for (TmpPrefabInit const& rPrefab : prefabInits)
+    {
+        auto &rImportData = rResources.data_get<osp::ImporterData const>(
+                gc_importer, rPrefab.m_importerRes);
+        auto &rPrefabData = rResources.data_get<osp::Prefabs const>(
+                gc_importer, rPrefab.m_importerRes);
+
+        auto const objects = rPrefabData.m_prefabs[rPrefab.m_prefabId];
```
jonesmz, line 56: you're copying this container onto the stack, but you only ever use it to access the .size().

Do `auto const& objects` instead?

Or even `size_t const numObjects = rPrefabData.m_prefabs[rPrefab.m_prefabId].size();`

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+        for (int i = 0; i < objects.size(); ++i)
```
jonesmz, line 114: imho, boost::iterator::zip, or (well, with C++23, anyway): https://en.cppreference.com/w/cpp/ranges/zip_view

We could add a dependency on rangesv3, if you want. it would have the same stuff.

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -0,0 +1,201 @@
...
+                rCtxPhys.m_shape.emplace(ent, shape);
+                rCtxPhys.m_solid.emplace(ent);
+                rPhysIn.m_colliderDirty.push_back(ent);
+            }
+
+            if (float mass = rPrefabData.m_objMass[objects[i]];
+               mass != 0.0f)
+            {
+                osp::Vector3 const inertia
+                        = osp::phys::collider_inertia_tensor(shape != EShape::None ? shape : EShape::Sphere, rImportData.m_objTransforms[objects[i]].scaling(), mass);
```
jonesmz, line 194: this line is pretty long. maybe break the trinary if into more lines?

src/osp/Active/SysPrefabInit.h
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * This allows separate systems (doing physics, hierarchy, drawables, ...) to
+ * work in arallel on initializing a prefab.
+ *
+ * Intended to be created and quickly destroyed once the prefab is created.
+ * This is often the span of a single frame.
+ */
+struct TmpPrefabInit
+{
+    osp::ResId m_importerRes;
```
jonesmz, line 51:
```
Variable 'osp::active::TmpPrefabInit::m_importerRes' is uninitialized. Always initialize a member variable (type.6).

Warning
Variable 'osp::active::TmpPrefabInit::m_importerRes' is uninitialized. Always initialize a member variable (type.6).
Show more details 
```

If you're going to initialize m_pTransform, then you should initialize all of them.

src/osp/Active/parts.h
```diff
@@ -0,0 +1,82 @@
...
+#include "../types.h"
+
+#include <longeron/id_management/registry.hpp>
+#include <longeron/containers/intarray_multimap.hpp>
+
+namespace osp::active
+{
+
+using PartEnt_t = uint32_t;
+using RigidGroup_t = uint32_t;
```
jonesmz, line 39: Not using the strong-typedef of the enum classes for these?

No objection, just asking.

src/osp/Resource/ImporterData.h
```diff
@@ -83,7 +84,12 @@ struct ImporterData
 struct Prefabs
 {
     // [prefab Id][object]
-    lgrn::IntArrayMultiMap<int, int>    m_prefabs;
+    lgrn::IntArrayMultiMap<int, int>    m_prefabs{};
+    lgrn::IntArrayMultiMap<int, int>    m_prefabParents{};
```
jonesmz, line 88: is the `{}` needed to get the initialization behavior you want?

The default constructor would be called either way, no?

src/osp/Resource/ImporterData.h
```diff
@@ -83,7 +84,12 @@ struct ImporterData
...
+    std::vector<std::string_view>       m_prefabNames;
```
jonesmz, line 89: storing string_views as member variables sounds like a dangling pointer waiting to happen.

Can you document in the code the lifetime of these?

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -350,3 +350,119 @@ ResId osp::load_tinygltf_file(std::string_view filepath, Resources &rResources,
     return res;
 }
 
+static phys::EShape shape_from_name(std::string_view name) noexcept
+{
+    using phys::EShape;
```
jonesmz, line 355: This doesn't appear to do much for this function

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -350,3 +350,119 @@ ResId osp::load_tinygltf_file(std::string_view filepath, Resources &rResources,
...
+    rPrefabs.m_prefabParents.data_reserve(objCount);
+    rPrefabs.m_prefabParents.ids_reserve(topLevelSpan.size());
+
+    rPrefabs.m_prefabNames  .reserve(topLevelSpan.size());
+
+    // OSP parts are specified as top-level gltf nodes on the first scene
+    // with a name that starts with "part_"
+    // these rules may change
+
+    std::vector<int> prefabObjs;
```
jonesmz, line 405: i suggest one or more of:

1. Typedef these ints
2. Use a fixed sized integer type
3. Use an integer type that's "well known" to change size based on architecture, like ssize_t (but not long, since that changes per OS, not per processor arch)

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -350,3 +350,119 @@ ResId osp::load_tinygltf_file(std::string_view filepath, Resources &rResources,
...
+    std::vector<int> prefabParents;
+    prefabObjs.reserve(objCount);
+    prefabParents.reserve(objCount);
+
+    auto const process_obj_recurse = [&prefabObjs, &prefabParents, &rPrefabs, &rNodeExtras = *pNodeExtras, &rImportData = *pImportData] (auto&& self, int obj, int parent) -> void
```
jonesmz, line 410: this line is really long, recommend breaking it into 2 lines. One for capture class, one for params.

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -350,3 +350,119 @@ ResId osp::load_tinygltf_file(std::string_view filepath, Resources &rResources,
...
+    {
+        auto const &name = rImportData.m_objNames[obj];
+        tinygltf::Value const &extras = rNodeExtras[obj];
+
+        if (extras.IsObject())
+        {
+            if (name.hasPrefix("col_"))
+            {
+                // is Collider
+                std::string_view const shapeName = extras.Get("shape").Get<std::string>();
```
jonesmz, line 420: No dangling pointer?

Might be better to do "auto const&" to bind the temporary as whatever the actual type is, to avoid the possibility of pointing to a temporary.

src/osp/Resource/load_tinygltf.cpp
```diff
@@ -350,3 +350,119 @@ ResId osp::load_tinygltf_file(std::string_view filepath, Resources &rResources,
...
+            continue;
+        }
+
+        rPrefabs.m_prefabNames.emplace_back(name.exceptPrefix("part_"));
+
+        // Read children and populate prefabObjs and prefabParents
+        process_obj_recurse(process_obj_recurse, obj, -1);
+
+        rPrefabs.m_prefabs.emplace(
+                prefabIdNext, std::begin(prefabObjs), std::end(prefabObjs));
```
jonesmz, line 460: std::move_iterator

src/osp/Resource/resources.cpp
```diff
@@ -71,6 +71,13 @@ ResId Resources::find(ResTypeId typeId, PkgId pkgId, std::string_view name) cons
     return findIt->second;
 }
 
+std::string_view Resources::name(ResTypeId typeId, ResId resId) const noexcept
+{
+    PerResType const &rPerResType = get_type(typeId);
+
+    return rPerResType.m_resNames[std::size_t(resId)];
```
jonesmz, line 78:
```


Check warning on line 78 in src/osp/Resource/resources.cpp

Code scanning

Do not slice (es.63).

Warning
Do not slice (es.63).
Show more details 
```

This function should return `decltype(rPerResType.m_resNames[std::size_t(resId)])`

src/osp/Resource/resources.h
```diff
@@ -97,6 +97,15 @@ class Resources
...
     ResId find(ResTypeId typeId, PkgId pkgId, std::string_view name) const noexcept;
 
+    /**
+     * @brief Get name of Resource Id
+     *
+     * @param typeId
+     * @param resId
+     * @return
+     */
+    std::string_view name(ResTypeId typeId, ResId resId) const noexcept;
```
jonesmz, line 107: nodiscard

src/test_application/VehicleBuilder.cpp
```diff
@@ -24,223 +24,81 @@
...
-    if (protoIndex == m_vehicle.m_prototypes.size())
+VehicleBuilder::~VehicleBuilder()
+{
+    // clear resource owners
+    for (auto & [_, value] : m_prefabs)
     {
-        // Add the unlisted prototype to the end
-        m_vehicle.m_prototypes.emplace_back(protoPart);
+        m_pResources->owner_destroy(gc_importer, std::move(value.m_importer));
     }
```
jonesmz, line 42:
```cpp
VehicleBuilder::~VehicleBuilder()
{
    // clear resource owners
    for ([[maybe_unused]] auto && [_, value] : std::move(m_prefabs))
    {
        m_pResources->owner_destroy(gc_importer, std::move(value.m_importer));
    }

    for (osp::PrefabPair && rPrefabPair : std::move(m_data.m_partPrefabs))
    {
        m_pResources->owner_destroy(gc_importer, std::move(rPrefabPair.m_importer));
    }
}
```

src/test_application/VehicleBuilder.cpp
```diff
@@ -24,223 +24,81 @@
...
 {
-    Vector3 oset1{0.0f};
-    Vector3 oset2{0.0f};
-
-    for (osp::PCompName const& name : attachTo.m_partName)
+    for (SetPrefab const& set : setPrefabs)
     {
-        if (name.m_name == attachToName)
+        auto const foundIt = m_prefabs.find(set.m_prefabName);
+        if (foundIt != std::end(m_prefabs))
```
jonesmz, line 55: pre-bind the result of std::end(m_prefabs) priot to the for-loop, unless you think you'll cause iterator invalidation.

src/test_application/VehicleBuilder.h
```diff
@@ -52,115 +64,62 @@ class VehicleBuilder
...
 
-    osp::BlueprintVehicle m_vehicle;
+    VehicleData m_data;
 
+    // put more attachment data here
 };
 
 
+template <std::size_t N>
+[[nodiscard]] std::array<PartId, N> VehicleBuilder::create_parts()
```
jonesmz, line 114: you only need to annotate the declaration with [[nodiscard]].

No need to put it on the definition.,

jonesmz:
> > Carrying over from the existing '.sturdy.gltf' scheme, any glTF node with a name prefixed with "part_" will be considered a 'prefab'. However, machine data stored in the .sturdy.gltf files ([example](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/edc5dec6d906d8912546b87cb326930b4e2ddd56/bin/OSPData/adera/ph_rcs.sturdy.gltf#L56-L76)) will be ignored (forever).
> 
> Shouldn't we have a "prefab_" prefix for things that are prefabs, but not parts?

I also think that using the prefix of a filename is a bit more awkward than instead having a suffix.

something like `.ospprefab` `.osppart` would work, and they can simply be expected to be gltf.

jonesmz (approved):

src/osp/Active/SysPrefabInit.cpp
Capital-Asterisk: These are given as Magnum::Int and use '-1' as a null value, will typedef these somewhere.

src/osp/Active/SysPrefabInit.cpp
Capital-Asterisk: I like the syntax of zip and ranges, but I'm quite skeptical over how helpful they will really be to this project. There isn't an immediate problem that needs to be solved; I heard the compile time isn't great, and it may make it more difficult to get new members onto the team (the code is already pretty weird but still).

src/osp/Active/parts.h
Capital-Asterisk: I figured it isn't that helpful due to the excessive std::size_t(...) casts required. They're only useful when used as function parameters, and so far I don't have a good solution to generate a function that accepts an enum class just to access a single container. Prefixing the variable names with what type of index they accept decently helpful.

src/osp/Resource/ImporterData.h
Capital-Asterisk: I think I put it there to try to fix an error, but the error wasn't actually there?

src/osp/Resource/load_tinygltf.cpp
Capital-Asterisk: There's suppose to be more shapes added to this function; cube and cylinder are the only ones that exist in the gltf files so far.

src/osp/Resource/load_tinygltf.cpp
Capital-Asterisk: I don't believe this is any helpful when dealing with integer types?

src/osp/Resource/resources.cpp
Capital-Asterisk: So this is osp::SharedString being casted to osp::string_view; apparently that brings up this warning?

Capital-Asterisk:
> Shouldn't we have a "prefab_" prefix for things that are prefabs, but not parts?

Perhaps? I mostly didn't want to modify the existing part files that we currently have. There wouldn't really be a difference between parts and prefabs on the gltf-level.

> I also think that using the prefix of a filename is a bit more awkward than instead having a suffix.
> 
> something like `.ospprefab` `.osppart` would work, and they can simply be expected to be gltf.

These name prefixes are the names of the nodes in the gltf files. File names are not important. A single gltf file can contain multiple parts/prefabs. I'd highly prefer to still keep the .gltf extension, so it can be recognised by editors and model viewers.

src/osp/Resource/load_tinygltf.cpp
jonesmz: ah, yea, it wouldn't be.

src/osp/Resource/resources.cpp
jonesmz: Imho you should preserve the type so that consumers of the funciton's return'ed value can choose wether they want to treat it as a string view, or treat it as a shared string

Capital-Asterisk:
Forgot I also needed to update the Longeron++ since I fixed a major bug affecting IdRegistry

PR #196, opened 2022-05-07
Update to the v4.01 tag of newton-dynamics, but continue using the 3.…

jonesmz:
…14 folder instead of the 4.0 folder


jonesmz:
Contrary to the title, this PR doesn't actually bring us to tag 4.01, because I had to submit a PR to fix a mistake for the Mac OS X os.

That being said, i'm pretty sure that newton dynamics isn't really making any changes to the 3.14 subfolder, they just... keep it around in the branch for no reason at all?

So while this brings us to the master branch, unless we also switch to the 4.0 series of code, it's essentially just the 3.14 release series.

Capital-Asterisk (approved):
Can confirm that nothing broke and the build artifacts are working. 4.00 is almost a complete rewrite, and many things have changed; I don't expect switching to it any time soon. Again, noting using PhysX instead, an even better option now that z-adams has some experience working with it.

PR #197, opened 2022-05-09
Use CMakeLists.txt to ensure a working windows kit, instead of the CI system's configuration.

jonesmz:


jonesmz:
This helps with the case where someone tries to build the code on windows without explicitly setting the SDK version to one that will work.

It's not perfect, because CMake has some missing gaps here, but it's better to have some kind of warning about it than nothing.

We might need to add some additional logic to attempt to account for differing versions of the sdk, but for now this at least works on github.

PR #203, opened 2022-05-27
Implement rotations for CoordTransformer

Capital-Asterisk:
Planets rotate, the Active Area rotates, and transforming coordinates between them will lead to a rapid decline in sanity when taking the wrong approach. Luckily, there is now CoordTransformer.

CoordTransformer is useful for almost every operation involving coordinate spaces. eg:
* Transferring satellites between coordinate spaces
* Transforming positions of satellites to local-space for rendering

The implementation is quite algebra-heavy, and may take a while to fully understand. I find it quite mathematically elegant how some parts 'just work' or cancel out. I put heavy considerations in respecting integer range and floating point precision.

I did not intend to put this on the main repo, but oh well.

### DIFF
--- a/src/osp/universe/coordinates.h
+++ b/src/osp/universe/coordinates.h
@@ -31,6 +31,18 @@
 namespace osp::universe
 {
 
+inline Vector3g rotate_vector3g(Vector3g const in, Quaterniond const rot) noexcept
+{
+    return Vector3g(rot.transformVector(Vector3d(in)));
+}
+
+constexpr bool quat_non_zero(Quaterniond const in) noexcept
+{
+    // scalar = cos(angle / 2); no angle means scalar = 1
+    return in.scalar() != 1.0;
+}
+
+
 /**
  * @brief Describes a mathematical function used to transform positions between
  *        coordinate spaces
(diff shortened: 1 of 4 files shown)


src/osp/universe/coordinates.h
```diff
@@ -31,6 +31,18 @@
 namespace osp::universe
 {
 
+inline Vector3g rotate_vector3g(Vector3g const in, Quaterniond const rot) noexcept
+{
+    return Vector3g(rot.transformVector(Vector3d(in)));
```
jonesmz, line 36: i take it these can't be constexpr because Vector3g and Quaterniond aren't able ot be?

src/osp/universe/universe.h
```diff
@@ -38,26 +38,96 @@
...
-    std::array<typename View_t::Size, DIMENSIONS_T> m_offsets;
+    std::array<std::size_t, DIMENSIONS_T> m_offsets;
 
     // Assuming xyz all have the same stride
-    typename View_t::Stride m_stride;
+    std::ptrdiff_t m_stride;
+
+    constexpr View_t view(Data_t data, std::size_t count, int dimension) const
+    {
+        using namespace Corrade::Containers;
```
jonesmz, line 68: is this using namespace actually used?

src/osp/universe/universe.h
```diff
@@ -38,26 +38,96 @@
...
 
     // Offsets of xyz components in bytes
-    std::array<typename View_t::Size, DIMENSIONS_T> m_offsets;
+    std::array<std::size_t, DIMENSIONS_T> m_offsets;
 
     // Assuming xyz all have the same stride
-    typename View_t::Stride m_stride;
+    std::ptrdiff_t m_stride;
+
+    constexpr View_t view(Data_t data, std::size_t count, int dimension) const
```
jonesmz, line 66: should dimension be int, or size_t ?

jonesmz (approved):

src/osp/universe/coordinates.h
Capital-Asterisk: transformVector specifically; constructors and accessors work in constexpr

src/osp/universe/universe.h
Capital-Asterisk: nope, likely had something else more complex here but removed it.

src/osp/universe/universe.h
Capital-Asterisk: There would only be about 2, 3, or 4 dimensions. maybe 16 for a 4x4 matrix? An int is more than enough.

jonesmz: it's not about whether it has the capacity, it's about consistency in types.

"int" implies negative numbers are allowed.

size_t is both "number of objects in memory", as well as the "index into container starting from 0"

PR #204, opened 2022-05-28
Add Universe Test Scene

Capital-Asterisk:
Universe Test Scene features a self-contained universe consisting of tiny planets unrealistically orbiting a black hole. It is mainly intended to demonstrate coordinate spaces.

This also refactors CommonTestScene a bit to better support custom rendering functions.

### DIFF
--- a/src/osp/universe/universe.h
+++ b/src/osp/universe/universe.h
@@ -125,6 +125,7 @@ struct CoSpaceTransform
     // Ignore and use m_parentSat's position and rotation instead if non-null
     Quaterniond     m_rotation;
     Vector3g        m_position;
+
     int             m_precision{10}; // 1 meter = 2^m_precision
 };
 
--- a/src/osp/universe/universe.h
+++ b/src/osp/universe/universe.h
@@ -84,6 +84,13 @@ constexpr VEC_T to_vec(INDEX_T i, RANGE_T&& ... args) noexcept
     return {args[std::size_t(i)] ...};
 }
 
+
+struct CoSpaceHierarchy
+{
+    CoSpaceId       m_parent{lgrn::id_null<CoSpaceId>()};
+    SatId           m_parentSat{lgrn::id_null<SatId>()};
+};
+
 /**
  * @brief Get StridedArrayView1Ds from all components of a SatVectorDesc
  *
@@ -111,12 +118,6 @@ constexpr auto sat_views(
         return std::array{ satVec.view(Corrade::Containers::arrayView(rData), satCount, COUNT_T) ... };
     }
 }
-
-struct CoSpaceHierarchy
-{
-    CoSpaceId       m_parent{lgrn::id_null<CoSpaceId>()};
-    SatId           m_parentSat{lgrn::id_null<SatId>()};
-};
 
 struct CoSpaceTransform
 {
--- a/src/osp/universe/universe.h
+++ b/src/osp/universe/universe.h
@@ -84,13 +84,6 @@ constexpr VEC_T to_vec(INDEX_T i, RANGE_T&& ... args) noexcept
     return {args[std::size_t(i)] ...};
 }
 
-
-struct CoSpaceHierarchy
-{
-    CoSpaceId       m_parent{lgrn::id_null<CoSpaceId>()};
-    SatId           m_parentSat{lgrn::id_null<SatId>()};
-};
-
 /**
  * @brief Get StridedArrayView1Ds from all components of a SatVectorDesc
  *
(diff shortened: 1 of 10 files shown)


src/osp/universe/universe.h
```diff
@@ -124,18 +88,19 @@ struct CoSpaceTransform
...
     Quaterniond     m_rotation;
     Vector3g        m_position;
+
     int             m_precision{10}; // 1 meter = 2^m_precision
 };
 
-
 struct CoSpaceSatData
 {
     uint32_t        m_satCount;
```
jonesmz, line 97: Variable 'osp::universe::CoSpaceSatData::m_satCount' is uninitialized. Always initialize a member variable (type.6).

Warning
Variable 'osp::universe::CoSpaceSatData::m_satCount' is uninitialized. Always initialize a member variable (type.6).

src/osp/universe/universe.h
```diff
@@ -124,18 +88,19 @@ struct CoSpaceTransform
...
     uint32_t        m_satCapacity;
 
     Corrade::Containers::Array<unsigned char> m_data;
     SatVectorDesc<spaceint_t, 3>    m_satPositions;
-    SatVectorDesc<float, 3>         m_satVelocities;
+    SatVectorDesc<double, 3>        m_satVelocities;
+    SatVectorDesc<double, 4>        m_satRotations;
```
jonesmz, line 103: Variable 'osp::universe::CoSpaceSatData::m_satRotations' is uninitialized. Always initialize a member variable (type.6).

Warning
Variable 'osp::universe::CoSpaceSatData::m_satRotations' is uninitialized. Always initialize a member variable (type.6). 






Not clear to me if this is valid for a complex type like this.

Wonder if we need to adjust rules.

src/osp/universe/universe.h
```diff
@@ -162,4 +127,58 @@ struct Universe
...
+/**
+ * @brief Get StridedArrayView1Ds from all components of a SatVectorDesc
+ *
+ * @param satVec   [in] SatVectorDesc describing layout in rData
+ * @param rData    [in] unsigned char* Satellite data, optionally const
+ * @param satCount [in] Range of valid satellites
+ *
+ * @return std::array of views, intended to work with structured bindings
+ */
+template <typename T, int DIMENSIONS_T, typename DATA_T, int ... COUNT_T>
```
jonesmz, line 146: dimensions and count should be unsigned integer types, probably size_t.

signed vs unsigned is important because of the knock-on effects that it can have with integer overflow during multiplcation and all sorts of other crappy issues that are very hidden deep in implicit behavior land.

src/test_application/activescenes/scenarios.h
```diff
@@ -115,6 +115,13 @@ struct VehicleTest
 };
 
 
+struct UniverseTest
+{
+    static void setup_scene(CommonTestScene &rScene, osp::PkgId pkg);
```
jonesmz, line 120: setup_scene should be noexcept, shouldn't it?

Or will it throw exception on failure? if so, use `noexcept(false)` to convey that that's intentional.

src/test_application/activescenes/scenarios_universetest.cpp
```diff
@@ -0,0 +1,701 @@
...
+
+// for the 0xrrggbb_rgbf and angle literals
+using namespace Magnum::Math::Literals;
+
+using Corrade::Containers::Array;
+
+namespace testapp::scenes
+{
+
+constexpr float gc_physTimestep = 1.0 / 60.0f;
```
jonesmz, line 77: put these in an anymous namespace.

Helps the linker discard symbols that don't need to be visible outside the translation unit.

src/test_application/activescenes/scenarios_universetest.cpp
```diff
@@ -0,0 +1,701 @@
...
+    {
+        x[i] += vx[i] * scaleDelta;
+        y[i] += vy[i] * scaleDelta;
+        z[i] += vz[i] * scaleDelta;
+
+        // Apply arbitrary inverse-square gravity towards origin
+        Vector3d const pos       = Vector3d( Vector3g( x[i], y[i], z[i] ) ) * scale;
+        float const r           = pos.length();
+        float const c_gm        = 10000000000.0f;
+        Vector3d const accel     = -pos * delta * c_gm / (r * r * r);
```
jonesmz, line 437: Arithmetic overflow: Using operator '' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '' to avoid overflow (io.2).

Warning
Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).
Show more details

jonesmz (approved):

src/osp/universe/universe.h
jonesmz: If you didn't know, you can view these automated code scanning reports by looking at the Files Changed tab.

PR #210, opened 2022-05-30
Minor tweaks and small modernizations

jonesmz:


src/osp/Resource/load_tinygltf.cpp
```diff
@@ -136,7 +136,7 @@ static void load_gltf(TinyGltfImporter &rImporter, ResId res, std::string_view n
         }
 
         // Create and keep track of resource Id
-        ResId const imgRes = rResources.create(gc_image, pkg, format_name(rImporter.image2DName(i), i));
+        ResId const imgRes = rResources.create(gc_image, pkg, SharedString::create(format_name(rImporter.image2DName(i), i)));
```
Capital-Asterisk, line 139: No way to implicitly create shared strings from other standard strings? This can also be moved to the format_name lambda instead.

src/osp/Resource/machines.h
```diff
@@ -58,11 +58,10 @@ template<class WIRETYPE_T>
...
 
 template<class TYPE_T>
-constexpr TYPE_T nullvalue()
+constexpr TYPE_T nullvalue() noexcept
 {
-    return TYPE_T(std::numeric_limits<
-                  typename std::underlying_type_t<TYPE_T>
-                  >::max());
+    using Underlying_t = std::underlying_type_t<TYPE_T>;
+    return TYPE_T(std::numeric_limits<Underlying_t>::max());
```
Capital-Asterisk, line 64: This entire file is no longer used. It can be deleted.

src/osp/Resource/resources.cpp
```diff
@@ -44,16 +44,17 @@ ResId Resources::create(ResTypeId typeId, PkgId pkgId, std::string_view name)
...
+    auto const& [newIt, success] = rPkgType.m_nameToResId.emplace(std::move(name), newResId);
+    auto const& [key, value]     = *newIt;
+    assert(success); // emplace should always succeed
+
     rPerResType.m_resNames.resize(rPerResType.m_resIds.capacity());
-    SharedString& rResName = rPerResType.m_resNames[std::size_t(newResId)];
-    rResName = SharedString::create(name); // allocates name
-    [[maybe_unused]] auto newIt = rPkgType.m_nameToResId.emplace(rResName, newResId);
-    assert(newIt.second); // emplace should always succeed
+    rPerResType.m_resNames[std::size_t(value)] = key;
```
Capital-Asterisk, line 52: I feel that m_resNames may not have to exist. Names can be stored entirely in PerPkg. There eventually needs to be a way to get which PkgId a ResId belongs to, and be able to obtain the name that way.

src/osp/id_set.h
```diff
@@ -35,12 +35,12 @@ class IdSet
...
     {
-        auto const& [it, success] = m_data.emplace(key, ID_T(m_data.size()));
+        [[maybe_unused]] auto const& [it, success] = m_data.emplace(std::move(key), ID_T(m_data.size()));
         return it->second;
     }
 
 private:
-    std::unordered_map<KEY_T, ID_T> m_data;
+    std::unordered_map<KEY_T, ID_T, std::less<>> m_data;
 };
```
Capital-Asterisk, line 44: This class is unused. If I remember correctly, this came from my redstone repo where it turned out to be more of a burden than anything. It can be deleted.

src/osp/universe/coordinates.h
```diff
@@ -301,7 +302,7 @@ inline CoordTransformer coord_child_to_parent(
     return {
         .m_rotIn    = child.m_rotation,
         .m_c        = child.m_position,
-        .m_n        = -precisionDiff,
+        .m_n        = std::negate{}(precisionDiff),
```
Capital-Asterisk, line 305: Honestly, I find this incredibly noisy and slightly more difficult to read. I may even argue that it's easier to miss when mixed with other variable names and functions. Perhaps just use brackets or something? `foo = ( - bar)`

src/osp/universe/universe.h
```diff
@@ -169,15 +173,22 @@ constexpr auto sat_views(
...
-            .m_position = to_vec<Vector3g>(sat, x, y, z),
+            .m_rotation = {to_vec<Vector3d>(sat,
+                                            std::forward<ROTVIEW_T>(qx),
+                                            std::forward<ROTVIEW_T>(qy),
+                                            std::forward<ROTVIEW_T>(qz)),
+                           std::forward<ROTVIEW_T>(qw)[std::size_t(sat)]},
+            .m_position = to_vec<Vector3g>(sat,
+                                           std::forward<POSVIEW_T>(x),
+                                           std::forward<POSVIEW_T>(y),
+                                           std::forward<POSVIEW_T>(z)),
```
Capital-Asterisk, line 191: POSVIEW_T and ROTVIEW_T are intended to be ranges of potentially-const long ints and doubles. I don't believe rvalues or forward is ever needed here.

src/planet-a/SubdivSkeleton.h
```diff
@@ -142,11 +142,11 @@ class SubdivIdTree : private lgrn::IdRegistry<ID_T>
...
      */
     static constexpr combination_t
-    hash_id_combination(ID_T a, ID_T b) noexcept
+    hash_id_combination(ID_T const a, ID_T const b) noexcept
     {
         // Sort to make A and B order-independent
-        id_int_t const ls = (a > b) ? id_int_t(a) : id_int_t(b);
-        id_int_t const ms = (a > b) ? id_int_t(b) : id_int_t(a);
+        id_int_t const ls = id_int_t(std::max(a, b));
+        id_int_t const ms = id_int_t(std::max(a, b));
```
Capital-Asterisk, line 149: Perhaps you meant `std::min` on the least-significant?

src/planet-a/icosahedron.h
```diff
@@ -30,12 +30,14 @@
...
+#include <cstddef>
+
 namespace planeta
 {
 
 // Fundamental geometric properties of an icosahedron
-inline constexpr int const gc_icoVrtxCount = 12;
-inline constexpr int const gc_icoTriCount = 20;
+inline constexpr std::size_t const gc_icoVrtxCount = 12;
+inline constexpr std::size_t const gc_icoTriCount = 20;
```
Capital-Asterisk, line 40: Isn't std::size_t intended data sizes and array indices? This feels very out of place here describing a mathematical property.

src/test_application/activescenes/common_scene.cpp
```diff
@@ -57,16 +57,16 @@ void CommonTestScene::update_hierarchy_delete()
...
         m_deleteTotal.push_back(ent);
     });
 }
 
 void CommonTestScene::update_delete()
 {
-    auto first = std::cbegin(m_deleteTotal);
-    auto last = std::cend(m_deleteTotal);
+    auto const& first = std::cbegin(m_deleteTotal);
+    auto const& last = std::cend(m_deleteTotal);
```
Capital-Asterisk, line 69: Is it intentional that these are references?

src/osp/Resource/load_tinygltf.cpp
jonesmz: The reason to avoid doing that is to avoid surprise allocations.

I made the sharedstring class require the factory function for things other than copy/move construction to specificly make it hard to get one by accident.

src/planet-a/icosahedron.h
jonesmz: You're describing a "count" of something. Count's can't be negative and frequently are used as indicies into other objects, or are compared against other objects.

src/test_application/activescenes/common_scene.cpp
jonesmz: Yes. Many data structures do not return by value from their .begin() or .end() functions. Taking the result by const& instead of "auto" potentially avoids a copy-constructor call.

src/osp/Resource/resources.cpp
jonesmz: Certainly, but a bit outside what i was trying to do.

src/planet-a/icosahedron.h
Capital-Asterisk: How about just an unsigned int? std::size_t conveys an intention that this variable has something important to do with hardware, at least for me.

jonesmz: std::size_t has nothing to do with hardware, beyond having the same maximum value as the maximum number of addressable locations in memory on a particular platform.

std::size_t is "the number of things" variable. This is a count of vertices, std::size_t is the right variable type.

Capital-Asterisk: > std::size_t has nothing to do with hardware, beyond [reason why std::size_t has something to do with hardware]

???

'Counts' can include things like money, score, or number of items acquired, which may not correlate with something in memory. In this case, I don't believe size_t is intended for every count.

Consider other mathematical constants like Avogadro's number (~6.02e23), which is a count of particles in an SI mole. This is a bit of an extreme case but I hope it gets the point across.

I'm not concerned of any performance differences or potential widening/narrowing; simply the intention of what this variable is for. It can be used with array, but it's a general purpose mathematical constant that has nothing to do with hardware/platform/memory.

jonesmz: uint8_t better?

Capital-Asterisk: If I remember correctly, the compiler/analysers should complain less when compared with unsigned values, vs being a signed int before. That should be fine

PR #213, opened 2022-06-22
Add Tagged Tasks (for future Systems)

Capital-Asterisk:
Tagged tasks is a versatile solution for organising ECS systems and similar cases. A task graph is often not the right choice for a few reasons:
* Can't represent cases where multiple tasks have the same dependencies, but can't be run in parallel.
* Tasks directly depending on each other makes adding systems (VERY) difficult.

Tagged tasks, as the name implies, organizes tasks by assigning them tags. Rules can be set between tags that affect the tasks they are added to:
* Dependencies: Tags can depend on each other, fulfilling the same capabilities as a task graph. Dependencies restricts a task from running until all tasks containing tags it depends on are complete.
* Limits: Limit sets how many tasks using a certain tag can run simultaneously. This can be used as a semaphore.

Example usage can be found in the newly added unit test.

As most of my new code goes, Tasks and Tags are simply integer IDs. The implementation also makes good use of arrays and bitsets, which in theory leads to really good performance. e.g., which Tags a Task contains is implemented with bits, so bitwise operations are used to compare 64 tags at once.

This PR does NOT add multi-threading. Single threaded performance is already blazingly fast, so I'm not concerned about it at the moment.


z-adams (approved):

src/osp/tasks/execute_simple.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+void task_list_available(TaskTags const& tags, ExecutionContext const& exec, BitSpan_t tasksOut)
+{
+    assert(tasksOut.size() == tags.m_tasks.vec().size());
+
+    // Bitmask makes it easy to compare the tags of a task
+    // 1 = allowed (default), 0 = not allowed
+    // All tags of a task must be allowed for the entire task to run.
+    // aka: All ones in a task's bits must corrolate to a one in the mask
+    std::vector<uint64_t> mask(tags.m_tags.vec().size(), ~uint64_t(0));
```
jonesmz, line 104: std::numeric_limits<uint64_t>::max()

~uint64_t(0) is not the same on all platforms.

jonesmz (approved):

src/osp/tasks/execute_simple.cpp
Capital-Asterisk: This int is intended to have all the bits set to 1, and not a max value.

I'm aware there's undefined behaviour associated with signed integers, but I'm unsure of how `~uint64_t(0) ` can be problematic?

jonesmz: reading comprehension fail on my part. I see you had a comment explaining that. Never mind!

PR #214, opened 2022-08-25
Implement Fully Compsable scenes

Capital-Asterisk:
This makes test scenes extremely composable. 

Making scenes is a matter of slapping together the right functions and data, but this is done in a way that the functions will automatically organize to run in the correct order (or in parallel, but that's not yet implemented).

## Problem

Previously, scene data was stored in a 'CommonTestScene' that featured an `std::vector<entt::any>` to store various different unique types. Access to these types are uncontrolled, and large structs were encouraged. Updating the scene was simply a very long function called from the renderer. Separating scene code and rendering code was rather difficult, and there's already a few bugs related to certain function being called in the wrong order. I insisted that multithreading it would be pretty easy, but not until some kind of task system was implemented (this PR). Code is also duplicated between the physics, vehicle, and universe test scenes.

## New architecture

New architecture goes something like this:
* The state of most things in the application are stored in a top-level `std::vector<entt::any>` in main.cpp (g_appTopData). This stores the **TopData**, which are various non-unique types addressable using a TopDataId enum integer. 
* The TopData is accessed by **TopTasks**. Also at the top-level of the application in main.cpp,
* TopTasks are organized using **TaskTags** to form task graphs and events.
* TopData, TopTasks, and TaskTags can be conveniently bundled together into **Sessions**.  Various Sessions can depend on each other by sharing TopData and TaskTags.

Sessions are roughly intended to add support for something to exist in the world. Scenes are created by setting up lots of Sessions. The Physics test scene for example, is a composition of a common scene session + a material session + generic physics session + newton physics session + ...  see scenarios.cpp

Renderer sessions are setup when OpenGL and Magnum is used (when the window is opened). These sessions can be things like camera controls or shaders. They are removed when the window is closed.

The application is responsible for managing Sessions. Adding support for multiple scenes, a universe, or a user interface will likely be just adding more std::vectors or other containers full of Sessions.

Vehicle test scene and Universe test scene are not yet ported over to the new architecture.

Other goodies:
* Test scenes can now more easily share code
* Full separation between rendering, physics engine, and pretty much everything
* Parallel multithreading can now be easily added by rewriting osp::top_run_blocking in src/osp/tasks/top_execute.cpp
* Very elegant. I'm quite confident in this architecture. Nothing seems out-of-place or unnecessary, and thing seem to 'fit together' perfectly while also highly respecting cache locality. Performance should be quite high in theory, with plenty of room to optimize.

Other changes
* Replace camera entity and component with a Camera struct in the common scene renderer session. why did the camera need to be an entity with a camera component in the first place?


src/test_application/activescenes/scenarios.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+            auto & [scnRender, simpleCam, shVisual, camThrow] = unpack<4>(rendererOut);
+            scnRender   = setup_scene_renderer      (builder, rTopData, rTags, magnum, scnCommon, mainView.m_idResources);
+            simpleCam   = setup_simple_camera       (builder, rTopData, rTags, magnum, scnCommon, scnRender);
+            shVisual    = setup_shader_visualizer   (builder, rTopData, rTags, magnum, scnCommon, scnRender, matVisual);
+            camThrow    = setup_thrower             (builder, rTopData, rTags, magnum, scnRender, simpleCam, shapeSpawn);
+
+            setup_magnum_draw(mainView, magnum, scnCommon, scnRender);
+        };
+    });
```
Capital-Asterisk, line 159: Important part. This is where the physics test scene is created, which is composed of Sessions created by these setup_\* functions. A new scenario can be easily created by copy-pasting this whole block of code and giving it a different name.

jonesmz (dismissed):
I have no complaints, all looks ok to me.

jonesmz:
Errrr
scratch that, somehow github only showed me one commit.

I need more time to review, i'm going out of town.

Let me look next week.

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -144,9 +144,9 @@ void SysPrefabInit::init_drawing(
 
             }
 
-            MaterialData &rMaterial = rCtxDraw.m_materials[materialId];
-            rMaterial.m_comp.emplace(ent);
-            rMaterial.m_added.push_back(ent);
+            //MaterialData &rMaterial = rCtxDraw.m_materials[materialId];
```
jonesmz, line 147: Minor: Either remove the commented out code, or add an explanation for why it's being kept.

src/osp/Active/SysRender.h
```diff
@@ -113,6 +113,11 @@ struct RenderGroup
 
 struct ACtxRenderGroups
 {
+    // Required for std::is_copy_assignable to work properly inside of entt::any
+    ACtxRenderGroups() = default;
+    ACtxRenderGroups(ACtxRenderGroups const& copy) = delete;
```
jonesmz, line 118: Are you meaning that you're intentionally disabling copying?

jonesmz: If so, probably should say as much.

But why bother disabling copying?

src/osp/Active/drawing.h
```diff
@@ -131,6 +120,11 @@ struct ACtxDrawing
  */
 struct ACtxDrawingRes
 {
+    // Required for std::is_copy_assignable to work properly inside of entt::any
+    ACtxDrawingRes() = default;
+    ACtxDrawingRes(ACtxDrawingRes const& copy) = delete;
```
jonesmz, line 125: same feedback as ACtxRenderGroups

src/osp/Shaders/MeshVisualizer.h
```diff
@@ -57,10 +57,38 @@ void draw_ent_visualizer(
         active::ViewProjMatrix const& viewProj,
         active::EntityToDraw::UserData_t userData) noexcept;
 
-
-void assign_visualizer(
-        active::RenderGroup::ArrayView_t entities,
+template<typename ITA_T, typename ITB_T>
+void sync_visualizer(
+        ITA_T dirtyFirst, ITB_T const& dirtyLast,
```
jonesmz, line 62: minor: consider splitting this to a separate line like the other params.

src/osp/Shaders/Phong.h
```diff
@@ -74,20 +74,70 @@ void draw_ent_phong(
...
  * @param diffuse               [in] Storage for diffuse texture component
  * @param rData                 [in] Phong shader data, stable memory required
  */
-void assign_phong(
-        active::RenderGroup::ArrayView_t entities,
-        active::RenderGroup::Storage_t *pStorageOpaque,
-        active::RenderGroup::Storage_t *pStorageTransparent,
+template<typename ITA_T, typename ITB_T>
+void sync_phong(
+        ITA_T dirtyFirst, ITB_T const& dirtyLast,
```
jonesmz, line 88: minor split to two lines

src/osp/tasks/builder.h
```diff
@@ -86,9 +93,10 @@ class TaskBuilder
             return *this;
         }
 
-        TaskTags::Task const m_taskId;
-        TaskTags &m_rTags;
-        DATA_T &m_rData;
+        TaskId const    m_taskId;
+        Tags            &m_rTags;
```
jonesmz, line 97: please put a space between the & and the variable.
I almost missed that these were references.

src/osp/tasks/execute_simple.cpp
```diff
@@ -25,58 +25,67 @@
...
-void task_enqueue(TaskTags const& tags, ExecutionContext &rExec, BitSpanConst_t const query)
+bool any_bits_match(BitSpanConst_t const lhs, BitSpanConst_t const rhs)
 {
-    std::size_t const tagIntSize = tags.tag_ints_per_task();
-    assert(query.size() == tagIntSize);
+    return ! std::equal(
+            std::cbegin(lhs), std::cend(lhs),
+            std::cbegin(rhs), std::cend(rhs),
+            [] (uint64_t const lhsInt, uint64_t const rhsInt)
+    {
```
jonesmz, line 44: What's forcing the need for this lambda?

and why not use std::ranges::equal() ? i got the codebase compiling with C++20 didn't i?

src/osp/tasks/top_execute.cpp
```diff
@@ -0,0 +1,169 @@
...
+    rExec.m_tagIncompleteCounts  .resize(tags.m_tags.capacity(), 0);
+    rExec.m_tagRunningCounts     .resize(tags.m_tags.capacity(), 0);
+    rExec.m_taskQueuedCounts     .resize(tasks.m_tasks.capacity(), 0);
+
+    std::vector<uint64_t> tagsToRun(tags.m_tags.vec().size());
+    to_bitspan(tagsEnq, tagsToRun);
+    task_enqueue(tags, tasks, rExec, tagsToRun);
+}
+
+static void check_task_dependencies(Tags const& tags, Tasks const& tasks, TopTaskDataVec_t const& taskData, std::vector<uint32_t> &rPath, uint32_t const task, bool& rGood)
```
jonesmz, line 97: static functions aren't visible outside the cpp file they're defined in, but a lot of people get tripped up on that.

I recommend putting this into an anonymous namespace (as well as leaving it static) so that there's no ambiguity

src/osp/tasks/top_execute.cpp
```diff
@@ -0,0 +1,169 @@
...
+        if (any_bits_match(taskTags2d[currTask].asContiguous(), dependencyTagInts))
+        {
+            auto const &last = std::end(rPath);
+            if (last == std::find(std::begin(rPath), last, currTask))
+            {
+                check_task_dependencies(tags, tasks, taskData, rPath, currTask, rGood);
+            }
+            else
+            {
+                rGood = false;
```
jonesmz, line 137: this is the only place where this reference parameter is modified.

i recommend making it a return value instead of a reference param.

src/osp/tasks/top_execute.cpp
```diff
@@ -0,0 +1,169 @@
...
+                {
+                    ss << "* [" << pathTask << "]: " << taskData.m_taskData[pathTask].m_debugName << "\n";
+                }
+                ss << "* Loops back to [" << currTask << "]\n";
+                OSP_LOG_ERROR("{}", ss.str());
+            }
+        }
+    }
+
+    rPath.pop_back();
```
jonesmz, line 151: if you instead were to return on line 146 or line 133, the compiler would be able to implement this recursive call as a tail-call, which would drastically cut down on stack usage and functioncall overhead.

src/osp/tasks/top_session.cpp
```diff
@@ -0,0 +1,98 @@
...
+using Corrade::Containers::arrayView;
+
+namespace osp
+{
+
+void top_close_session(Tags& rTags, Tasks& rTasks, TopTaskDataVec_t& rTaskData, ArrayView<entt::any> topData, ExecutionContext& rExec, ArrayView<Session> sessions)
+{
+    // Run cleanup tasks
+    {
+        // Accumolate together all cleanup tags from all sessons
```
jonesmz, line 45: Accumulate

src/osp/tasks/top_tasks.h
```diff
@@ -0,0 +1,63 @@
...
+#include <vector>
+
+namespace osp
+{
+
+// TODO: ignore debug names on release builds
+
+struct TopTask
+{
+    std::string m_debugName;
```
jonesmz, line 40: #ifndef _NDEBUG

i believe.

But better would be to define a new type "DebugString" that has no member variables in debug builds and returns empty string views for all it's functions.

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+
+namespace osp
+{
+
+[[nodiscard]] inline TopDataId top_reserve(
+        ArrayView<entt::any> const topData, TopDataId current = 0)
+{
+    while ( bool(topData[current]) && (current < topData.size()) )
+    {
+        ++current;
```
jonesmz, line 45: auto const& it = std::ranges::find(topData, true);
current = std::distance(std::begin(topData), it);

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+
+
+
+template<typename FUNCTOR_T>
+struct wrap_args_trait
+{
+    template<typename ... ARGS_T, std::size_t ... INDEX_T>
+    static constexpr void apply_void(ArrayView<entt::any> topData, [[maybe_unused]] std::integer_sequence<std::size_t, INDEX_T...> indices) noexcept
+    {
+        FUNCTOR_T{}(entt::any_cast<ARGS_T&>(topData[INDEX_T]) ...);
```
jonesmz, line 111: use std::invoke

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+    {
+        assert(topData.size() == sizeof...(ARGS_T));
+
+        apply_void<ARGS_T ...>(topData, std::make_integer_sequence<std::size_t, sizeof...(ARGS_T)>{});
+
+        return TopTaskStatus::Success;
+    }
+
+    template<typename ... ARGS_T, std::size_t ... INDEX_T>
+    static constexpr TopTaskStatus apply(ArrayView<entt::any> topData, [[maybe_unused]] std::integer_sequence<std::size_t, INDEX_T...> indices) noexcept
```
jonesmz, line 125: just omit the name of 'indicies'

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+    {
+        FUNCTOR_T{}(entt::any_cast<ARGS_T&>(topData[INDEX_T]) ...);
+    }
+
+    template<typename ... ARGS_T>
+    static TopTaskStatus wrapper_void([[maybe_unused]] WorkerContext& rCtx, ArrayView<entt::any> topData) noexcept
+    {
+        assert(topData.size() == sizeof...(ARGS_T));
+
+        apply_void<ARGS_T ...>(topData, std::make_integer_sequence<std::size_t, sizeof...(ARGS_T)>{});
```
jonesmz, line 119: std::make_index_sequence

src/osp/tasks/top_utils.h
jonesmz: and re-type it as std::index_sequence

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+    {
+        return FUNCTOR_T{}(entt::any_cast<ARGS_T&>(topData[INDEX_T]) ...);
+    }
+
+    template<typename ... ARGS_T>
+    static TopTaskStatus wrapper([[maybe_unused]] WorkerContext& rCtx, ArrayView<entt::any> topData) noexcept
+    {
+        assert(topData.size() == sizeof...(ARGS_T));
+
+        return apply<ARGS_T ...>(topData, std::make_integer_sequence<std::size_t, sizeof...(ARGS_T)>{});
```
jonesmz, line 135: std::index_sequence_for<ARGS_T...>{}

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    template<typename ... ARGS_T>
+    static constexpr TopTaskFunc_t unpack([[maybe_unused]] TopTaskStatus(*func)(ARGS_T...))
```
jonesmz, line 139: it appears that all of these only ever work with noexcept functions.

if so, mark the functions that you're taking as parameters as noexcept so that no one gives you an exception throwing function.

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * wrap_args creates a wrapper function that will automatically cast the
+ * entt::anys and call the underlying function.
+ *
+ * @param a[in]
+ *
+ * @return TopTaskFunc_t function pointer to wrapper
+ */
+template<typename FUNC_T>
+constexpr TopTaskFunc_t wrap_args(FUNC_T const&& funcArg)
```
jonesmz, line 167: const&& doesn't exist.

What are you trying to do here?

src/osp/tasks/top_utils.h
```diff
@@ -0,0 +1,183 @@
...
+{
+    // TODO: wrap function pointers into a functor
+    static_assert ( ! std::is_function_v<FUNC_T>, "Support for function pointers not yet implemented");
+
+    using static_lambda_t = std::remove_cvref_t<decltype(funcArg)>;
```
jonesmz, line 172: This seems like a pretty big assumption.

I'm struggling to understand what your code is trying to accomplish.

src/osp/unpack.h
```diff
@@ -22,26 +22,29 @@
...
-    entt::any sceneAny = entt::make_any<FlightScene>();
-    FlightScene &rScene = entt::any_cast<FlightScene&>(sceneAny);
-
-    rScene.m_nwtWorld = std::make_unique<ospnewton::ACtxNwtWorld>(1);
+    using ptr_t = decltype(rIn.data());
+    using type_t = std::remove_pointer_t<ptr_t>;
 
-    return sceneAny;
+    assert(N <= rIn.size());
+    return *reinterpret_cast<type_t(*)[N]>(std::data(rIn));
```
jonesmz, line 45: i don't think you should be trying to cast the return value of std::data() to pointer-to-array.

What are you actually trying to accomplish?

src/test_application/activescenes/scenarios_universetest.cpp
```diff
@@ -22,6 +22,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
+#if 0
```
jonesmz, line 25: just delete files that you're deleting. having a bunch of stuff commented out, or disabled with #if 0, only adds to the confusion of other contributors.

jonesmz (approved):
Approved, but i recommend addressing some of my feedback before merging.

src/osp/Active/SysRender.h
Capital-Asterisk: Some members of this struct are not copyable, making the struct implicitly uncopyable. I'm unsure of the exact details, but not writing this explicitly this causes a compile error.

src/osp/tasks/execute_simple.cpp
Capital-Asterisk: std::equal just happens to work as a way to run a functor element-wise against two ranges, and return true if they all return true. The original 'equals' intent of this function is misleading here, and I might as well use a for loop here.

I was not aware of std::ranges::equal when writing this.

src/osp/tasks/top_execute.cpp
Capital-Asterisk: Is there really any risk here worth adding an extra pair of braces to keep track of? What exactly is this trip up?

src/osp/tasks/top_execute.cpp
Capital-Asterisk: I don't think it's possible to call `rPath.pop_back();` after a tail call? This goes less than 10 calls deep and is primarily intended for checking if a state is valid, so I don't think it's worth figuring this out.

src/osp/tasks/top_execute.cpp
Capital-Asterisk: In this case, I find it easier to understand when there is only one instance of the 'good' boolean.

src/osp/tasks/top_utils.h
Capital-Asterisk: Well this caused an unexpected segfault. For a comparison between entt::any and bool, the boolean is used to initialize a new entt::any with the bool. I may bug skypjack about this.

src/osp/tasks/top_utils.h
Capital-Asterisk: Assuming you're not referring to the doc, I put const there likely because funcArg isn't modified? This function only needs to get the type FUNC_T to be able to convert it to a function pointer.

src/osp/tasks/top_utils.h
Capital-Asterisk: I think this line in particular was left over from some removed code? It still worked when removed, and FUNC_T can be used directly.

src/osp/unpack.h
Capital-Asterisk: I'd agree this may be sketchy, but so far it works. This function should return a type that works with structured bindings (a C-array) to assign variable names to elements in a (contiguous) range, inspired by unpacking in Python.

i.e.
```cpp
std::vector<int> stuff = {24, 165, 5};
auto & [a, b, c] = osp::unpack<3>(stuff);
```

src/osp/tasks/top_execute.cpp
jonesmz: anonymous namespaces and static keywords are both signals to the reader about properties of the thing being annotated. Both is less easy to miss than only one.

PR #221, opened 2022-10-21
Add Vehicle Scenario

Capital-Asterisk:
This PR adapts previously-deleted vehicle code new task system, being far more fine-grained along with other improvements. Unfortunately, the don't apply thrust yet, and only print values to the terminal.

Task improvements:
* Remove some unused task enqueue features
* Add way to make tasks enqueue other tasks by tag
* Changed circular dependency checker into a deadlock detector instead (wiring tasks call each other to loop)

Other changes:
* Separated rendering and input tasks from camera controller
* Name changes and other code improvements


Capital-Asterisk:
By the looks of it, some of the compilers in the CI don't fully support C++20, and are failing to find some std::ranges functions

src/osp/tasks/top_session.cpp
```diff
@@ -90,6 +98,9 @@ void top_close_session(Tags& rTags, Tasks& rTasks, TopTaskDataVec_t& rTaskData,
             if (tag != lgrn::id_null<TagId>())
             {
                 rTags.m_tags.remove(tag);
+
+                auto tagDepends = depends2d[std::size_t(tag)].asContiguous();
+                std::fill(std::begin(tagDepends), std::end(tagDepends), lgrn::id_null<TagId>());
```
jonesmz, line 103: std::ranges::fill?

src/osp/tasks/top_utils.h
```diff
@@ -106,48 +106,45 @@ template <typename T>
...
+    static TopTaskStatus wrapped_task([[maybe_unused]] WorkerContext ctx, ArrayView<entt::any> topData) noexcept
     {
-        // Parameter pack used to extract function arguments, func is discarded
-        return &wrapper<ARGS_T ...>;
+        if constexpr (std::is_void_v<RETURN_T>)
+        {
+            cast_args<ARGS_T ...>(topData, ctx, std::make_index_sequence<sizeof...(ARGS_T)>{});
+            return TopTaskStatus::Success;
+        }
+        else if (std::is_same_v<RETURN_T, TopTaskStatus>)
```
jonesmz, line 137: Need constexpr here too

src/test_application/VehicleBuilder.cpp
```diff
@@ -226,6 +229,10 @@ void VehicleBuilder::finalize_machines()
             }
         }
     }
+
+    VehicleData dataOut{std::move(*m_data)};
```
jonesmz, line 233: Consider using std::exchange here

src/test_application/VehicleBuilder.h
```diff
@@ -156,12 +162,12 @@ template <std::size_t N>
...
-    m_data.m_partIds.create(std::begin(out), std::end(out));
+    m_data->m_partIds.create(std::begin(out), std::end(out));
 
-    std::size_t const capacity = m_data.m_partIds.capacity();
-    m_data.m_partMachCount.resize(capacity);
-    m_data.m_partPrefabs.resize(capacity);
-    m_data.m_partTransforms.resize(capacity);
+    std::size_t const capacity = m_data->m_partIds.capacity();
+    m_data->m_partMachCount.resize(capacity);
+    m_data->m_partPrefabs.resize(capacity);
```
jonesmz, line 169: Consider making a helper function for these repeated resize calls

src/test_application/activescenes/scene_misc.cpp
```diff
@@ -143,14 +143,61 @@ void add_floor(ArrayView<entt::any> const topData, Session const& scnCommon, Ses
...
+            TopDataIds_t{                            idCamCtrl,        idCamera},
+            wrap_args([] (ACtxCameraController const& rCamCtrl, Camera &rCamera) noexcept
+    {
+        rCamera.m_transform = rCamCtrl.m_transform;
+    }));
+
+    return cameraCtrl;
+}
+
+Session setup_camera_free(Builder_t& rBuilder, ArrayView<entt::any> const topData, Tags& rTags, Session const& app, Session const& scnCommon, Session const& camera)
```
jonesmz, line 173: I recommend adding some.new lines

jonesmz (approved):

src/test_application/VehicleBuilder.cpp
Capital-Asterisk: I recall attempting that, but can't quite figure it out. VehicleData is move-only, m_data is an optional saying the move assignment operator is deleted.
```cpp
// error: use of deleted function ‘std::optional<testapp::VehicleData>& std::optional<testapp::VehicleData>::operator=(std::optional<testapp::VehicleData>&&)’
VehicleData dataOut{std::exchange(m_data, {}).value()};
```

src/test_application/VehicleBuilder.h
Capital-Asterisk: This is the only place these are called, adding a new function would just unnecessary lines and more complexity in this case.

Additionally, these parallel vectors are everywhere, and it's possible to put them into single allocations. I haven't yet finalized the best way to do this, but I have a few bad ideas.

EhWhoAmI (approved):

Capital-Asterisk:
I don't believe it's worth breaking this much compiler support just to save less than a line of code here and there. I'll just revert these instead.

PR #224, opened 2022-12-11
Rework Hierarchy into "Scene Graph" and Simplify Newton integration

Capital-Asterisk:
_ok, now we got vehicles spawning again, time to make them apply thrust
...how lol?_

## New Scene Graph (replaces SysHierarchy) 

Instead of a linked-list of entities and components, the new scene graph is just a couple of int vectors to describe entities and descendant counts.

From basic.h
```cpp
struct ACtxSceneGraph
{
    // Tree structure stored using an array of descendant count in parallel with
    // identificaton (entities)
    // A(B(C(D)), E(F(G(H,I)))) -> [A,B,C,D,E,F,G,H,I] and [8,2,1,0,4,3,2,0,0]
    std::vector<ActiveEnt>  m_treeToEnt{lgrn::id_null<ActiveEnt>()};
    std::vector<uint32_t>   m_treeDescendants{std::initializer_list<uint32_t>{0}};

    std::vector<ActiveEnt>  m_entParent;
    std::vector<TreePos_t>  m_entToTreePos;

    // ...
};
```

Traversing a subtree is extremely fast, greatly simplifying and speeding up everything, such as calculating transforms for drawing. Other operations are far easier to do. For example, checking if an entity is a descendant of another entity is a matter of checking if its position in the vector is contained within its ancestor.

Random insertions and deletions from the tree is a bit harder, but this is never done anywhere. The architecture favors separate steps for deleting and creating new instances. All things that need to be deleted are all deleted before and all things that need to be created are created.

Entire subtrees must be inserted at once, the number of entities should be known beforehand. The SubtreeBuilder interface is added to assist with this.

## Simplified Newton Dynamics integration

Previously there was some kind of 'generic physics interface' in the form of ECS-style components. ie: ACompPhysLinearVel, ACompPhysDynamic, ... in physics.h.

This is the wrong abstraction, and gets extremely complex as EVERY feature we need from the physics engine must also be implemented by generic physics.

`OSP Vehicles` <--code--> `OSP Generic Physics` <--code--> `Newton Dynamics`

The new architecture can cleanly handle a more direct approach.

`OSP Vehicles` <--code--> `Newton Dynamics`

Simplified snippet from scenarios.cpp. Newton-specific code is separated into their own tasks and data:
```cpp
parts               = setup_parts(...)
vehicleSpawn        = setup_vehicle_spawn(... parts)
newton              = setup_newton(...)
vehicleSpawnNwt     = setup_vehicle_spawn_newton(... parts, vehicleSpawn, newton)

// just write a setup_physx and setup_vehicle_spawn_physx if we were to add PhysX support
```

## How to Rocket thrust? Use ForceFactors!

Previously, lots of separate systems would write forces into rigid bodies. Rocket thrust, gravity, and potentially aerodynamic forces are separate calls that roughly write to a single net force value for each rigid body.

This multiple-writers scheme is simply difficult to multi-thread. Atomics would help, but the problem is best avoided entirely. 

Solution: Make the physics engine read from multiple sources to calculate all factors that contribute to force for each rigid body.

Each rigid body is assigned a bitset, which describe which functions are called in order to calculate its net force (and torque too). These functions are referred to as "ForceFactors." See `setup_newton_force_accel(...)` in scene_newton.cpp for a constant-acceleration ForceFactor used for gravity.

## Other changes

* Rename RigidGroup to Weld. Planning on documenting what these are later, ask if curious.


src/adera/Shaders/PlumeShader.cpp
```diff
@@ -69,14 +69,14 @@ void PlumeShader::draw_plume(
     ACtxPlumeData &rData = *reinterpret_cast<ACtxPlumeData*>(userData[0]);
 
     // Collect uniform data
-    ACompDrawTransform const &drawTf    = rData.m_pDrawTf->get(ent);
+    Matrix4 const &drawTf               = rData.m_pDrawTf->get(ent);
```
jonesmz, line 72: I guess that I personally find it difficult to notice the ampersand, if it's attached to the variable name. But it's not a big deal

src/adera/Shaders/PlumeShader.cpp
```diff
@@ -69,14 +69,14 @@ void PlumeShader::draw_plume(
...
     ACompExhaustPlume const &comp       = rData.m_pExaustPlumes->get(ent);
     PlumeEffectData const &effect       = comp.m_effect;
     MeshGlId const meshId               = rData.m_pMeshId->get(ent);
     Magnum::GL::Mesh &rMesh             = rData.m_pMeshGl->get(meshId);
     Magnum::GL::Texture2D &rTmpTex      = rData.m_pTexGl->get(rData.m_tmpTex);
 
-    Magnum::Matrix4 entRelative = viewProj.m_view * drawTf.m_transformWorld;
+    Magnum::Matrix4 entRelative = viewProj.m_view * drawTf;
```
jonesmz, line 79: minor: const

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -76,75 +70,114 @@ using osp::Matrix3;
 using osp::Matrix4;
 using osp::Vector3;
 
-namespace
-{
 
 // Callback called for dynamic rigid bodies for applying force and torque
-static void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
+void SysNewton::cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex)
```
jonesmz, line 75: I recommend making a typedef to repreent the thread index type.

It can just be "using ThreadIndex_t = intl" for all i care, just always bugs me to see "int" is all.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -76,75 +70,114 @@ using osp::Matrix3;
...
-    uintptr_t const userData = reinterpret_cast<uintptr_t>(
+    auto const userData = reinterpret_cast<uintptr_t>(
             NewtonBodyGetUserData(pBody));
-    ActiveEnt const ent = static_cast<ActiveEnt>(userData);
-
-    // Apply force
-
-    Vector3 force;
-    bool hasForce = false;
+    auto const bodyId = static_cast<BodyId>(userData);
```
jonesmz, line 84: Are we using the pointer, casted to an int, as the Id?

That's fair i guess.

Are you certain that BodyId and uintptr_t will fit properly? Is that asserted somewhere?

Honestly might be worth considering making an adapter function to "hide" how someone converts from the return type of NewtonBodyGetUserData -> BodyId

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -76,75 +70,114 @@ using osp::Matrix3;
...
+
+NewtonCollision* SysNewton::create_primative(
+        ACtxNwtWorld&   rCtxWorld,
+        EShape          shape)
+{
+    NewtonWorld* pNwtWorld = rCtxWorld.m_world.get();
+    NewtonCollision* pCollision;
+    switch (shape)
+    {
+    case EShape::Sphere:
```
jonesmz, line 133: Minor: 

Consider the new (C++20) "using EShape;" syntax for switch-cases for readability.

This switch-case is small so the improvement is small.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -76,75 +70,114 @@ using osp::Matrix3;
...
+{
+    if (shape != EShape::Cylinder)
+    {
+        Matrix4 const matrix = Matrix4::from(rotation, translation);
+        NewtonCollisionSetMatrix(pCollision, matrix.data());
+        NewtonCollisionSetScale(pCollision, scale.x(), scale.y(), scale.z());
+    }
+    else
+    {
+        constexpr Matrix3 const rotX90deg
```
jonesmz, line 166: minor: static.

Not all compilers get confused by this, but some do. I've seen assembler output where the static keyword is what makes the compiler omit the call to the constructor and do inlining properly.

That might be fixed on newer compilers though, i haven't checked in a long time.

src/newtondynamics_physics/SysNewton.cpp
```diff
@@ -162,164 +195,69 @@ void SysNewton::update_translate(ACtxPhysics& rCtxPhys, ACtxNwtWorld& rCtxWorld)
...
-    // requires additional indirection
-    rCtxWorld.m_forceTorqueIn.resize(inputs.size());
-    for (int i = 0; i < inputs.size(); i ++)
-    {
-        std::swap(rCtxWorld.m_forceTorqueIn[i].m_force, inputs[i].m_physNetForce);
-        std::swap(rCtxWorld.m_forceTorqueIn[i].m_torque, inputs[i].m_physNetTorque);
+        NewtonBodySetVelocity(pBody, vel.data());
     }
 
+    rCtxWorld.m_pTransform = &rTf;
```
jonesmz, line 218: i recommend std::addressof() to get the address of things.

The & is fine, but it's hard to notice if you're reading quickly.

src/newtondynamics_physics/SysNewton.h
```diff
@@ -52,12 +50,30 @@ class SysNewton
...
-    using ACompHierarchy            = osp::active::ACompHierarchy;
+    using ACtxSceneGraph            = osp::active::ACtxSceneGraph;
     using ACompTransform            = osp::active::ACompTransform;
     using ACompTransformControlled  = osp::active::ACompTransformControlled;
     using ACompTransformMutable     = osp::active::ACompTransformMutable;
 public:
 
+    static void cb_force_torque(const NewtonBody* pBody, dFloat timestep, int threadIndex);
+
+    static void cb_set_transform(NewtonBody const* const pBody, const dFloat* const pMatrix, int threadIndex);
```
jonesmz, line 61: const dFloat* const -> dFloat const* const

src/newtondynamics_physics/SysNewton.h
```diff
@@ -52,12 +50,30 @@ class SysNewton
...
+
+    static void resize_body_data(ACtxNwtWorld& rCtxWorld);
+
+    [[nodiscard]] static NewtonCollision* create_primative(
```
jonesmz, line 65: This lacks lifetime management.

I recommend either returning an std::unique_ptr<>, to clearly signal that the returned pointer isn't managed by the library

or something related to these attributes: https://clang.llvm.org/docs/AttributeReference.html#handle-attributes

src/newtondynamics_physics/factors.h
```diff
@@ -24,10 +24,15 @@
...
-#include <unordered_map>
+namespace ospnewton
+{
 
-using MapSatToEnt_t = std::unordered_map<osp::universe::Satellite,
-                                         osp::active::ActiveEnt>;
+// Each rigid body is given 64 bits to enable/disable forces
+// These determine which physics calculations are required for a certain
+// rigid body, such as gravity, thurst, or aerodynamics.
+using ForceFactors_t = std::array<uint64_t, 1u>;
```
jonesmz, line 36: Bit lost.

This is a 64bit wide bitfield?

I recommend using either a proper enum holding the designated bits, or use a struct that holds a uint64_t that's broken out into bit fields at the language level. e.g.

struct Forces
{
    uint64_t m_gravity : 1;
    uint64_t m_solarWind : 1;
    uint64_t m_wind : 1;
    uint64_t m_explosions : 1;
    uint64_t m_unspecified : 60;
}

src/newtondynamics_physics/ospnewton.h
```diff
@@ -38,30 +46,35 @@ struct DeleterNewtonBody
     { NewtonDestroyBody(pCollision); }
 };
 
-using ACompNwtBody_t = std::unique_ptr<NewtonBody const, DeleterNewtonBody>;
+using NwtBodyPtr_t = std::unique_ptr<NewtonBody, DeleterNewtonBody>;
 
 struct DeleterNewtonCollision
```
jonesmz, line 51: You might consider making a utility class like utils/pointer_to_function.h
```
template<auto FUNCTION>
struct PointerToFunctionAdapter
{
    static_assert(nullptr != FUNCTION)
    decltype(auto) operator()(auto && ... args)
    {
        return std::invoke(FUNCTION, std::forward<decltype(args)>(args)...);
    }
};
```

This lets you define the struct used for your unique_ptr once, and then save on having to define a new struct.

src/newtondynamics_physics/ospnewton.h
```diff
@@ -70,34 +83,28 @@ struct ACtxNwtWorld
     };
 
     ACtxNwtWorld(int threadCount)
-     : m_nwtWorld(NewtonCreate())
-     , m_perThread(threadCount)
+     : m_world(NewtonCreate())
```
jonesmz, line 86: I guess either we should fix the thing that the CodeScanner is complaining about, or we should turn that warning off.

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -41,35 +42,43 @@ using osp::phys::EShape;
 namespace osp::active
 {
 
-void SysPrefabInit::init_hierarchy(
-        ACtxPrefabInit const&               rPrefabInit,
+void SysPrefabInit::add_to_subtree(
+        TmpPrefabInitBasic const&           basic,
+        ArrayView<ActiveEnt const>          ents,
```
jonesmz, line 47: would std::span<> work as a replacement for "ArrayView"?

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -41,35 +42,43 @@ using osp::phys::EShape;
...
+            gc_importer, basic.m_importerRes);
+    auto const &rPrefabData = rResources.data_get<osp::Prefabs const>(
+            gc_importer, basic.m_importerRes);
 
-    for (TmpPrefabInitBasic const& rPfBasic : rPrefabInit.m_basic)
-    {
-        auto const &rPrefabData = rResources.data_get<osp::Prefabs const>(
-                gc_importer, rPfBasic.m_importerRes);
+    auto const objs     = rPrefabData.m_prefabs[basic.m_prefabId];
+    auto const parents  = rPrefabData.m_prefabParents[basic.m_prefabId];
```
jonesmz, line 57: i assume these are cheap to copy?

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -41,35 +42,43 @@ using osp::phys::EShape;
...
+    auto itEnt      = std::begin(ents);
 
-        for (std::size_t i = 0; i < objCount; ++i)
-        {
-            ObjId const prefabParent = parents[i];
-            ActiveEnt const parent = (prefabParent == -1)
-                    ? rPfBasic.m_parent
-                    : (*itPfEnts)[prefabParent];
+    auto const add_child_recurse
+            = [&rImportData, &itObj, &itEnt, &parents] (auto&& self, SubtreeBuilder& bldParent) -> void
```
jonesmz, line 63: conceptually, since this lambda modifies things it points to, you would probably want to mark it as mutable.

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -121,19 +131,25 @@ void SysPrefabInit::init_drawing(
 
         for (std::size_t i = 0; i < objects.size(); ++i)
         {
+            ActiveEnt const ent = (*itPfEnts)[i];
+
+            // TODO: Don't actually do this. This marks every entity as drawable,
```
jonesmz, line 136: ??

src/osp/Active/SysSceneGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+        while (parentNotNull)
+        {
+            parent = rScnGraph.m_entParent[std::size_t(parent)];
+            parentNotNull = (parent != lgrn::id_null<ActiveEnt>());
+            TreePos_t const parentPos = parentNotNull ? rScnGraph.m_entToTreePos[std::size_t(parent)] : 0;
+            rScnGraph.m_treeDescendants[parentPos] -= removeTotal;
+        }
+
+        // Clear values for entities to delete
+        std::for_each(itTreeEntsFirst + (*itDel), itTreeEntsFirst + (*itDel + removeTotal),
```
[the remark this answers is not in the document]
jonesmz: imho: the warnings right, should cast first.

src/osp/Active/SysSceneGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+                      [&rScnGraph] (ActiveEnt const ent)
+        {
+            rScnGraph.m_entParent[std::size_t(ent)]     = lgrn::id_null<ActiveEnt>();
+            rScnGraph.m_entToTreePos[std::size_t(ent)]  = lgrn::id_null<TreePos_t>();
+        });
+
+        // Update tree positions for elements to shift
+        std::for_each(itTreeEntsFirst + keepFirst, itTreeEntsFirst + keepLast,
```
jonesmz, line 193: use std::next(itTreeEntsFirst, keepFirst)

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2020 Open Space Program Project
```
jonesmz, line 3: Probably copyright -> 2022 now.

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @brief Add a child that might have more descendants
+     *
+     * @return
+     */
+    [[nodiscard]] SubtreeBuilder add_child(ActiveEnt ent, uint32_t descendantCount);
+
+    void add_child(ActiveEnt ent)
+    {
+        (void) add_child(ent, 0);
```
jonesmz, line 63: Lol, [[nodiscard]] -- immediately discards.

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    std::size_t remaining()
+    {
+        return m_last - m_first;
```
jonesmz, line 68: minor: assert that first is not larger than last. It's a class invarient, as far as I can tell.

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+    constexpr ChildIterator& operator=(ChildIterator const& copy) noexcept = default;
+    constexpr ChildIterator& operator=(ChildIterator&& move) noexcept = default;
+
+    ChildIterator& operator++() noexcept
+    {
+        m_pos += 1 + m_pScnGraph->m_treeDescendants[m_pos];
+        return *this;
+    }
+
+    auto operator<=>(ChildIterator const&) const = default;
```
jonesmz, line 109: Minor: My implementation experience with operator<=> at work makes me recommend uisng
```
friend auto operator<=>(ChildIterator const& lhs,ChildIterator const& rhs) noexcept = default;
```

Something about the way the operator<=> auto-generates things causes some hard to comprehend compiler errors in rare situations that using "friend" makes go away.

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+
+    value_type operator*() const noexcept
```
jonesmz, line 111: Wouldn't you want operator& as well here?

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+    {
+        return m_pScnGraph->m_treeToEnt[m_pos];
+    }
+
+private:
+    ACtxSceneGraph const    * m_pScnGraph   {nullptr};
+    TreePos_t               m_pos           {0};
```
jonesmz, line 118: alignment's slightly wonky.

src/test_application/VehicleBuilder.cpp
```diff
@@ -66,12 +66,23 @@ void VehicleBuilder::set_prefabs(std::initializer_list<SetPrefab> const& setPref
...
+
+    PartId *pPartInWeld = m_data->m_weldToParts.emplace(weld, setTransform.size());
+
     for (SetTransform const& set : setTransform)
     {
-        m_data->m_partTransforms[std::size_t(set.m_part)] = set.m_transform;
+        m_data->m_partTransformWeld[set.m_part] = set.m_transform;
+
+        m_data->m_partToWeld[set.m_part] = weld;
+        (*pPartInWeld) = set.m_part;
```
jonesmz, line 81: minor: I don't know that parenthesis here really improve readability.

src/test_application/activescenes/identifiers.h
```diff
@@ -61,15 +76,11 @@
 
 #define OSP_DATA_TESTAPP_PHYSICS 3, \
     idPhys, idHierBody, idPhysIn
-#define OSP_TAGS_TESTAPP_PHYSICS 5, \
-    tgPhysBodyDel,      tgPhysBodyMod,      tgPhysBodyReq,      \
-    tgPhysMod,          tgPhysReq
-
+#define OSP_TAGS_TESTAPP_PHYSICS 6, \
+    tgPhysPrv,          tgPhysDel,          tgPhysMod,          tgPhysReq,      \
```
jonesmz, line 80: minor: Alignment of the \ characters

src/test_application/activescenes/scenarios.cpp
```diff
@@ -142,18 +142,21 @@ static ScenarioMap_t make_scenarios()
         auto &rTags             = mainView.m_rTags;
         Builder_t builder{rTags, mainView.m_rTasks, mainView.m_rTaskData};
 
-        sceneOut.resize(9);
-        auto & [scnCommon, matVisual, physics, newton, shapeSpawn, droppers, gravity, bounds, thrower] = unpack<9>(sceneOut);
+        sceneOut.resize(10);
+        auto & [scnCommon, matVisual, physics, shapeSpawn, droppers, bounds, newton, nwtGravSet, nwtGrav, shapeSpawnNwt] = unpack<10>(sceneOut);
```
jonesmz, line 146: wow that's a mouthful

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -477,22 +494,18 @@ Session setup_vehicle_spawn_vb(
...
-            {
-                continue;
-            }
-
-            auto remapPart = arrayView(rVehicleSpawnVB.m_remapPartAll).exceptPrefix(*itRemapPartOffset);
+            auto remapPart = arrayView(rVehicleSpawnVB.m_remapParts).exceptPrefix(*itRemapPartOffset);
+            std::advance(itRemapPartOffset, 1);
 
-            // remapMach[source MachAnyId] -> destination MachAnyId
             std::vector<MachAnyId> remapMach(pVData->m_machines.m_ids.capacity());
```
jonesmz, line 509: Doesn't this intantiate a bunch of values with "default" construction?

You sure that's the right constructor?

jonesmz (approved):
What's the over/under of switching newton versions to the 4.0 release series? The guy still don't understand why having two versions in the same repository instead of using branches is a stupid idea, but any time i report a bug he asks why we aren't using the new version yet.

jonesmz:
Something's not working right with the continuous integration

```
In file included from /home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp:25:
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h: At global scope:
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h:109:18: error: declaration of ‘operator<=’ as non-function
  109 |     auto operator<=>(ChildIterator const&) const = default;
      |                  ^~
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h:109:18: error: expected ‘;’ at end of member declaration
  109 |     auto operator<=>(ChildIterator const&) const = default;
      |                  ^~
      |                    ;
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h:109:20: error: expected unqualified-id before ‘>’ token
  109 |     auto operator<=>(ChildIterator const&) const = default;
      |                    ^
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp: In static member function ‘static osp::active::SubtreeBuilder osp::active::SysSceneGraph::add_descendants(osp::active::ACtxSceneGraph&, uint32_t, osp::active::ActiveEnt)’:
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp:79:14: error: ‘shift_right’ is not a member of ‘std’
   79 |         std::shift_right(rScnGraph.m_treeToEnt.begin() + subFirst,
      |              ^~~~~~~~~~~
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp:82:14: error: ‘shift_right’ is not a member of ‘std’
   82 |         std::shift_right(rScnGraph.m_treeDescendants.begin() + subFirst,
      |              ^~~~~~~~~~~
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp: In static member function ‘static void osp::active::SysSceneGraph::do_delete(osp::active::ACtxSceneGraph&)’:
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp:200:14: error: ‘shift_left’ is not a member of ‘std’
  200 |         std::shift_left(itTreeDescFirst + done, itTreeDescFirst + keepLast, shift);
      |              ^~~~~~~~~~
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.cpp:201:14: error: ‘shift_left’ is not a member of ‘std’
  201 |         std::shift_left(itTreeEntsFirst + done, itTreeEntsFirst + keepLast, shift);
      |              ^~~~~~~~~~
make[3]: *** [src/CMakeFiles/osp-magnum.dir/build.make:174: src/CMakeFiles/osp-magnum.dir/osp/Active/SysSceneGraph.cpp.o] Error 1
make[3]: *** Waiting for unfinished jobs....
```


Looks like ubuntu 18.04, and 22.04 both work, but not ubuntu 20.04.

Similarly, Mac is broken.

Can you see if you can get at least all the linux targets to work?

If you want to drop support for an older OS that's fine with me, I really don't care about anything older than $CurrentMajorRelease beyond the extra compatibility checking that we have to do to even pretend supporting older stuff.

src/newtondynamics_physics/SysNewton.cpp
Capital-Asterisk: uintptr_t just has to be large enough. Anything in the code referred to as an Id denotes some instance starting at zero and is usually a uint32. I don't consider this important enough to need an assert.

I just realized `reinterpret_cast<void*>(bodyId)` used to set the user data elsewhere likely reads some extra bytes. I guess this highlights the need for such an adapter function.

src/newtondynamics_physics/factors.h
Capital-Asterisk: The forces are assigned at runtime, I guess I should include this in the comment

src/newtondynamics_physics/ospnewton.h
Capital-Asterisk: I like this idea. No need to be too generalized imo, this seems to work too:
```cpp
template<auto FUNC_T>
struct NwtDeleter
{
    void operator() (auto const* pNwtType)
    {
        FUNC_T(pNwtType);
    }
};
```

src/osp/Active/SysPrefabInit.cpp
Capital-Asterisk: These are just spans. I often use auto for any kind of range, though I guess it's possible to return some larger types in many other cases. Would it be good to use const-ref instead?

src/osp/Active/SysPrefabInit.cpp
Capital-Asterisk: ????

what part of this doesn't make sense?

src/osp/Active/SysSceneGraph.cpp
Capital-Asterisk: I started having second thoughts about these. `itTreeEntsFirst + keepFirst` is far more readable than `std::next(itTreeEntsFirst, keepFirst)`. Can't std::next/advance/begin/end be reserved for template functions only, where the iterator types are actually at risk of not having the right operations?

src/osp/Active/SysSceneGraph.h
```diff
@@ -0,0 +1,188 @@
...
+namespace osp::active
+{
+
+/**
+ * @brief Helps add entities to a reserved space in ACtxSceneGraph
+ *
+ * All reserved space must be used, or else this class will assert on
+ * destruction.
+ */
+class SubtreeBuilder
```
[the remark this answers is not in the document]
Capital-Asterisk: As far as I'm aware, that's a fully featured graphics library? This scene graph is tailored to best fit the use cases of osp-magnum. This project is mostly just a bunch of separate components, so it's possible to use any library when the need arises. Open 3D Engine has a nice separated renderer that I think can be integrated quite well as an option.

src/osp/Active/SysSceneGraph.h
Capital-Asterisk: I'm pretty sure it's optional, and it's unsafe to modify the tree this way (unless a janky specialized handle was made?).

src/osp/Active/SysSceneGraph.h
Capital-Asterisk: memory alignment? or the formatting? I personally don't see anything wrong with either?

src/test_application/activescenes/scene_vehicles.cpp
Capital-Asterisk: This is right; I need a vector with a given size. The default values aren't used though, they technically just be garbage.

Capital-Asterisk:
> Something's not working right with the continuous integration ....

Various C++20 features, the spaceship operator specifically, just isn't supported on some of the compilers unfortunately. I recall one of them is using gcc 9.4.

jonesmz:
/home/runner/work/osp-magnum/osp-magnum/3rdparty/longeronpp/src/longeron/id_management/registry_stl.hpp:101:18: warning: attributes at the beginning of statement are ignored [-Wattributes]
  101 |             else [[likely]]
      |                  ^~~~~~~~~~

jonesmz:
/home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h:31:10: fatal error: compare: No such file or directory
   31 | #include <compare>
      |          ^~~~~~~~~

Capital-Asterisk:
> /home/runner/work/osp-magnum/osp-magnum/3rdparty/longeronpp/src/longeron/id_management/registry_stl.hpp:101:18: warning: attributes at the beginning of statement are ignored [-Wattributes]

Longeron++ issue, thinking of removing this eventually.

> /home/runner/work/osp-magnum/osp-magnum/src/osp/Active/SysSceneGraph.h:31:10: fatal error: compare: No such file or directory

C++20 not fully supported

Issue #223, opened 2022-12-06
Build fails on Clang Release with LTO


Capital-Asterisk:
```
[100%] Linking CXX executable ../Release/osp-magnum
/usr/bin/ld: /tmp/lto-llvm-ad2d68.o: in function `Magnum::Trade::MaterialAttributeData::value() const':
MaterialData.cpp:(.text._ZNK6Magnum5Trade21MaterialAttributeData5valueEv+0x31): undefined reference to `Corrade::Containers::Implementation::stringFindCharacter(char const*, unsigned long, char)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[3]: *** [src/CMakeFiles/osp-magnum.dir/build.make:717: Release/osp-magnum] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:3004: src/CMakeFiles/osp-magnum.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:3011: src/CMakeFiles/osp-magnum.dir/rule] Error 2
gmake: *** [Makefile:980: osp-magnum] Error 2
```
Magnum and Corrade are statically linked. The stringFindCharacter function is present in `libCorradeUtility.a` and is part of the link command. I tried a few things, but only disabling LTO worked.

This might have something to do with how Corrade is doing some kind of runtime CPU dispatching to optimize certain functions (stringFindCharacter specifically) depending on the machine.

PR #230, opened 2023-02-05
Fix sanitizer and Clang LTO builds, Drop CI compilers with limited C++20 support

### LINKED ISSUE #229: Continuous integration broken
Started with commit https://github.com/TheOpenSpaceProgram/osp-magnum/commit/3153dd4cd30161f65b41e012e0216324f4b1cc81, which was one of several commits in a PR, so one of those.

Capital-Asterisk:
* Fixed OSP_BUILD_SANATIZER, it simply didn't work before
* Turning OFF CORRADE_BUILD_CPU_RUNTIME_DISPATCH fixes #229 #223
* Dropped support for some compilers used in github workflows that don't support some C++20 features used.


.github/workflows/linux.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         compiler: [gcc, clang]
         config: [Release, Debug]
-        image: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04]
+        image: [ubuntu-20.04, ubuntu-22.04]
```
jonesmz, line 16: Fine with me

.github/workflows/linux.yml
```diff
@@ -43,10 +43,10 @@ jobs:
         sudo update-alternatives --verbose --install /usr/bin/g++ g++ /usr/bin/g++-10 100
 
     - name: Update Compiler (Clang)
-      if: ${{ matrix.image == 'ubuntu-18.04' && matrix.compiler == 'clang' }}
+      if: ${{ matrix.image == 'ubuntu-20.04' && matrix.compiler == 'clang' }}
       run: |
         sudo apt update
-        sudo apt install clang-10 clang++-10
+        sudo apt install clang-10
```
jonesmz, line 49: isn't clang++-10 a seperate package?

.github/workflows/linux.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         key: linux-${{ matrix.image }}-${{ matrix.compiler }}-${{ matrix.config }}
 
     - name: Update Compiler (GCC)
-      if: ${{ matrix.image == 'ubuntu-18.04' && matrix.compiler == 'gcc' }}
+      if: ${{ matrix.image == 'ubuntu-20.04' && matrix.compiler == 'gcc' }}
```
jonesmz, line 37: I don't believe we need this entire part of the matrix anymore, since the reason for it was to work around the ubuntu 18.04 image being too old.

alternatively, instead of dropping 18.04, maybe we need to change the version of clang being installed from 10 to 11?

.github/workflows/macos.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         config: [Release, Debug]
-        image: [macos-10.15, macos-11, macos-12]
```
jonesmz, line 15: fine with me

jonesmz (approved):

.github/workflows/linux.yml
Capital-Asterisk: I think they were combined? It says `clang++-10` isn't found.

.github/workflows/linux.yml
Capital-Asterisk: Both clang-10 and clang-11 are already installed. update-alternatives doesn't seem to be doing anything. I can just remove this part.

jonesmz: this specific comment is about gcc, but i said clang by mistake.

Clang 10 and clang 11 are both installed on ubuntu 18.04? that's surprising to me, since they weren't when i first set it up.

Honestly i have no real interest in seeing 18.04 supported still. that's 5 years old.

So lets first start with removing anything referring to 18.04, like you already did.

Then if more stuff can get removed, that's great too.

Capital-Asterisk:
Not sure why CodeQL Release here is failing. Process getting terminated just as it starts linking osp-magnum. It mentions PlumeShader.cpp, but that's likely a red herring as it's the first source file from osp it compiled.

PR #233, opened 2023-03-03
Re-implement Vehicle inertia and RCS Drivers

Capital-Asterisk:
Re-adding quite a bit of math that was previously removed.

## Wire/links

Fixed up to actually work properly.

Copied from a comment, referring to scene_vehicles.h/cpp

 Passing values:

1. Tasks write new values to idSigUpdFloat
2. The "Reduce Signal-Float Nodes" task reads new values from idSigUpdFloat(s) and writes them
    into idSigValFloat. This changes the input values of connected Machines, marking them dirty.
    Tags for each unique dirty machine type is added to rMachUpdEnqueue.
    Other 'reduce node' tasks could be running in parallel here.

3. The "Enqueue Machine & Node update tasks" task from setup_parts runs, and enqueues machine
    tasks from rMachUpdEnqueue as well as every tgNodeUpdEvt task, including
    "Reduce Signal-Float Nodes".

 4. Repeat until nothing is dirty

 This seemingly complex scheme allows different node types to interoperate seamlessly.

 eg. A float signal can trigger a fuel valve that triggers a pressure sensor which outputs
     another float signal, all running within a single frame.

## RCS Driver

Previously named "MachineRCSController" or something. Renamed since it conflicts with other things named 'controller'.

The previous version was aware of its own transform and center of mass of its vehicle. The newer version is dumb and is not aware of this. It's simply a function that takes takes float inputs:
* xyz Position
* xyz Direction
* xyz Linear command (translation)
* xyz Angular command (pitch/yaw/roll)
(12 total floats)
then outputs a single float throttle value.

Getting center of mass and awareness of the positions of parts within a vehicle can be handled by a separate machine. part 

## ForceFactors

Text below copied from PR #224, this time it's actually used for rockets. 

Previously, lots of separate systems would write forces into rigid bodies. Rocket thrust, gravity, and potentially aerodynamic forces are separate calls that roughly write to a single net force value for each rigid body.

This multiple-writers scheme is simply difficult to multi-thread. Atomics would help, but the problem is best avoided entirely.

Solution: Make the physics engine read from multiple sources to calculate all factors that contribute to force for each rigid body.

Each rigid body is assigned a bitset, which describe which functions are called in order to calculate its net force (and torque too). These functions are referred to as "ForceFactors." See setup_newton_force_accel(...) in scene_newton.cpp for a constant-acceleration ForceFactor used for gravity.

### DIFF
--- a/src/adera/machines/links.cpp
+++ b/src/adera/machines/links.cpp
@@ -49,7 +49,14 @@ float thruster_influence(Vector3 const pos, Vector3 const dir, Vector3 const cmd
      * In the future, it would be neat to implement PWM so that unthrottlable
      * thrusters would pulse on and off in order to deliver reduced thrust
      */
-    return std::clamp(angInfluence + linInfluence, 0.0f, 1.0f);
+    float const totalInfluence = std::clamp(angInfluence + linInfluence, 0.0f, 1.0f);
+
+    if (totalInfluence != totalInfluence) // if NaN
+    {
+        return 0.0f;
+    }
+
+    return totalInfluence;
 }
 
 
--- a/src/adera/machines/links.cpp
+++ b/src/adera/machines/links.cpp
@@ -1,3 +1,27 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2023 Open Space Program Project
+ *
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
 #include "links.h"
 
 using namespace osp;
@@ -17,46 +41,30 @@ float thruster_influence(Vector3 const pos, Vector3 const dir, Vector3 const cmd
     using Magnum::Math::cross;
     using Magnum::Math::dot;
 
-    // Thrust is applied in opposite direction of thruster nozzle direction
-    Vector3 const thrustDir = std::negate{}(dir.normalized());
+    float influence = 0.0f;
 
-    float angInfluence = 0.0f;
     if (cmdAng.dot() > 0.0f)
     {
-        Vector3 const torque = cross(pos, thrustDir).normalized();
-        angInfluence = dot(torque, cmdAng.normalized());
+        Vector3 const torque = cross(pos, dir).normalized();
+        influence += dot(torque, cmdAng.normalized());
     }
 
-    float linInfluence = 0.0f;
     if (cmdLin.dot() > 0.0f)
     {
-        linInfluence = dot(thrustDir, cmdLin.normalized());
+        influence += dot(dir, cmdLin.normalized());
     }
 
-    // Total component of thrust in direction of command
-    float const total = angInfluence + linInfluence;
-
-    if (total < 0.01f)
+    if (influence < 0.01f)
     {
-        /* Ignore small contributions from imprecision
-         * Real thrusters can't throttle this deep anyways, so if their
-         * contribution is this small, it'd be a waste of fuel to fire them.
-         */
-        return 0.0f;
+        return 0.0f; // Ignore small contributions
     }
 
-    /* Compute thruster throttle output demanded by current command.
-     * In the future, it would be neat to implement PWM so that unthrottlable
-     * thrusters would pulse on and off in order to deliver reduced thrust
-     */
-    float const totalInfluence = std::clamp(angInfluence + linInfluence, 0.0f, 1.0f);
-
-    if (totalInfluence != totalInfluence) // if NaN
+    if (Magnum::Math::isNan(influence))
     {
         return 0.0f;
     }
 
-    return totalInfluence;
+    return std::clamp(influence, 0.0f, 1.0f);
 }
 
 
--- a/src/adera/machines/links.cpp
+++ b/src/adera/machines/links.cpp
@@ -1,12 +1,56 @@
 #include "links.h"
 
+using namespace osp;
+
 using osp::link::MachTypeReg_t;
 using osp::link::MachTypeId;
 
 namespace adera
 {
 
-MachTypeId const gc_mtUserCtrl = osp::link::MachTypeReg_t::create();
-MachTypeId const gc_mtMagicRocket = osp::link::MachTypeReg_t::create();
+MachTypeId const gc_mtUserCtrl      = MachTypeReg_t::create();
+MachTypeId const gc_mtMagicRocket   = MachTypeReg_t::create();
+MachTypeId const gc_mtRcsDriver     = MachTypeReg_t::create();
+
+float thruster_influence(Vector3 const pos, Vector3 const dir, Vector3 const cmdLin, Vector3 const cmdAng) noexcept
+{
+    using Magnum::Math::cross;
+    using Magnum::Math::dot;
+
+    // Thrust is applied in opposite direction of thruster nozzle direction
+    Vector3 const thrustDir = std::negate{}(dir.normalized());
+
+    float angInfluence = 0.0f;
+    if (cmdAng.dot() > 0.0f)
+    {
+        Vector3 const torque = cross(pos, thrustDir).normalized();
+        angInfluence = dot(torque, cmdAng.normalized());
+    }
+
+    float linInfluence = 0.0f;
+    if (cmdLin.dot() > 0.0f)
+    {
+        linInfluence = dot(thrustDir, cmdLin.normalized());
+    }
+
+    // Total component of thrust in direction of command
+    float const total = angInfluence + linInfluence;
+
+    if (total < .01f)
+    {
+        /* Ignore small contributions from imprecision
+         * Real thrusters can't throttle this deep anyways, so if their
+         * contribution is this small, it'd be a waste of fuel to fire them.
+         */
+        return 0.0f;
+    }
+
+    /* Compute thruster throttle output demanded by current command.
+     * In the future, it would be neat to implement PWM so that unthrottlable
+     * thrusters would pulse on and off in order to deliver reduced thrust
+     */
+    return std::clamp(angInfluence + linInfluence, 0.0f, 1.0f);
+}
+
 
 } // namespace adera
--- a/src/adera/machines/links.cpp
+++ b/src/adera/machines/links.cpp
@@ -36,7 +36,7 @@ float thruster_influence(Vector3 const pos, Vector3 const dir, Vector3 const cmd
     // Total component of thrust in direction of command
     float const total = angInfluence + linInfluence;
 
-    if (total < .01f)
+    if (total < 0.01f)
     {
         /* Ignore small contributions from imprecision
          * Real thrusters can't throttle this deep anyways, so if their
--- a/src/adera/machines/links.h
+++ b/src/adera/machines/links.h
@@ -52,6 +52,8 @@ using osp::link::gc_ntSigFloat;
 using osp::link::gc_sigIn;
 
 PortEntry const gc_throttleIn       { gc_ntSigFloat, 0, gc_sigIn };
+PortEntry const gc_multiplierIn     { gc_ntSigFloat, 1, gc_sigIn };
+
 }
 
 } // namespace adera
--- a/src/adera/machines/links.h
+++ b/src/adera/machines/links.h
@@ -39,6 +39,8 @@ extern osp::link::MachTypeId const gc_mtUserCtrl;
 extern osp::link::MachTypeId const gc_mtMagicRocket;
 extern osp::link::MachTypeId const gc_mtRcsDriver;
 
+constexpr osp::Vector3 gc_rocketForward{0.0f, 0.0f, 1.0f};
+
 namespace ports_userctrl
 {
 PortEntry const gc_throttleOut      { gc_ntSigFloat, 0, gc_sigOut };
--- a/src/adera/machines/links.h
+++ b/src/adera/machines/links.h
@@ -30,15 +30,17 @@
 namespace adera
 {
 
+using osp::link::PortEntry;
+using osp::link::gc_ntSigFloat;
+using osp::link::gc_sigIn;
+using osp::link::gc_sigOut;
+
 extern osp::link::MachTypeId const gc_mtUserCtrl;
 extern osp::link::MachTypeId const gc_mtMagicRocket;
+extern osp::link::MachTypeId const gc_mtRcsDriver;
 
 namespace ports_userctrl
 {
-using osp::link::PortEntry;
-using osp::link::gc_ntSigFloat;
-using osp::link::gc_sigOut;
-
 PortEntry const gc_throttleOut      { gc_ntSigFloat, 0, gc_sigOut };
 PortEntry const gc_pitchOut         { gc_ntSigFloat, 1, gc_sigOut };
 PortEntry const gc_yawOut           { gc_ntSigFloat, 2, gc_sigOut };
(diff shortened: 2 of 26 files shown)


src/adera/machines/links.h
```diff
@@ -47,11 +51,27 @@ PortEntry const gc_rollOut          { gc_ntSigFloat, 3, gc_sigOut };
...
+PortEntry const gc_dirYIn           { gc_ntSigFloat, 4, gc_sigIn };
+PortEntry const gc_dirZIn           { gc_ntSigFloat, 5, gc_sigIn };
+PortEntry const gc_cmdLinXIn        { gc_ntSigFloat, 6, gc_sigIn };
+PortEntry const gc_cmdLinYIn        { gc_ntSigFloat, 7, gc_sigIn };
+PortEntry const gc_cmdLinZIn        { gc_ntSigFloat, 8, gc_sigIn };
+PortEntry const gc_cmdAngXIn        { gc_ntSigFloat, 9, gc_sigIn };
+PortEntry const gc_cmdAngYIn        { gc_ntSigFloat, 10, gc_sigIn };
+PortEntry const gc_cmdAngZIn        { gc_ntSigFloat, 11, gc_sigIn };
+PortEntry const gc_throttleOut      { gc_ntSigFloat, 12, gc_sigOut };
+}
```
Capital-Asterisk, line 73: There was previously specific types for these, such as "AttitudeControl" and "Percent", but honestly that just made things harder.

src/osp/Active/SysPhysics.cpp
```diff
@@ -1,2 +1,72 @@
...
+void SysPhysics::calculate_subtree_mass_center(
+        acomp_storage_t<ACompTransform> const&  rTf,
+        ACtxPhysics&                            rCtxPhys,
+        ACtxSceneGraph&                         rScnGraph,
+        ActiveEnt                               root,
+        Vector3&                                rMassPos,
+        float&                                  rTotalMass,
+        Matrix4 const&                          currentTf)
+{
+    for (ActiveEnt const child : SysSceneGraph::children(rScnGraph, root))
```
Capital-Asterisk, line 16: Cool thing about the hierarchy here is that subtrees are contiguous int arrays. This recursive function is actually a linear array iteration.

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -187,44 +258,102 @@ Session setup_mach_rocket(
...
+            "RCS Drivers calculate new values",
+            TopDataIds_t{           idScnParts,                      idUpdMach,                       idSigValFloat,                    idSigUpdFloat },
+            wrap_args([] (ACtxParts& rScnParts, UpdMachPerType const& rUpdMach, SignalValues_t<float>& rSigValFloat, UpdateNodes<float>& rSigUpdFloat) noexcept
     {
         Nodes const &rFloatNodes = rScnParts.m_nodePerType[gc_ntSigFloat];
-        PerMachType &rRockets = rScnParts.m_machines.m_perType[gc_mtMagicRocket];
+        PerMachType &rRockets = rScnParts.m_machines.m_perType[gc_mtRcsDriver];
 
-        for (MachLocalId const local : rUpdMach.m_localDirty[gc_mtMagicRocket].ones())
+        for (MachLocalId const local : rUpdMach.m_localDirty[gc_mtRcsDriver].ones())
```
Capital-Asterisk, line 298: RCS driver code starts here. This loop iterates all RCS drivers currently marked dirty. It reads all 12 connected floats, then uses z-adam's thruster_influence function to calculate a throttle output.

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -276,79 +407,85 @@ Session setup_vehicle_spawn_vb(
...
 
         std::size_t partTotal       = 0;
         std::size_t remapPartTotal  = 0;
 
         std::size_t weldTotal       = 0;
         std::size_t remapWeldTotal  = 0;
 
+        for (NewVehicleId vhId = 0; vhId < newVehicleCount; ++vhId)
         {
-            auto itRemapPartOffsetsOut  = std::begin(rVehicleSpawnVB.m_remapPartOffsets);
```
Capital-Asterisk, line 316: Replaced quite a few iterators to use index-access instead. It became extremely confusing, noisy, and generally poor for my sanity when it's impossible to tell which containers the iterators are writing to.

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -580,38 +979,70 @@ Session setup_test_vehicles(
 
     // Build "PartVehicle"
```
Capital-Asterisk, line 980: Vehicle building code starts here. Basically, this generates a VehicleData struct, which is then read by some tasks elsewhere in this source file that copies it to the scene based on vehicle spawn requests from scenarios.cpp.

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
+
 using osp::link::MachTypeReg_t;
 using osp::link::MachTypeId;
 
 namespace adera
 {
 
-MachTypeId const gc_mtUserCtrl = osp::link::MachTypeReg_t::create();
-MachTypeId const gc_mtMagicRocket = osp::link::MachTypeReg_t::create();
+MachTypeId const gc_mtUserCtrl      = MachTypeReg_t::create();
```
jonesmz, line 11: Are these supposed to be accessible outside of this translation unit?

If not, best to move them to an anonymous namespace and make them static, to communicate to the linker that they can't escape the translation unit other than via being returned as a pointer/reference.

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
+    Vector3 const thrustDir = std::negate{}(dir.normalized());
+
+    float angInfluence = 0.0f;
+    if (cmdAng.dot() > 0.0f)
+    {
+        Vector3 const torque = cross(pos, thrustDir).normalized();
+        angInfluence = dot(torque, cmdAng.normalized());
+    }
+
+    float linInfluence = 0.0f;
```
jonesmz, line 30: minor: You could use a trinary conditional on these so that linInfluence is const.

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
+         * contribution is this small, it'd be a waste of fuel to fire them.
+         */
+        return 0.0f;
+    }
+
+    /* Compute thruster throttle output demanded by current command.
+     * In the future, it would be neat to implement PWM so that unthrottlable
+     * thrusters would pulse on and off in order to deliver reduced thrust
+     */
+    float const totalInfluence = std::clamp(angInfluence + linInfluence, 0.0f, 1.0f);
```
jonesmz, line 52: i recommend re-using "total" here instead of adding twice.

Obviously an optimizer should catch this, but it'd help with casual readers of the code as well.

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
+
+    if (totalInfluence != totalInfluence) // if NaN
```
jonesmz, line 54: https://en.cppreference.com/w/cpp/numeric/math/isnan

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
 using osp::link::MachTypeId;
 
 namespace adera
 {
 
-MachTypeId const gc_mtUserCtrl = osp::link::MachTypeReg_t::create();
-MachTypeId const gc_mtMagicRocket = osp::link::MachTypeReg_t::create();
+MachTypeId const gc_mtUserCtrl      = MachTypeReg_t::create();
+MachTypeId const gc_mtMagicRocket   = MachTypeReg_t::create();
+MachTypeId const gc_mtRcsDriver     = MachTypeReg_t::create();
```
jonesmz, line 13: Should these go into a headerfile as "inline" ?

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
 #include "links.h"
```
jonesmz, line 1: Copyright block?

src/adera/machines/links.h
```diff
@@ -47,11 +51,27 @@ PortEntry const gc_rollOut          { gc_ntSigFloat, 3, gc_sigOut };
...
-using osp::link::gc_ntSigFloat;
-using osp::link::gc_sigIn;
-
 PortEntry const gc_throttleIn       { gc_ntSigFloat, 0, gc_sigIn };
+PortEntry const gc_multiplierIn     { gc_ntSigFloat, 1, gc_sigIn };
 }
 
+namespace ports_rcsdriver
+{
+PortEntry const gc_posXIn           { gc_ntSigFloat, 0, gc_sigIn };
```
jonesmz, line 60: Can't these be constexpr?

If not constexpr, perhaps constinit?

src/adera/machines/links.cpp
```diff
@@ -1,12 +1,63 @@
...
 namespace adera
 {
 
-MachTypeId const gc_mtUserCtrl = osp::link::MachTypeReg_t::create();
-MachTypeId const gc_mtMagicRocket = osp::link::MachTypeReg_t::create();
+MachTypeId const gc_mtUserCtrl      = MachTypeReg_t::create();
+MachTypeId const gc_mtMagicRocket   = MachTypeReg_t::create();
+MachTypeId const gc_mtRcsDriver     = MachTypeReg_t::create();
+
+float thruster_influence(Vector3 const pos, Vector3 const dir, Vector3 const cmdLin, Vector3 const cmdAng) noexcept
```
jonesmz, line 15: My reading of this function is that it's a constexpr candidate.

If not, consider https://stackoverflow.com/questions/29117836/attribute-const-vs-attribute-pure-in-gnu-c

`__attribute((const))`

since it only reads it's inputs and then returns a single output.

src/osp/Active/SysPhysics.h
```diff
@@ -35,22 +36,32 @@ class SysPhysics
...
+            ACtxPhysics&                            rCtxPhys,
+            ACtxSceneGraph&                         rScnGraph,
+            ActiveEnt                               root,
+            Matrix3&                                rInertiaTensor,
+            Matrix4 const&                          currentTf = {});
 
     template<typename IT_T>
-    static void update_delete_phys(
-            ACtxPhysics &rCtxPhys, IT_T first, IT_T const& last);
+    static void update_delete_phys(ACtxPhysics& rCtxPhys, IT_T const& first, IT_T const& last);
```
jonesmz, line 57: consider changing the template type for the iterators to have a separate type for the first vs the last, to allow for the C++20 "sentinel" iterator feature to work.

src/osp/Active/SysPrefabInit.cpp
```diff
@@ -196,36 +197,77 @@ void SysPrefabInit::init_physics(
         auto const &rPrefabData = rResources.data_get<osp::Prefabs const>(
                 gc_importer, rPfBasic.m_importerRes);
 
-        auto const objects = rPrefabData.m_prefabs[rPfBasic.m_prefabId];
+        ArrayView<ActiveEnt const>  ents        = *itPfEnts;
+        lgrn::Span<int const>       objects     = rPrefabData.m_prefabs[rPfBasic.m_prefabId];
+        lgrn::Span<int const>       parents     = rPrefabData.m_prefabParents[rPfBasic.m_prefabId];
```
jonesmz, line 202: these spans / views could be, themselves, const.

src/osp/link/machines.h
```diff
@@ -90,10 +83,14 @@ struct UpdMachPerType
...
+struct MachinePair
 {
     MachLocalId     m_local{lgrn::id_null<MachLocalId>()};
     MachTypeId      m_type{lgrn::id_null<MachTypeId>()};
+};
+
+struct Junction : MachinePair
+{
     JuncCustom      m_custom{0};
 };
```
jonesmz, line 95: MachinePair is 32bit + 16bit with an align of 32bit, so it has 16bit of padding.

Does Junction inherit from MachinePair in such a way that sizeof(Junction) == 64bit ? Or is it tacking on JuncCustom (16bit) after the padding from MachinePair, giving you a size of 96bit ?

I recommend a static assert on the sizeof(Junction) to make sure the compiler is behaving itself.

src/test_application/VehicleBuilder.cpp
```diff
@@ -234,13 +240,43 @@ VehicleData VehicleBuilder::finalize_release()
...
+        rData.m_partToMachines.emplace(part, m_partMachCount[part]);
+    }
+
+    // Assign part-to-machine partitions
+    for (MachAnyId const mach : rData.m_machines.m_ids.bitview().zeros())
+    {
+        MachLocalId const local = rData.m_machines.m_machToLocal[mach];
+        MachTypeId const  type  = rData.m_machines.m_machTypes[mach];
+        PartId const      part  = rData.m_machToPart[mach];
+        lgrn::Span<MachinePair> machines = rData.m_partToMachines[part];
```
jonesmz, line 263: the span itself is constant, it's only the things that it points to that mutate

src/test_application/VehicleBuilder.cpp
```diff
@@ -234,13 +240,43 @@ VehicleData VehicleBuilder::finalize_release()
...
+        // Reuse machine count to track how many are currently added.
+        // By the end, these should all be zero
+        auto &rPartMachCount = m_partMachCount[part];
+        assert(rPartMachCount != 0);
+        -- rPartMachCount;
+
+        machines[rPartMachCount] = { .m_local = local, .m_type = type };
+    }
+
+    assert(std::all_of(m_partMachCount.begin(), m_partMachCount.end(),
```
jonesmz, line 274: assert(std::ranges::all_of(m_partMachCount, [](auto const& count){return 0 == count;}));

src/test_application/activescenes/scenarios.cpp
```diff
@@ -191,27 +191,28 @@ static ScenarioMap_t make_scenarios()
         auto &rTags             = mainView.m_rTags;
         Builder_t builder{rTags, mainView.m_rTasks, mainView.m_rTaskData};
 
-        sceneOut.resize(21);
+        sceneOut.resize(24);
```
jonesmz, line 194: i'd say move this constant 24 out to a static constexpr variable

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -558,9 +849,117 @@ Session setup_vehicle_spawn_vb(
...
+        for (NewVehicleId vhId = 0; vhId < newVehicleCount; ++vhId)
+        {
+            VehicleData const* pVData = rVSVB.m_dataVB[vhId];
+            if (pVData == nullptr)
+            {
+                continue;
+            }
+
+            PerNodeType const&  srcFloatNodes       = pVData->m_nodePerType[gc_ntSigFloat];
+            entt::any const     srcFloatValuesAny   = srcFloatNodes.m_nodeValues;
```
jonesmz, line 876: making a copy ?

src/test_application/activescenes/scene_vehicles.cpp
```diff
@@ -580,38 +979,70 @@ Session setup_test_vehicles(
...
+        float const rcsRadius       = 1.1f;
+        float const rcsThrust       = 3000.0f;
+
+        for (int ring = 0; ring < rcsRingCount; ++ring)
+        {
+            Vector3 const rcsOset{rcsRadius, 0.0f, rcsRingZ + ring*rcsZStep };
+
+            for (Rad ang = 0.0_degf; ang < Rad(360.0_degf); ang += Rad(360.0_degf)/rcsRingBlocks)
+            {
+               Quaternion rotZ = Quaternion::rotation(ang, {0.0f, 0.0f, 1.0f});
```
jonesmz, line 1038: const

jonesmz (approved):

jonesmz:
I did approve the pullrequest, but best not to submit it until the continuous integration is passing.

We already dropped all the older operating systems because of C++20 partial support, i'd rather not drop entire platforms.

src/adera/machines/links.h
Capital-Asterisk: gc_ntSigFloat is dynamically initialized. I'm having a difficult time telling whether there is any ordering issues here, address sanitizer doesn't seem to mind though.

src/adera/machines/links.cpp
Capital-Asterisk: It can. I vaguely recall there was something similar to this a long time ago that was originally inline, then you suggested changing it to extern.

Either way, I'd assume there has to be some minor advantages to being declared in just one translation unit.

src/adera/machines/links.cpp
Capital-Asterisk: thruster_influence requires a square root, so constexpr isn't possible (or just very difficult). `__attribute__((const))` seems really obscure, and I have doubts it would make any difference at this time; also noting LTO.

src/osp/link/machines.h
Capital-Asterisk: In this case it's just easier to copy the members; this eliminates the problem.

src/osp/Active/SysPrefabInit.cpp
Capital-Asterisk: Others comment that the amount of `const`s needed in C++ is getting silly. I did started using this style in other places; I think this is more readable:
```cpp
auto const ents     = ArrayView<ActiveEnt const>{*itPfEnts};
auto const objects  = lgrn::Span<int const>     {rPrefabData.m_prefabs[rPfBasic.m_prefabId]};
auto const parents  = lgrn::Span<int const>     {rPrefabData.m_prefabParents[rPfBasic.m_prefabId]};
```

src/test_application/VehicleBuilder.cpp
Capital-Asterisk: From recent experiences, I think I'd rather refrain from using any std::ranges functions.

src/test_application/activescenes/scenarios.cpp
Capital-Asterisk: This number is related to the variables declared the very next line. In this case, moving the constant elsewhere hurts readability, though it seems best for me to just wrap the two lines into a template function.

PR #235, opened 2023-05-24
Add git hash

### LINKED ISSUE #85: Add git commit hashes to build
https://github.com/andrew-hardin/cmake-git-version-tracking

EhWhoAmI:
Fixes #85

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,14 @@ if(NOT MSVC)
     add_compile_options(-fno-omit-frame-pointer)
 endif()
 
+include(FetchContent)
+
+FetchContent_Declare(cmake_git_version_tracking                   
+  GIT_REPOSITORY https://github.com/andrew-hardin/cmake-git-version-tracking.git
+  GIT_TAG 904dbda1336ba4b9a1415a68d5f203f576b696bb
+)
+FetchContent_MakeAvailable(cmake_git_version_tracking)
+
 # Compiler warnings can help find problems in code.
 IF(OSP_ENABLE_COMPILER_WARNINGS)
   # Enable additional warnings.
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -66,6 +66,14 @@ if(NOT MSVC)
...
 endif()
 
+include(FetchContent)
+
+FetchContent_Declare(cmake_git_version_tracking                   
+  GIT_REPOSITORY https://github.com/andrew-hardin/cmake-git-version-tracking.git
+  GIT_TAG 904dbda1336ba4b9a1415a68d5f203f576b696bb
+)
+FetchContent_MakeAvailable(cmake_git_version_tracking)
+
```
Capital-Asterisk, line 76: Can you try using a submodule instead? This is for consistency with how all other libraries are added. It can be put it in the 3rdparty directory, and added using add_subdirectory. the other dependencies work as examples.

Capital-Asterisk: Something I'm curious/concerned about is how this library selects which git repository to read info from. Most likely, it won't select itself. It somehow has to obtain the parent that added it.

EhWhoAmI: Alright I'll do some testing and see how that works out

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "gitlens.currentLine.enabled": false
+}
```
jonesmz, line 3: generally you should prefer having a trailing newline, to prevent tools like github from giving you angry red circles.

src/test_application/main.cpp
```diff
@@ -64,6 +64,14 @@
 #include <string_view>
 #include <thread>
 #include <unordered_map>
+#if __has_include("git.h")
+#include "git.h"
```
jonesmz, line 68: Why wouldn't it have this header?

PR #245, opened 2023-09-17
Fix Universe test scene

Capital-Asterisk:
adds stuff

## Move DrawEnts to renderer

The universe test scene, cursor, and other stuff actually call OpenGL-specific functions and do a bit of heavy lifting when all they really need to do is 'put a sphere at xyz', motivating the 'phase-2' of separating out DrawEnts from ActiveEnt.

The idea of "Draw Entities" came around when separating the concerns the common game engine pattern of "representing the world using a tree of entities, nodes, or game objects". This causes spaghetti as the tree handles too much of everything. Things intended only for visual effect can overlap and interfere with other game logic. Unfortunately, I was following this pattern because I didn't know any better at the time, leading to a tree of ActiveEnts representing everything.

DrawEnts were separated out from ActiveEnt back in #234, but the DrawEnts still resided as scene state. Nothing was changed structurally. Now, it's moved to its own 'scene renderer' session. This means all DrawEnts are deleted when the window closes, and must be regenerated from existing scene state when re-opened.

DrawEnts evolved to basically describes an API-agnostic drawable 3d scene; a retained-mode graphics library or additional layer of indirection. It uses the same 'ECS-inspired' pattern of each DrawEnt being a unique integer starting from zero, used as an index to many separate containers.

This completely cuts most or all scene-renderer coupling (I feel like I write this before but that was a joke), which means in theory, multiple windows/renderers can now be added simultaneously.

In order to draw things from the scene, the scene state needs to be synchronized with DrawEnts (this was already the case before). The shape spawning system, for example, has the renderer session `setup_shape_spawn_draw()` which basically goes:

```
for (each shape spawned) { make a DrawEnt for it }
```

Additionally, DrawEnts entirely optional. It's still possible to write tasks that call API-specific functions:

```
for (each shape spawned) { call OpenGL draw functions }
```

## Universe Test scene

It looks cool. Same one from #204 but updated for the current architecture.

Screenshot_20230917_143155

## Other changes

* Increased power of the shape thrower to throw 25 spheres at the same time. this is more fun.
* Added bigger and more interesting floor for the test scenes
* Updated Flat and Phong shaders for the new architecture
* Re-add cursor session. This creates a green wireframe cube to indicate the center of where the camera is orbiting around in test scenes. It uses DrawEnts.

### DIFF
--- a/src/osp/Active/drawing.h
+++ b/src/osp/Active/drawing.h
@@ -73,7 +73,6 @@ using TexIdOwner_t      = TexRefCount_t::Owner_t;
  */
 struct ACtxDrawing
 {
-
     // Scene-space Meshes
     lgrn::IdRegistryStl<MeshId>             m_meshIds;
     MeshRefCount_t                          m_meshRefCounts;
@@ -104,7 +103,7 @@ struct ACtxDrawingRes
 
 using DrawEntColors_t = KeyedVec<DrawEnt, Magnum::Color4>;
 using DrawEntTextures_t = KeyedVec<DrawEnt, TexIdOwner_t>;
-
+using DrawTransforms_t = KeyedVec<DrawEnt, Matrix4>;
 
 struct ACtxSceneRender
 {
@@ -121,9 +120,15 @@ struct ACtxSceneRender
         bitvector_resize(m_transparent, size);
         bitvector_resize(m_visible,     size);
 
+        m_drawTransform .resize(size);
         m_color         .resize(size, {1.0f, 1.0f, 1.0f, 1.0f}); // Default white
         m_diffuseTex    .resize(size);
         m_mesh          .resize(size);
+
+        for (uint32_t matInt : m_materialIds.bitview().zeros())
+        {
+            bitvector_resize(m_materials[MaterialId(matInt)].m_ents, size);
+        }
     }
 
     void resize_active(std::size_t const size)
@@ -141,6 +146,8 @@ struct ACtxSceneRender
 
     DrawEntSet_t                            m_needDrawTf;
     KeyedVec<ActiveEnt, DrawEnt>            m_activeToDraw;
+
+    DrawTransforms_t                        m_drawTransform;
 
     // Meshes and textures assigned to DrawEnts
     KeyedVec<DrawEnt, TexIdOwner_t>         m_diffuseTex;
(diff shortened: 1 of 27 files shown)


src/osp/Active/drawing.h
```diff
@@ -80,16 +73,62 @@ using TexIdOwner_t      = TexRefCount_t::Owner_t;
...
+/**
+ * @brief Associates mesh/texture resources Ids from ACtxDrawing with Resources
+ */
+struct ACtxDrawingRes
+{
+    // Required for std::is_copy_assignable to work properly inside of entt::any
+    ACtxDrawingRes() = default;
+    ACtxDrawingRes(ACtxDrawingRes const& copy) = delete;
+    ACtxDrawingRes(ACtxDrawingRes&& move) = default;
+
```
jonesmz, line 94: not also defining operator= ?

src/osp/Shaders/Flat.h
```diff
@@ -69,75 +68,60 @@ void draw_ent_flat(
...
+    {
+        auto value = (hasMaterial && args.opaque.test(entInt))
+                   ? std::make_optional(EntityToDraw{&draw_ent_flat, {&args.rData, pShader}})
+                   : std::nullopt;
+
+        storage_assign(*args.pStorageOpaque, ent, std::move(value));
     }
 }
 
+template<typename ITA_T, typename ITB_T>
```
jonesmz, line 114: minor: since we're C++20, you could use the C++20 template thing `std::sentinal_for<ITA_T> ITB_T` over `typename`

jonesmz (approved):

src/osp/Active/drawing.h
Capital-Asterisk: Noted. At the moment this is harmless as no assignments are done.

I'm planning on reorganizing and cleaning the code on the next PR, as functions are a little scattered after all the rewriting and it's hard to navigate. I'll probably add a "move-only" macro for this.

src/osp/Shaders/Flat.h
Capital-Asterisk: This looks rather noisy for and gets passed to for_each anyways.

I'd actually prefer to limit the use of C++20, especially ranges. Consider that there's relatively few C++ developers in the gamedev world, and even fewer of them know C++20. It's already difficult enough to get new developers on this project.

Capital-Asterisk:
Thinking of using a struct with an int instead of an enum class for ID types so they can have a default value and never be uninitialized.

PR #246, opened 2023-09-24
Reorganize directory structure and other clean-ups

Capital-Asterisk:
It was getting difficult to navigate as lots of code was in the wrong place. This reorganizes the directory structure and moves some code around to stay sane.

The `src/osp` directory now has 10 subdirectories: `core, tasks, universe, scientific, vehicles, link, activescene, drawing, util, drawing_gl` which depend on each other in a clean way:
ospdirs
`src/osp`, `src/adera`, and `src/ospnewton` is used 'as a library' by the test application `src/testapp` to make the executable.

Other important changes:
* Added "StrongId" class to replace enum classes used as 'strong typedefs for Ids. These have the benefit of being default-initialized to their respective 'null' values instead of zero or undefined. 
* Added copy/move macros as shorthands for defining copy/move constructors and assignment operators


src/osp/core/strong_id.h
```diff
@@ -0,0 +1,89 @@
...
+    }
+};
+
+// Longeron++ underlying_int_type
+template<typename TYPE_T>
+struct lgrn::underlying_int_type< TYPE_T, std::enable_if_t< std::is_integral_v<typename TYPE_T::entity_type> > >
+{
+    using type = typename TYPE_T::entity_type;
+};
+
```
Capital-Asterisk, line 89: The new StrongId class

src/osp/core/strong_id.h
```diff
@@ -0,0 +1,90 @@
...
```
Capital-Asterisk, line 90: New StrongId class. Adding move constructors breaks on msvc for some reason, but move constructors don't really have an effect anyways.

jonesmz (approved):

PR #247, opened 2023-10-29
Re-add vehicles again

Capital-Asterisk:
This 100% finishes the pipeline rewrite and fixes everything commented-out in #244
* Update all the old task functions to the new Pipeline one
* Re-Add Signal Links (wiring). It just works and is far more elegant than what was previously there due to the Pipeline nested loop feature
* Re-Add Machines: Rockets, RCS drivers, User Controls
* Add 'Thrust Indicators' to indicate which rockets are firing. This required adding 'Draw Transform Observers'
* Split code into different files

### DrawTransformObservers

from drawing_fn.h:

```cpp
/**
 * @brief Function pointers called when new draw transforms are calculated
 *
 * Draw transforms (Matrix4) are calculated by traversing the Scene graph (tree of ActiveEnts).
 * These matrices are not always stored in memory since they're slightly expensive. By default,
 * they are only saved for DrawEnts associated with an ActiveEnt in ACtxSceneRender::activeToDraw.
 *
 * Draw transforms can be calculated by SysRender::update_draw_transforms, or potentially by a
 * future system that takes physics engine interpolation or animations into account.
 * DrawTfObservers provides a way to tap into this procedure to call custom functions for other
 * systems.
 *
 * To use, write into DrawTfObservers::observers[i]
 * Enable per-DrawEnt by setting ACtxSceneRender::drawTfObserverEnable[drawEnt] bit [i]
 *
 */
struct DrawTfObservers
```

### Other notes

One may notice that in many places I'm dropping the `m_` naming convention for structs that don't have (or have very few) member functions. The `m_` is useful for member functions as C++'s implicit `this->` can make it difficult to tell which variables are or aren't member variables. If there are no member functions, then they're redundant and contribute to making my eyes bleed as I try to understand what the code is doing, so they should be removed.


jonesmz:
Well that's an enormous diff.

I reviewed it as best i could and didn't see anything glaring. But no promises that i didn't overlook something.

jonesmz:
if you rebase on master, everything in the CI should work except for the windows analyzer.

jonesmz:
Ok now CI is 100% green.

jonesmz:
You can see the code scanning results here: https://github.com/TheOpenSpaceProgram/osp-magnum/security/code-scanning

Capital-Asterisk:
Thanks for the CI fixes!

PR #253, opened 2023-10-31
Various C++20 tweaks

jonesmz:


src/adera/activescene/VehicleBuilder.cpp
```diff
@@ -101,7 +103,7 @@ void VehicleBuilder::index_prefabs()
             continue; // No prefab data
         }
 
-        for (osp::PrefabId j = 0; j < pPrefabData->m_prefabNames.size(); ++j)
+        for (osp::PrefabId const j : std::views::iota(0u, pPrefabData->m_prefabNames.size()))
```
Capital-Asterisk, line 106: I don't believe there's anything wrong with the original for loop. It's more modern? but no improvements in type safety or anything, slightly longer compile times, and probably confuses new contributes or less experienced programmers.

I'd avoid the ranges if other parts of the library are not being used (and yes I probably have a bit of a grudge against it as it made of my other projects a huge headache, and also Rust exists). I feel that ranges is something we should give a few more years to mature before using seriously.

Also noting that many of these identifiers are less of 'numbers', and are more of 'identifiers with the special property of being sequential'. PrefabId is planned to eventually be changed to StrongId<...>, so iota probably wouldn't work.

src/adera/activescene/VehicleBuilder.h
```diff
@@ -133,11 +135,11 @@ class VehicleBuilder
     void set_prefabs(std::initializer_list<SetPrefab> const& setPrefab);
 
 
-    WeldId weld(osp::ArrayView<PartToWeld const> toWeld);
+    WeldId weld(std::span<PartToWeld const> toWeld);
```
Capital-Asterisk, line 138: I chose Corrade ArrayView over std::span as it features more human-friendly and easier to read methods for making 'subspans'. eg: methods like prefix() and suffix().

ArrayView also supports constructing from std::initializer_list, which is relied upon in a few places here, likely on this line too, which probably causes a build break.

I remember trying to help someone with their C++ project in a voice call one time, explaining std::initializer_list, std::array, and c-style array since they didn't know which one to use. They did not want to do anymore C++ shortly afterwards.

src/adera/drawing_gl/phong_shader.h
```diff
@@ -110,11 +112,11 @@ inline void sync_drawent_phong(osp::draw::DrawEnt ent, ArgsForSyncDrawEntPhong c
...
+template<std::input_iterator IT_T, std::sentinel_for<IT_T> SENT_T>
 void sync_drawent_phong(
-        ITA_T const&                    first,
-        ITB_T const&                    last,
-        ArgsForSyncDrawEntPhong const   args)
+        IT_T const&                   first,
+        SENT_T const&                 last,
+        ArgsForSyncDrawEntPhong const args)
 {
     std::for_each(first, last, [&args] (osp::draw::DrawEnt const ent)
```
Capital-Asterisk, line 121: One of the things I've been considering, is that there is a very low to zero chance that this function will ever be used with sentinels. I don't believe it's worth the extra typing and additional C++ knowledge required to understand its purpose. Regular std::for_each also doesn't support sentinels.

Capital-Asterisk:
I'm not sure if there's a noticeable change in my views, but I'd like to enforce "no \<ranges\>" for the time being. It's...
* something I personally don't want to work with. I don't like using authority, but I'm the one writing the majority of the code.
* doesn't fulfill any 'requirements' nor improves the code (besides ranges compatibility itself)
* sets the bar quite high for required C++ knowledge for potential contributors. The project architecture is already kind of weird, which I believe is a requirement to represent the rocket-science stuff with the least moving parts. Unnecessarily advance C++ will make explaining it even more difficult.

Many other C++20 features, like concepts, I'm perfectly cool with.

src/adera/activescene/VehicleBuilder.h
jonesmz: Initializer list works with span though?

jonesmz:
> I'm not sure if there's a noticeable change in my views, but I'd like to enforce "no <ranges>" for the time being. 

Sure, we can do that.

I think it's pretty clear that I don't share the same opinion, but i don't mind the rule overly much.

PR #269, opened 2024-04-03
Hush various clang-tidy warnings

jonesmz:


src/osp/tasks/execute.cpp
```diff
@@ -525,6 +526,7 @@ static void pipeline_advance_reqs(Tasks const& tasks, TaskGraph const& graph, Ex
                 return true; // On the right stage and task not running. This means it's done
             }
         }();
+        // NOLINTEND(ugprone-branch-clone)
```
Capital-Asterisk, line 529: `ugprone-branch-clone` -> `bugprone-branch-clone`

This check seems kind of annoying? maybe disable it?

src/osp/core/global_id.h
```diff
@@ -24,26 +24,31 @@
...
+    static_assert(std::is_integral_v<underlying_t>);
+
+    [[nodiscard]] static constexpr ID_T create() noexcept
     {
-        return ID_T(sm_count++);
+        return ID_T{sm_count++};
     }
 
-    [[nodiscard]] static inline std::size_t size() noexcept
+    [[nodiscard]] static constexpr underlying_t size() noexcept
```
Capital-Asterisk, line 45: I want this to return std::size_t since this is always used to resize containers

src/osp/activescene/basic_fn.cpp
```diff
@@ -198,8 +198,8 @@ void SysSceneGraph::do_delete(ACtxSceneGraph& rScnGraph)
         });
 
         // Shift over tree data
-        std::shift_left(itTreeDescFirst + done, itTreeDescFirst + keepLast, shift);
-        std::shift_left(itTreeEntsFirst + done, itTreeEntsFirst + keepLast, shift);
+        std::shift_left(itTreeDescFirst + done, itTreeDescFirst + keepLast, decltype(rScnGraph.m_treeDescendants)::difference_type(shift));
+        std::shift_left(itTreeEntsFirst + done, itTreeEntsFirst + keepLast, decltype(rScnGraph.m_treeToEnt)::difference_type(shift));
```
Capital-Asterisk, line 202: just change `shift` to std::ptrdiff_t

src/osp/core/global_id.h
jonesmz: doing that causes warnings in almost every place that the size() function is used.

implicit promotion to size_t almost never warns, but implicit narrowing always does.

src/osp/tasks/execute.cpp
jonesmz: imho it's a useful check, just not in this specific situation.

src/osp/core/global_id.h
jonesmz: well i guess i'm not seeing what the problem is if the thing being returned implicitly converts to std::size_t?

Here it's used to allow looping over all the instances of the id.


https://github.com/TheOpenSpaceProgram/osp-magnum/blob/a747cab17ae3f5bea02a800c8b9cfbe8a1e345c3/src/testapp/sessions/vehicles.cpp#L286-L289

same here

https://github.com/TheOpenSpaceProgram/osp-magnum/blob/a747cab17ae3f5bea02a800c8b9cfbe8a1e345c3/src/testapp/sessions/vehicles.cpp#L445-L455

Perhaps we need to add a new `largest()` function or something that returns the largest value as the type?

src/testapp/sessions/vehicles_machines.cpp
```diff
@@ -234,24 +234,26 @@ Session setup_thrust_indicators(
...
-            float const     throttle        = std::clamp(rSigValFloat[throttleIn], 0.0f, 1.0f);
-            float const     multiplier      = rSigValFloat[multiplierIn];
-            float const     thrustMag       = throttle * multiplier;
-
-            rCtxScnRdr.m_drawTransform[drawEnt]
-                    = drawTf
-                    * Matrix4::scaling({1.0f, 1.0f, thrustMag * rThrustIndicator.indicatorScale})
-                    * Matrix4::translation({0.0f, 0.0f, -1.0f})
-                    * Matrix4::scaling({0.2f, 0.2f, 1.0f});
+          if (pair.type == gc_mtMagicRocket)
```
Capital-Asterisk, line 238: accidental 2 space indent?

I've been doing this weird `for (...) if` here and there, as I think the slightly reduced nesting is worth it for weirdness as a tradeoff.

test/tasks/main.cpp
```diff
@@ -63,7 +63,7 @@ void randomized_singlethreaded_execute(Tasks const& tasks, TaskGraph const& grap
 
         if (runTasksLeft != 0)
         {
-            TaskId const        randomTask  = rExec.tasksQueuedRun.at(rRand() % runTasksLeft);
+            TaskId const        randomTask  = rExec.tasksQueuedRun.begin()[std::ptrdiff_t(rRand()) % runTasksLeft];
```
Capital-Asterisk, line 66: I'm curious of the benefits of `begin()[i]` vs just operator[]?

Capital-Asterisk (approved):

test/tasks/main.cpp
jonesmz: the deprecation notice says to use begin()[], but it seems like it works so i'll switch to just operator[].

src/testapp/sessions/vehicles_machines.cpp
jonesmz: Hrmm, alright, i can do that.

src/osp/core/global_id.h
Capital-Asterisk: I see. These don't yet use the strong ID types. Ideally I'd prefer something like
```cpp
for (TypeId const id : TypeReg_t::ids()) {...}
```

though I'd agree to just leave this function as underlying_t for now.

jonesmz: I did add a `largest()` function, but you can rename it if you want in a followup PR, no need to get a review.

PR #276, opened 2024-05-08
Updating README.md and .gitignore

Horashi0:
Added user control information to README.md

Added .vscode to .gitignore so it will stop staging it for commit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ CMakeSettings.json
 out/
 .vs/
 build/
-.vscode
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ CMakeSettings.json
 out/
 .vs/
 build/
+.vscode
--- a/README.md
+++ b/README.md
@@ -91,6 +91,54 @@ cd build-osp-magnum/Release
 
 If you just want to test out the project so far, then see the [Actions](https://github.com/TheOpenSpaceProgram/osp-magnum/actions) tab on GitHub to obtain automated builds for Linux or Windows.
 
+## Controls
+The most important thing to remember, the controls!
+
+#### Navigating
+```bash
+V - Switch game mode
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -91,6 +91,54 @@ cd build-osp-magnum/Release
 
 If you just want to test out the project so far, then see the [Actions](https://github.com/TheOpenSpaceProgram/osp-magnum/actions) tab on GitHub to obtain automated builds for Linux or Windows.
 
+## Controls
+The most important thing to remember, the controls!
```
SIGSTACKFAULT, line 95: maybe "they will be rather familiar if you have played Kerbal Space Program".

SIGSTACKFAULT, line 95: maybe "they will be rather familiar if you have played Kerbal Space Program".

SIGSTACKFAULT (changes requested):

.gitignore
```diff
@@ -13,3 +13,4 @@ CMakeSettings.json
 out/
 .vs/
 build/
+.vscode
```
jonesmz, line 16: Don't you need the trailing slash like the other entries?

README.md
jonesmz: That seems reasonable.

jonesmz (changes requested):
.

.gitignore
Horashi0: You dont have to but the / just shows that its a folder which helps for clarity so ill add it

README.md
Horashi0: I'm not sure, I don't want to get into the habit of talking about KSP in the documentation as it might come off to users that we are trying to be a KSP clone

I don't mind the change so I'll add it, just something to think about

Horashi0 (commented):
Updated changes

jonesmz (approved):

PR #277, opened 2024-05-09
Scenario System Refactor

Ben-Bingham:
I refactored the current method of loading scenarios, the main changes are:

1) Switched from storing scenarios in an unordered map where the name of the scenario was the key and the description and setup function were the value, to a singleton class that has member functions that search for you.

2) Separated each Scenario into its own file, this will be especially helpful in the future when multiple scenarios are being worked on on multiple different branches.

3) In many parts of the code scenarios are referred to as "scenes" and I changed a lot of those to all be scenario, there are still some left, but they affect parts of the code I have not looked at yet.

Overall I didn't change any functionality, at least nothing was changed intentionally, mostly just changes to improve readability


Capital-Asterisk (changes requested):
I have quite a few major issues with this.

### 1. Making new scenarios

A long time ago, scenarios use to be in separate files. Creating new scenarios was extremely difficult because one would have to copy an entire file and rejiggle a bunch of functions to do so. I thought it was easy. I then found a way to throw scenarios together like Legos (#214). Now if someone wanted to make a new scenario, they can copy paste a block of code and "bam!" there it was. I did a 10 second speedrun of this just to show off my cool system. Scenarios didn't need their own files since they were constructed with individual opt-in features. 

### 2. ScenarioManager???

why?

### 3. Whitespace issues

For all new new files in src/scenarios, there's mixed spaces and tabs everywhere and no newlines at the end of the file.

--

I don't want to be too harsh here, but you found a small problem but created a bigger problem. Regular software engineering moment though. A rough rule of thumb is that a pull request on refactoring should remove more or around the same amount of new code added.

Ben-Bingham:
Thank you for the feedback, I think your exactly correct in that I went way to overboard on one idea and way overcomplicated it, I'm going to try again and this time just straighten out the files as opposed to reinventing everything.

Also I don't think you were too harsh at all, I'm relatively new to programming/software dev and am still very much learning, I really appreciate that you were honest with me, as it will definitely help me become better.

PR #278, opened 2024-05-09
Planet terrain

Capital-Asterisk:
This PR works on the incomplete planet terrain generator 'planeta' in `src/planet-a`.

Planet terrain is created by first managing a "Skeleton." The Skeleton is a triangle mesh that uses int64 coordinates, able to represent large planets. Each triangle is subdividable into 4 more triangles (triforce style), forming a quad tree of triangles (vertices are shared and handled accordingly).

High detail triangle grid meshes (called 'chunks') can then be created, copying coordinated from the skeleton triangles. This can be used for rendering or physics colliders. This is currently not yet implemented.

Certain invariants constrain the skeleton in such a way that generated meshes are completely closed and seamless (a manifold mesh). This allows it to work well with vertex shaders with displacement, such as ocean waves.

I'm thinking of making an infographic as part of the docs, to help describe some of the inner workings.

### Terrain test scenario

New test scenario! This uses the existing camera controls to move around a 3D scene, containing a single 'planet' which subdivides and increases in detail when close to the camera's target (the green wireframe cube). A chunk mesh of the entire planet is managed in the background (there's no code that sends it to the GPU yet). Every 10 seconds, The mesh is exported is exported as an .obj file to the current directory, where it can be viewed with 3d modelling software like Blender.

### Important non-planet related changes:

* Add ArrayView2DWrapper - wraps an ArrayView for a cleaner interface
* Add has_value member function for StrongID
* Add id_utils.h - MaybeNewId and RefCountStatus
* Add math_int64.h - some generic int64-related math functions
* Add minimum distance variable to CameraController
* Fix horrible bug related to drawing materials - there was an std::vector that just kept growing non-stop

:bird:  --- :globe_with_meridians: 

fun fact: The 'a' in planet-a refers to it being the first planet implementation, as if there was going to be a 'b'. The letter was never incremented even after a major rewrite. It might as well just be "planeta" as the Spanish or Filipino word for planet.


src/osp/core/array_view.h
```diff
@@ -28,10 +28,56 @@
...
+    return view.sliceSize(index, size);
+}
+
+/**
+ * @brief Anger the address sanitizer for invalid array views / spans / similar containers
+ */
+template<typename T>
+void debug_touch_container(T const& container)
+{
+    // This feels kinda stupid, but if you're reading this and have a better idea then go ahead
```
jonesmz, line 75: ydu're -> you're

src/osp/core/array_view.h
```diff
@@ -28,10 +28,56 @@
...
+ * @param size
+ */
+template<typename T>
+ArrayView<T> slice_2d_row(ArrayView<T> const& view, std::size_t index, std::size_t size) noexcept
+{
+    return view.sliceSize(index, size);
+}
+
+/**
+ * @brief Anger the address sanitizer for invalid array views / spans / similar containers
```
jonesmz, line 70: Why do you want to make the address sanitizer mad on purpose?

What were you using this for?

src/osp/core/math_int64.h
```diff
@@ -0,0 +1,78 @@
...
+
+    return std::uint64_t((lhs > rhs) ? (lhs - rhs) : (rhs - lhs));
+};
+
+/**
+ * @brief (distance between a and b) > threshold
+ *
+ * This function is quick and dirty, max threshold is limited to 1,431,655,765
+ */
+constexpr bool is_distance_near(Vector3l const a, Vector3l const b, std::uint64_t const threshold)
```
jonesmz, line 59: noexcept

src/osp/core/array_view.h
Capital-Asterisk: I don't see any mentions of "ydu're"? is there something changing text on your end?

src/osp/core/array_view.h
Capital-Asterisk: There's cases where I had an array view that exists, and I wanted to assert that it points to valid memory. There's no uses of it right now, but at the time it was actually quite helpful.

src/osp/core/array_view.h
jonesmz: hrmm, my computer was running updates, and i guess replaced a DLL firefox wanted, and it spazzed out.

Ignore, problem on my end.

src/osp/core/array_view.h
jonesmz: Ah interesting.

In windows there's actually a OS api to let you explicitly check that at runtime.

No idea how to go about doing it in linux though.

scripts/icosahedron_tables.py
```diff
@@ -0,0 +1,173 @@
...
+# Tower height
+# If identical towers were built on the two vertices spanning an edge, this is how high each tower
+# needs to be in order to see each other over the horizon.
+#  = exsec(0.5*acot(0.5) * levels)
+tower_heights = [sec(0.5*acot(0.5) * 0.5**x ) - 1 for x in range(0, 10)]
+
+print("inline constexpr std::array<float, 10> const gc_icoTowerOverHorizonVsSubdiv\n"
+      + "{\n"
+      + "    " + ", ".join(nstr_float(edge_length) for edge_length in tower_heights) + "\n"
+      + "};\n")
```
Capital-Asterisk, line 173: Noting the existence of this python script. The icosahedron stuff here use to be calculated at runtime, but only with doubles and implementation-defined std::sqrt. Precalculating all this with much much higher precision is a far more practical

Capital-Asterisk:
Added docs with the pictures: https://github.com/TheOpenSpaceProgram/osp-magnum/blob/1e6efc187ec045493a8be14c027176110337ccdf/docs/planeta.md

PR #279, opened 2024-05-10
Keplerian orbits

dylan-mcleod:
Added a Keplerian orbit type in scientific/kepler.h, to later build patched conics on top of

### DIFF
--- a/src/osp/scientific/kepler.cpp
+++ b/src/osp/scientific/kepler.cpp
@@ -27,14 +27,11 @@
 #include <algorithm>
 #include <cassert>
 
-
 using Magnum::Math::cross;
 using Magnum::Math::dot;
 
 namespace osp
 {
-
-    constexpr double KINDA_SMALL_NUMBER = 1.0E-8;
 
     /**
      * @brief Wrap an angle to the range [0, 2π), useful for keplerian orbits
@@ -104,7 +101,7 @@ namespace osp
     {
         // https://orbital-mechanics.space/time-since-periapsis-and-keplers-equation/universal-variables.html
         // This seems to converge disastrously slowly
-        // The author recommends using Laguerre's method for faster convergence 
+        // The author recommends using Laguerre's method for faster convergence
         double alpha = 1.0 / semiMajorAxis;
         double chi = std::sqrt(mu) * std::abs(alpha) * delta_t;
         double delta;
@@ -132,7 +129,6 @@ namespace osp
 
         // use newton's method
         // todo: implement something which converges more quickly
-        // we really shouldn't need to do 20000 steps
         do
         {
             double f_E = E - eccentricity * std::sin(E) - meanAnomaly;
@@ -140,9 +136,8 @@ namespace osp
 
             delta = f_E / df_E;
             E = E - delta;
-            
-            // Not strictly necessary but sometimes this fails to converge without it
-            E = wrap_angle(E);
+
+            E = wrap_angle(E); // Sometimes this sometimes fails to converge without this line
             --steps;
         } while (std::abs(delta) > tolerance && steps > 1);
         return E;
@@ -168,6 +163,9 @@ namespace osp
         return E;
     }
 
+
+
+
     double eccentric_anomaly_to_true_anomaly(const double &ecc, const double &E)
     {
         // This formula supposedly has better numerical stability than the one below
@@ -182,6 +180,16 @@ namespace osp
         return 2 * atan(sqrt((ecc + 1) / (ecc - 1)) * tanh(H / 2));
     }
 
+    double true_anomaly_to_eccentric_anomaly(const double &ecc, const double &v)
+    {
+        return 2.0 * std::atan(std::sqrt((1 - ecc) / (1 + ecc)) * std::tan(v / 2.0));
+    }
+
+    double true_anomaly_to_hyperbolic_eccentric_anomaly(const double &ecc, const double &v)
+    {
+        return 2.0 * std::atanh(std::clamp(std::sqrt((ecc - 1) / (ecc + 1)) * std::tan(v / 2.0), -1.0, 1.0));
+    }
+
     double mean_motion(double mu, double a)
     {
         return sqrt(mu / (a * a * a));
@@ -192,81 +200,113 @@ namespace osp
         return semiMajorAxis * (1 - ecc * ecc) / (1 + ecc * std::cos(trueAnomaly));
     }
 
-    Vector3d KeplerOrbit::get_position_and_velocity(double time, Vector3d &velocity) const
+
+
+
+    double KeplerOrbit::get_true_anomaly_from_eccentric(double eccentric) const
     {
-        double trueAnomaly = 0.0;
-        double E = 0.0;
-        if (m_eccentricity <= 1.0)
+        if (is_elliptic())
         {
-#if 1 // Solve for eccentric anomaly
+            return eccentric_anomaly_to_true_anomaly(m_eccentricity, eccentric);
+        }
+        else if (is_hyperbolic())
+        {
+            return hyperbolic_eccentric_anomaly_to_true_anomaly(m_eccentricity, eccentric);
+        }
+    }
+
+    double KeplerOrbit::get_eccentric_anomaly(double time) const
+    {
+        if (is_elliptic())
+        {
+#if 1 // Use standard kepler solver
             double nu = mean_motion(m_gravitationalParameter, m_semiMajorAxis);
             double Mt = m_meanAnomalyAtEpoch + (time - m_epoch) * nu;
             Mt = wrap_angle(Mt);
-            E = solve_kepler_elliptic(Mt, m_eccentricity, 20000);
-            trueAnomaly = eccentric_anomaly_to_true_anomaly(m_eccentricity, E);
+            return solve_kepler_elliptic(Mt, m_eccentricity, 20000);
 #else // Use universal kepler solver
-            
             double chi = solve_kepler_universal(m_r0, m_v0, m_semiMajorAxis, time - m_epoch, m_gravitationalParameter, 200);
-            E = chi / std::sqrt(m_semiMajorAxis) + m_E0;
-            E = wrap_angle(E);
-            trueAnomaly = eccentric_anomaly_to_true_anomaly(m_eccentricity, E);
-            // trueAnomaly = wrap_angle(trueAnomaly);
+            double E = chi / std::sqrt(m_semiMajorAxis) + m_E0;
+            return wrap_angle(E);
 #endif
         }
         else if (is_hyperbolic())
         {
-#if 1 // Solve for hyperbolic eccentric anomaly
+#if 1 // Use standard kepler solver
             double nu = mean_motion(m_gravitationalParameter, -m_semiMajorAxis);
             double Mt = m_meanAnomalyAtEpoch + (time - m_epoch) * nu;
             // Mt = wrap_angle(Mt);
-            E = solve_kepler_hyperbolic(Mt, m_eccentricity, 20000);
-            trueAnomaly = hyperbolic_eccentric_anomaly_to_true_anomaly(m_eccentricity, E);
+            return solve_kepler_hyperbolic(Mt, m_eccentricity, 20000);
 #else // Use universal kepler solver
-            
             double chi = solve_kepler_universal(m_r0, m_v0, m_semiMajorAxis, time - m_epoch, m_gravitationalParameter, 2000);
             double F = chi / std::sqrt(-m_semiMajorAxis) + m_F0;
-
-            trueAnomaly = hyperbolic_eccentric_anomaly_to_true_anomaly(m_eccentricity, F);
+            return F;
 #endif
         }
-        double r = m_semiMajorAxis * (1 - m_eccentricity * m_eccentricity) / (1 + m_eccentricity * std::cos(trueAnomaly));
+    }
 
+    void KeplerOrbit::get_fixed_reference_frame(Vector3d &radial, Vector3d &transverse, Vector3d &outOfPlane) const
+    {
         double a = m_argumentOfPeriapsis;
         double b = m_longitudeOfAscendingNode;
         double c = m_inclination;
 
-        Vector3d radial = Vector3d(std::cos(a) * std::cos(b) - std::sin(a) * std::sin(b) * std::cos(c),
-                                   std::cos(a) * std::sin(b) + std::sin(a) * std::cos(b) * std::cos(c),
-                                   std::sin(a) * std::sin(c));
+        radial = Vector3d(std::cos(a) * std::cos(b) - std::sin(a) * std::sin(b) * std::cos(c),
+                          std::cos(a) * std::sin(b) + std::sin(a) * std::cos(b) * std::cos(c),
+                          std::sin(a) * std::sin(c));
 
-        Vector3d transverse = Vector3d(-std::sin(a) * std::cos(b) - std::cos(a) * std::sin(b) * std::cos(c),
-                                       -std::sin(a) * std::sin(b) + std::cos(a) * std::cos(b) * std::cos(c),
-                                       std::cos(a) * std::sin(c));
+        transverse = Vector3d(-std::sin(a) * std::cos(b) - std::cos(a) * std::sin(b) * std::cos(c),
+                              -std::sin(a) * std::sin(b) + std::cos(a) * std::cos(b) * std::cos(c),
+                              std::cos(a) * std::sin(c));
 
-        Vector3d outOfPlane = Vector3d(std::sin(b) * std::sin(c),
-                                       -std::cos(b) * std::sin(c),
-                                       std::cos(c));
+        outOfPlane = Vector3d(std::sin(b) * std::sin(c),
+                              -std::cos(b) * std::sin(c),
+                              std::cos(c));
+    }
 
-        Vector3d newPosition = r * (std::cos(trueAnomaly) * radial + std::sin(trueAnomaly) * transverse);
+    void KeplerOrbit::get_state_vectors_at_time(double time, Vector3d &position, Vector3d &velocity) const
+    {
+        double E = get_eccentric_anomaly(time);
+        double trueAnomaly = get_true_anomaly_from_eccentric(E);
+        get_state_vectors_at_eccentric_anomaly(E, position, velocity);
+    }
+
+    void KeplerOrbit::get_state_vectors_at_eccentric_anomaly(double E, Vector3d &position, Vector3d &velocity) const
+    {
+        double trueAnomaly = get_true_anomaly_from_eccentric(E);
+        double r = get_orbiting_radius(m_eccentricity, m_semiMajorAxis, trueAnomaly);
+        Vector3d radial, transverse, outOfPlane;
+        get_fixed_reference_frame(radial, transverse, outOfPlane);
+
+        position = r * (std::cos(trueAnomaly) * radial + std::sin(trueAnomaly) * transverse);
         if (m_eccentricity < 1.0)
         {
-            Vector3d newVelocity = (std::sqrt(m_gravitationalParameter * m_semiMajorAxis) / r) * (-std::sin(E) * radial + (std::sqrt(1 - m_eccentricity * m_eccentricity) * std::cos(E)) * transverse);
-            velocity = newVelocity;
+            velocity = (std::sqrt(m_gravitationalParameter * m_semiMajorAxis) / r) * (-std::sin(E) * radial + (std::sqrt(1 - m_eccentricity * m_eccentricity) * std::cos(E)) * transverse);
         }
         else
         {
             double v_r = (m_gravitationalParameter / m_h * m_eccentricity * std::sin(trueAnomaly));
             double v_perp = (m_h / r);
-            Vector3d radusDir = newPosition.normalized();
+            Vector3d radusDir = position.normalized();
             Vector3d perpDir = cross(outOfPlane, radusDir).normalized();
 
-            Vector3d newVelocity = v_r * radusDir + v_perp * perpDir;
-            velocity = newVelocity;
+            velocity = v_r * radusDir + v_perp * perpDir;
         }
-        return newPosition;
     }
 
-    Vector3d KeplerOrbit::get_acceleration(Vector3d radius) const
+    void KeplerOrbit::rebase_epoch(double newEpoch)
+    {
+        double new_E0 = get_eccentric_anomaly(newEpoch);
+        double new_m0 = new_E0 - m_eccentricity * std::sin(new_E0);
+        Vector3d newPos, newVel;
+        get_state_vectors_at_eccentric_anomaly(new_E0, newPos, newVel);
+        m_r0 = newPos.length();
+        m_v0 = newVel.length();
+        m_E0 = new_E0;
+        m_meanAnomalyAtEpoch = new_m0;
+    }
+
+    Vector3d KeplerOrbit::get_acceleration(const Vector3d &radius) const
     {
         double a = m_gravitationalParameter / radius.dot();
         return -a * radius.normalized();
@@ -320,8 +360,7 @@ namespace osp
         const double inclination = acos(h.z() / h.length());
 
         // Compute true anomaly v and argument of periapsis w
-        double v;
-        double w;
+        double v, w;
         if (e > KINDA_SMALL_NUMBER)
         {
             // True anomaly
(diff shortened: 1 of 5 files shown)


src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+using Magnum::Math::cross;
+using Magnum::Math::dot;
+
+namespace osp
+{
+
+    /**
+     * @brief Wrap an angle to the range [0, 2π), useful for keplerian orbits
+     */
+    inline constexpr double wrap_angle(double angle)
```
jonesmz, line 39: `constexpr` functions are implicitly `inline`, no need to add the inline keyword.

but adding `noexcept` to the end of the function signature is worth doing.

jonesmz, line 39: when practical use `const` for function parameters or local variables that aren't modified.

src/osp/scientific/kepler.h
```diff
@@ -0,0 +1,161 @@
...
+    double m_meanAnomalyAtEpoch;
+    double m_epoch;
+    double m_r0;
+    double m_v0;
+    double m_h;
+
+    union
+    {
+        double m_F0;
+        double m_E0;
```
jonesmz, line 56: can you elaborate on why you have a union of two variables of the same type? 

doesn't seem like this would do any good.

src/osp/scientific/kepler.h
```diff
@@ -0,0 +1,161 @@
...
+     * @brief Find period of orbit if elliptic
+     */
+    std::optional<double> get_period() const;
+
+    /**
+     * @brief Find Periapsis of orbit
+     */
+    double get_periapsis() const;
+
+    inline bool is_elliptic() const { return m_eccentricity < 1.0 - KINDA_SMALL_NUMBER; }
```
jonesmz, line 137: inline keyword isn't needed on functions defined inside the class body like this.

better would be `constexpr` and `noexcept`

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+    {
+        return 2.0 * std::atanh(std::clamp(std::sqrt((ecc - 1) / (ecc + 1)) * std::tan(v / 2.0), -1.0, 1.0));
+    }
+
+    double mean_motion(double mu, double a)
+    {
+        return sqrt(mu / (a * a * a));
+    }
+
+    double get_orbiting_radius(double ecc, double semiMajorAxis, double trueAnomaly)
```
jonesmz, line 198: it looks like all of the functions near the top of this file are internal only to the cpp file.

if that's the case, they should be put into an anonymous namespace, and marked with `static` to help the linker understand they can't "escape" the translation unit.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+    {
+        double trueAnomaly = get_true_anomaly_from_eccentric(E);
+        double r = get_orbiting_radius(m_eccentricity, m_semiMajorAxis, trueAnomaly);
+        Vector3d radial, transverse, outOfPlane;
+        get_fixed_reference_frame(radial, transverse, outOfPlane);
+
+        position = r * (std::cos(trueAnomaly) * radial + std::sin(trueAnomaly) * transverse);
+        if (m_eccentricity < 1.0)
+        {
+            velocity = (std::sqrt(m_gravitationalParameter * m_semiMajorAxis) / r) * (-std::sin(E) * radial + (std::sqrt(1 - m_eccentricity * m_eccentricity) * std::cos(E)) * transverse);
```
jonesmz, line 284: it'd be nice to break this very long line up a bit with a newline.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+            velocity = v_r * radusDir + v_perp * perpDir;
+        }
+    }
+
+    void KeplerOrbit::rebase_epoch(double newEpoch)
+    {
+        double new_E0 = get_eccentric_anomaly(newEpoch);
+        double new_m0 = new_E0 - m_eccentricity * std::sin(new_E0);
+        Vector3d newPos, newVel;
```
jonesmz, line 301: i strongly discourage declaring more than one variable per line, with a comma like this.
It's very prone to confusing junior programmers.

Only exception would be situations like in a for-loop where the comma is the only way to keep the scoping concise and still get multiple variables.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+        {
+            // circular orbit, todo
+            E = v;
+            M0 = E - e * std::sin(E);
+        }
+
+        return KeplerOrbit(semiMajorAxis, e, inclination, w, LAN, M0, epoch, GM, radius.length(), velocity.length(), E, h.length());
+    }
+
+} // namespace osp
```
jonesmz, line 436: prefer having a single blank line at the end of files.

Tools like github code review get grumpy without it.

src/testapp/sessions/orbits.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+        auto const tgUCore = uniCore.get_pipelines<PlUniCore>();
+
+        auto &rUniverse = top_get<Universe>(topData, idUniverse);
+
+        constexpr int precision = 10;
+        constexpr int planetCount = 64;
+        constexpr int seed = 1337;
+        constexpr spaceint_t maxDist = math::mul_2pow<spaceint_t, int>(20000ul, precision);
+        constexpr float maxVel = 800.0f;
```
jonesmz, line 70: minor preference for `static constexpr`

src/testapp/sessions/orbits.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+        rBuilder.task()
+            .name("Advanced planet orbits")
+            .run_on(tgUCore.update(Run))
+            .sync_with({tgUSFrm.sceneFrame(Modify)})
+            .push_to(out.m_tasks)
+            .args({idUniverse, idPlanetMainSpace, idSatSurfaceSpaces, tgUniDeltaTimeIn})
+            .func([](Universe &rUniverse, CoSpaceId const rPlanetMainSpace, CoSpaceIdVec_t const &rSatSurfaceSpaces, float const uniDeltaTimeIn) noexcept
+                  {
+        auto &rMainSpaceCommon = rUniverse.m_coordCommon[rPlanetMainSpace];
```
jonesmz, line 188: the indentation here seems out of order?

jonesmz (approved):

jonesmz:
please also make sure the continuous integration passes before you finish your changes for this PR.

src/testapp/sessions/orbits.cpp
```diff
@@ -0,0 +1,234 @@
...
+namespace testapp::scenes
+{
+
+    osp::Session setup_orbit_planets(
+        TopTaskBuilder &rBuilder,
+        ArrayView<entt::any> topData,
+        Session &uniCore,
+        Session &uniScnFrame)
+    {
+        feenableexcept(FE_INVALID | FE_OVERFLOW);
```
jonesmz, line 55: this function appears to only exist on linux?

I'm not sure enabling floating point exception trapping in a platform specific manner is the right approach.

if you're doing math operations that are prone to encountering floating point exceptions, i'd rather see standardized / standard-library functionality used only, unless there's a substantial performance improvement to be had from calling out to platform specific functions.

seems like this header has the standardized stuff in it? https://en.cppreference.com/w/cpp/numeric/fenv

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include "kepler.h"
+#include <algorithm>
+#include <cassert>
+
+using Magnum::Math::cross;
+using Magnum::Math::dot;
+
+namespace osp
+{
+
```
Capital-Asterisk, line 35: namespaces don't use an extra indentation level

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+            delta = f_chi / df_chi;
+            chi = chi - delta;
+            --steps;
+        } while (std::abs(delta) > 1.0E-10 && steps > 1);
+        return chi;
+    }
+
+    double solve_kepler_elliptic(double meanAnomaly, double eccentricity, int steps)
+    {
+        const double tolerance = 1.0E-10;
```
Capital-Asterisk, line 121: Project's convention is to put const on the right side after the type. I usually put const wherever possible (literally everywhere, it's a little noisy though, if only const is the default)

src/osp/scientific/kepler.h
dylan-mcleod: It was due to a notational difference between hyperbolic and elliptic orbits, but the variable name would be expressed better as m_eccentricAnomalyAtEpoch regardless.

src/osp/scientific/kepler.cpp
dylan-mcleod: Should it be an anonymous namespace inside of namespace osp, or outside of namespace osp?

src/osp/scientific/kepler.h
```diff
@@ -0,0 +1,161 @@
...
+                double inclination,
+                double argumentOfPeriapsis,
+                double longitudeOfAscendingNode,
+                double meanAnomalyAtEpoch,
+                double epoch,
+                double gravitationalParameter,
+                double r0,
+                double v0,
+                double m_E0,
+                double h) : m_semiMajorAxis(semiMajorAxis),
```
Capital-Asterisk, line 60: The cool kids way of doing this would be
```cpp
KeplerOrbit(...)
 : m_semiMajorAxis(semiMajorAxis)
 , m_eccentricity(eccentricity)
 , m_inclination(inclination)
 ...
```

An alternative is that these parameters can be moved into a separate plain-old-data struct and stored as a private member. In this case, you only need to pass one thing through the constructor. you can drop the `m_` prefix and also initialize it with designated initializers, like `KeplerOrbit orbit{{semiMajorAxis = a, .eccentricity = b, ...}}`

src/osp/scientific/kepler.h
Capital-Asterisk: some extra context: function bodies inside classes/structs are automatically considered inline

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+        {
+            return hyperbolic_eccentric_anomaly_to_true_anomaly(m_eccentricity, eccentric);
+        }
+    }
+
+    double KeplerOrbit::get_eccentric_anomaly(double time) const
+    {
+        if (is_elliptic())
+        {
+#if 1 // Use standard kepler solver
```
Capital-Asterisk: whatever is going to be here in the final version, `if constexpr` can likely do the job here

src/osp/scientific/kepler.h
dylan-mcleod: Could you send me an example of that second approach? That's some C++ magic that I haven't used before

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,436 @@
...
+        }
+    }
+
+    void KeplerOrbit::get_fixed_reference_frame(Vector3d &radial, Vector3d &transverse, Vector3d &outOfPlane) const
+    {
+        double a = m_argumentOfPeriapsis;
+        double b = m_longitudeOfAscendingNode;
+        double c = m_inclination;
+
+        radial = Vector3d(std::cos(a) * std::cos(b) - std::sin(a) * std::sin(b) * std::cos(c),
```
Capital-Asterisk: I feel like it's practical to just have `using std::sin;` and `using std::cos;` at the top of the file, this feels a little noisy.

src/osp/scientific/kepler.cpp
jonesmz: It doesn't really matter.

Personally I like to put my anonymous namespaces at global scope so that any types from other namespaces are explicitly written out and there's no ambiguity, but other people do it differently.

I don't think we have any guidelines on which to do, so dealer's choice i guess.

src/osp/scientific/kepler.h
Capital-Asterisk:
```cpp
// store this as a private member
struct KeplerOrbitParams
{
    double semiMajorAxis;
    double eccentricity;
    double inclination;
    double argumentOfPeriapsis;
    // pretend the rest are here
};

// constructor would look something like
KeplerOrbit::KeplerOrbit(KeplerOrbitParams params) : m_params{params} { };


// in some function, maybe in from_initial_conditions(...)?
// using Designated Initializer

KeplerOrbitParams params{
    .semiMajorAxis = semiMajorAxis,
    .eccentricity = e,
    .inclination = inclination,
    .argumentOfPeriapsis = w,
};

return KeplerOrbit{params};
```

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,368 @@
...
+        double df_E = eccentricity * std::cosh(E) - 1.0;
+
+        delta = f_E / df_E;
+        E = E - delta;
+        --steps;
+    } while (std::abs(delta) > tolerance && steps > 1);
+    return E;
+}
+
+
```
jonesmz, line 102: can you remove the extra newlines?

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,368 @@
...
+        delta = f_E / df_E;
+        E = E - delta;
+
+        E = wrap_angle(E); // Sometimes this sometimes fails to converge without this line
+        --steps;
+    } while (std::abs(delta) > tolerance && steps > 1);
+    return E;
+}
+
+static double solve_kepler_hyperbolic(double const meanAnomaly, double const eccentricity, int steps)
```
jonesmz, line 80: can you add some doxygen documentation?

don't go overboard. Just a one sentence explanation of the point of the function is enough.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,437 @@
...
+        double df_E = eccentricity * cosh(E) - 1.0;
+
+        delta = f_E / df_E;
+        E = E - delta;
+        --steps;
+    } while (std::abs(delta) > tolerance && steps > 1);
+    return E;
+}
+
+static double elliptic_eccentric_anomaly_to_true_anomaly(double const &ecc, double const &E)
```
jonesmz, line 115: can you add some doxygen documentation?

don't go overboard. Just a one sentence explanation of the point of the function is enough.


extrapolate to all the functions :-)

src/osp/scientific/kepler.cpp
```diff
@@ -133,63 +143,63 @@ static double mean_motion(double const mu, double const a)
 
 static double get_orbiting_radius(double const ecc, double const semiMajorAxis, double const trueAnomaly)
```
jonesmz, line 144: all of these math functions look like they are `noexcept` right?

tigercoding56: i am a noob at c++ but in (76-77)  of kerpler.cpp  double df_E = 1.0 - eccentricity * cos(E); // if eccentricity happens to be 1,  a div by 0   , since c++ do-while loops   execute at least once,  idk if that causes issues in c++ but every language i actually have skills in throws a exception

jonesmz: my understanding is that floating point exceptions are not the same as C++ exceptions. I think if it does a divide by zero, the program will simply crash

regardless, if crashing isn't the right answer, then what *should* the function do?

jonesmz: More accurately,  on posix platforms it'll receive a signal that interrupts program flow and calls a signal handler (which we have no handler registered, so it'll default to crashing)

on windows it'll likely result in a structured exception handler exception being thrown.

Regardless, throwing a c++ exception isnt the right answer, we need to have a solid determination of what we want the program to do if div by zero happens in this function.

Crashing is one choice, or we can return some kind of error code, but we want to document what the behaivor is going to be and then stick to it.

dylan-mcleod: A floating point math error isn't necessarily an unrecoverable error for the end application, so I don't know if crashing is a good default choice choice for handling them.

I'm not well versed in c++ error handling though, so I don't know what the right way to handle floating point errors would be.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/**
+ * @brief Solve Kepler's equation for hyperbolic orbits (E = M + e * sinh(E))
+ * 
+ * @return Eccentric anomaly E
+ */
+static double solve_kepler_hyperbolic(double const meanAnomaly, double const eccentricity, int steps)
+{
+    double const tolerance = 1.0E-10;
```
jonesmz, line 96: you can make `tolerance` into a `constexpr` variable here can't you?

`static constexpr tolerance = 1.0E-10;`

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+    double delta;
+
+    // initial guess
+    double E = std::log(std::abs(meanAnomaly) + 0.1);
+
+    // newton-rhapson
+    do
+    {
+        double f_E = eccentricity * sinh(E) - E - meanAnomaly;
+        double df_E = eccentricity * cosh(E) - 1.0;
```
jonesmz, line 107: `f_E` and `df_E` can both be const.

I know this might seem very silly to point out, but I've watched so many people make really brain dead mistakes because they've modified a variable that they intended to never be modified.

I think it's a serious mis-design of C++ that variables aren't immutable by default, and require a keyword to make mutable. Instead of the situation we have now.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+        delta = f_E / df_E;
+        E = E - delta;
+        --steps;
+    } while (std::abs(delta) > tolerance && steps > 1);
```
jonesmz, line 112: minor: I recommend wrapping the individual portions of your conditional expression in parens.

Watched a lot of people mistake the order of evaluation here too... Had a few bugs in prod because of it at work.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+/**
+ * @brief Convert eccentric anomaly (E) to true anomaly (ν) for elliptic orbits
+ */
+static double elliptic_eccentric_anomaly_to_true_anomaly(double const &ecc, double const &E)
+{
+    // This formula supposedly has better numerical stability than the one below
+    double const beta = ecc / (1 + std::sqrt(1 - ecc * ecc));
+    return E + 2.0 * atan2(beta * sin(E), 1 - beta * cos(E));
+
+    // return 2.0 * atan2(std::sqrt(1 + ecc) * sin(E / 2.0), std::sqrt(1 - ecc) * cos(E / 2.0));
```
jonesmz, line 125: minor: Best to remove unused/commented-out code

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+ */
+static double true_anomaly_to_elliptic_eccentric_anomaly(double const &ecc, double const &v)
+{
+    return 2.0 * atan(std::sqrt((1 - ecc) / (1 + ecc)) * tan(v / 2.0));
+}
+
+/**
+ * @brief Convert true anomaly (ν) to hyperbolic eccentric anomaly (E)
+ */ 
+static double true_anomaly_to_hyperbolic_eccentric_anomaly(double const &ecc, double const &v)
```
jonesmz, line 147: Is there any value in taking `ecc` and `v` as references instead of by value?

At least on x86_64 linux, with clang, `sizeof(double)` is 8bytes, which is the same as a pointer.

So on most platforms, when passing by value, the double should be passed via register (assuming few enough parameters). On platforms where there isnt enough space to pass by register, it'll implicitly pass by reference / pointer regardless.

src/osp/scientific/kepler.cpp
jonesmz: My recommendation is to add `noexcept` unless someone provides a signal-handler -> exception translator, which very well might happen.

Until then, `noexcept` can result in better code-gen in many situations.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+void KeplerOrbit::get_uvw_vectors(Vector3d &u, Vector3d &v, Vector3d &w) const
+{
+    double const a = m_params.argumentOfPeriapsis;
+    double const b = m_params.longitudeOfAscendingNode;
+    double const c = m_params.inclination;
+
+    u = Vector3d(cos(a) * cos(b) - sin(a) * sin(b) * cos(c),
+                 cos(a) * sin(b) + sin(a) * cos(b) * cos(c),
+                 sin(a) * sin(c));
```
jonesmz, line 254: is it fair to assume that C++ standard library doesn't provide any special math functions that do much / any of this for us? i'm guessing that cos and sin are all we get?

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+    // newton-rhapson
+    do
+    {
+        double f_E = E - eccentricity * sin(E) - meanAnomaly;
+        double df_E = 1.0 - eccentricity * cos(E);
+
+        delta = f_E / df_E;
+        E = E - delta;
+
+        E = wrap_angle(E); // Sometimes this sometimes fails to converge without this line
```
jonesmz, line 83: sometimes this sometimes -> the algorithm sometimes

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    double delta;
+
+    // initial guess
+    double E = std::log(std::abs(meanAnomaly) + 0.1);
+
+    // newton-rhapson
+    do
+    {
+        double f_E = eccentricity * sinh(E) - E - meanAnomaly;
```
jonesmz, line 106: minor: not a show-stopper, but i personally prefer to see functions from the standard library explicitly called out as coming from the `std::` namespace.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+static void get_state_vectors(double   const eccentricAnomaly, 
+                              double   const trueAnomaly, 
+                              double   const eccentricity, 
+                              double   const semiMajorAxis, 
+                              double   const gravitationalParameter, 
+                              Vector3d const u, 
+                              Vector3d const v, 
+                              Vector3d const w,
+                              Vector3d      &position, 
+                              Vector3d      &velocity)
```
jonesmz, line 201: it doesn't appear that position and velocity are used as input parameters, but are just output parameters.

Consider returning `std::pair<Vector3d, Vector3d>` instead, and then unpacking the tuple at the place where the function is called.

This is less about codegen and more about just how using the function looks.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+    }
+    else
+    {
+        double const F = time_to_hyperbolic_eccentric_anomaly(m_params.meanAnomalyAtEpoch, m_params.eccentricity, 
+                    m_params.semiMajorAxis, m_params.gravitationalParameter, time - m_params.epoch);
+        return hyperbolic_eccentric_anomaly_to_true_anomaly(m_params.eccentricity, F);
+    }
+}
+
+void KeplerOrbit::get_uvw_vectors(Vector3d &u, Vector3d &v, Vector3d &w) const
```
jonesmz, line 246: same as my comment for `get_state_vectors`

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+using Magnum::Math::cross;
+using Magnum::Math::dot;
+
+namespace {
+
+using namespace osp;
+
+// M_PI isn't standard, so define it here.
+static constexpr double PI = Magnum::Math::Constants<double>::pi();
```
jonesmz, line 39: you can use pi_v from https://en.cppreference.com/w/cpp/numeric/constants, if you prefer.

src/osp/scientific/kepler.cpp
```diff
@@ -0,0 +1,480 @@
...
+        .meanAnomalyAtEpoch = M0,
+        .gravitationalParameter = GM,
+        .epoch = epoch,
+    };
+
+    return KeplerOrbit(params); 
+}
+
+std::optional<double> KeplerOrbit::get_next_time_radius_equals(const double r, const double currentTime) const {
+    std::optional<double> apoapsis = get_apoapsis();
```
jonesmz, line 425: i like std::optional a lot from a programmer ease of use perspective, but just wanted to double check that you don't consider it a barrier to good codegen compared to using explicit checks for NaN ?

src/osp/scientific/kepler.h
```diff
@@ -0,0 +1,158 @@
...
+        double argumentOfPeriapsis;
+        double longitudeOfAscendingNode;
+        double meanAnomalyAtEpoch;
+        double gravitationalParameter;
+        double epoch;
+    };
+
+    KeplerOrbitParams m_params;
+
+    KeplerOrbit(KeplerOrbitParams const params) : m_params(params) {}
```
jonesmz, line 60: `KeplerOrbitParams` is large enough that it should probably always be passed by const& to the function.

src/testapp/sessions/orbits.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    // Allocate data for all planets
+    rMainSpaceCommon.m_data = Array<unsigned char>{Corrade::NoInit, bytesUsed};
+
+    // Create easily accessible array views for each component
+    auto const [x, y, z] = sat_views(rMainSpaceCommon.m_satPositions, rMainSpaceCommon.m_data, planetCount);
+    auto const [vx, vy, vz] = sat_views(rMainSpaceCommon.m_satVelocities, rMainSpaceCommon.m_data, planetCount);
+    auto const [qx, qy, qz, qw] = sat_views(rMainSpaceCommon.m_satRotations, rMainSpaceCommon.m_data, planetCount);
+
+    std::mt19937 gen(seed);
```
jonesmz, line 117: you might get slightly better random number generation performance with the _64 version of the mt19937

PR #280, opened 2024-05-11
Documenting helper functions related to topData

Ben-Bingham:
Added some documentation to some functions that add/get/reserve value in topData, no code was modified. 

I tried to format the documentation in the same way that documentation was formatted in other parts of the codebase.

### DIFF
--- a/src/osp/tasks/top_session.h
+++ b/src/osp/tasks/top_session.h
@@ -46,8 +46,9 @@
 
 /**
  * @brief Gives the next available section of topData to a session
- * and declares an ID variable for each element that is given.
- * The same ID's are also stored inside of the sessions m_data member
+ *     and declares an ID variable for each element that is given.
+ *
+ * The same ID's are also stored inside of the sessions m_data member.
  *
  * @param count The number of id variables passed to the macro.
  * @param ids A set of variables that will each be assigned to a unique index of topData.
@@ -62,7 +63,7 @@
 
 /**
  * @brief Retrieves a section of data within topData that was already given to a session
- * and maps an ID variable to each element that is retrieved.
+ *     and maps an ID variable to each element that is retrieved.
  *
  * @param count The number of id variables passed to the macro.
  * @param ids A set of variables that will each be assigned to a unique index of topData.
--- a/src/osp/tasks/top_session.h
+++ b/src/osp/tasks/top_session.h
@@ -50,8 +50,9 @@
  *
  * The same ID's are also stored inside of the sessions m_data member.
  *
- * @param count The number of id variables passed to the macro.
- * @param ids A set of variables that will each be assigned to a unique index of topData.
+ * @param arglist - Made of two sections:
+ *     count: The number of id variables passed to the macro.
+ *     ids: A set of variables that will each be assigned to a unique index of topData.
  */
 #define OSP_DECLARE_CREATE_DATA_IDS(session, topData, arglist) OSP_AUX_DCDI_A(session, topData, arglist);
 
@@ -65,8 +66,9 @@
  * @brief Retrieves a section of data within topData that was already given to a session
  *     and maps an ID variable to each element that is retrieved.
  *
- * @param count The number of id variables passed to the macro.
- * @param ids A set of variables that will each be assigned to a unique index of topData.
+ * @param arglist - Made of two sections:
+ *     count: The number of id variables passed to the macro.
+ *     ids: A set of variables that will each be assigned to a unique index of topData.
  */
 #define OSP_DECLARE_GET_DATA_IDS(session, arglist) OSP_AUX_DGDI_A(session, arglist);
 
--- a/src/osp/tasks/top_session.h
+++ b/src/osp/tasks/top_session.h
@@ -45,7 +45,13 @@
 #define OSP_AUX_DCDI_A(...) OSP_AUX_DCDI_B(OSP_AUX_DCDI_C(__VA_ARGS__))
 
 /**
- * @brief
+ * @brief Gives the next available section of topData to a session
+ * and declares an ID variable for each element that is given.
+ * The same ID's are also stored inside of the sessions m_data member
+ *
+ * @param arglist A variadic argument broken into two sections:
+ *     Firstly: 'count' The count is an integer equal to the number of id's
+ *     Secondly: id's Every id is a unique index into 'topData'
  */
 #define OSP_DECLARE_CREATE_DATA_IDS(session, topData, arglist) OSP_AUX_DCDI_A(session, topData, arglist);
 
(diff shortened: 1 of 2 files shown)


src/osp/tasks/top_session.h
```diff
@@ -45,7 +45,13 @@
...
 
 /**
- * @brief
+ * @brief Gives the next available section of topData to a session
+ * and declares an ID variable for each element that is given.
+ * The same ID's are also stored inside of the sessions m_data member
+ *
+ * @param arglist A variadic argument broken into two sections:
+ *     Firstly: 'count' The count is an integer equal to the number of id's
+ *     Secondly: id's Every id is a unique index into 'topData'
```
jonesmz, line 54: instead of documenting implementation details like the macro taking a variadic argument list, just document what the macro takes.

sounds like the macro simply always requires a count parameter, so probably this OSP_DECLARE_CREATE_DATA_IDS() macros should be given the count parameter as an explicit parameter instead of relying on other macros that it call to break it out of the variadic argument.

src/osp/tasks/top_session.h
```diff
@@ -55,6 +61,17 @@
 #define OSP_AUX_DGDI_B(x) x
 #define OSP_AUX_DGDI_A(...) OSP_AUX_DGDI_B(OSP_AUX_DGDI_C(__VA_ARGS__))
```
jonesmz, line 62: If I'm reading how this macro expansion works, OSP_AUX_DCDI_A is passed OSP_DECLAR_CREATE_DATA_IDS() "session" parameter as the first arg, which stays as "session", and then "topData" becomes "count".

Maybe we need to normalize "topData" into always being named "count".

src/osp/tasks/top_session.h
```diff
@@ -55,6 +61,17 @@
 #define OSP_AUX_DGDI_B(x) x
```
jonesmz, line 61: Why is this "DGDI", and not "DCDI" ?

src/osp/tasks/top_session.h
```diff
@@ -45,7 +45,13 @@
 #define OSP_AUX_DCDI_A(...) OSP_AUX_DCDI_B(OSP_AUX_DCDI_C(__VA_ARGS__))
 
 /**
- * @brief
+ * @brief Gives the next available section of topData to a session
+ * and declares an ID variable for each element that is given.
+ * The same ID's are also stored inside of the sessions m_data member
```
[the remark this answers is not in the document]
jonesmz: Nevermind this, the review was hiding some important details so i misread what was going on.

src/osp/tasks/top_session.h
jonesmz: Nevermind this, the review was hiding some important details so i misread what was going on.

src/osp/tasks/top_session.h
jonesmz: Nevermind this, the review was hiding some important details so i misread what was going on.

src/osp/tasks/top_session.h
jonesmz: Nevermind this, the review was hiding some important details so i misread what was going on.

src/osp/tasks/top_session.h
```diff
@@ -45,7 +45,13 @@
...
+ *
+ * @param arglist A variadic argument broken into two sections:
+ *     Firstly: 'count' The count is an integer equal to the number of id's
```
jonesmz, line 53: instead of documenting implementation details like the macro taking a variadic argument list, just document what the macro takes.

sounds like the macro simply always requires a count parameter, so probably this OSP_DECLARE_CREATE_DATA_IDS() macros should be given the count parameter as an explicit parameter instead of relying on other macros that it call to break it out of the variadic argument.

src/osp/tasks/top_session.h
```diff
@@ -55,6 +61,17 @@
...
 #define OSP_AUX_DGDI_A(...) OSP_AUX_DGDI_B(OSP_AUX_DGDI_C(__VA_ARGS__))
 
+/**
+ * @brief Retrieves a section of data within topData that was already given to a session
+ * and maps an ID variable to each element that is retrieved.
+ *
+ * Because the data was already given, the indices are stored within the session and topData does
+ * not need to be accessed.
+ *
+ * @param arglist A variadic argument broken into two sections:
```
jonesmz, line 71: instead of documenting implementation details like the macro taking a variadic argument list, just document what the macro takes.

sounds like the macro simply always requires a count parameter, so probably this OSP_DECLARE_CREATE_DATA_IDS() macros should be given the count parameter as an explicit parameter instead of relying on other macros that it call to break it out of the variadic argument.

src/osp/tasks/top_utils.h
```diff
@@ -41,6 +41,11 @@
 namespace osp
 {
 
+/**
+ * @brief Reserves the first empty value found in topData after the index current
```
jonesmz, line 45: Reserves the first available space in topData after indicated index.

src/osp/tasks/top_utils.h
```diff
@@ -41,6 +41,11 @@
...
+ *
+ * @return The index of the reserved value, or if topData is full, the largest index in topData
```
jonesmz, line 47: the index of the reserved value, or the index after the end if no slots are available after the indicated index.

src/osp/tasks/top_utils.h
```diff
@@ -57,6 +62,13 @@ namespace osp
     return current;
 }
 
+/**
+ * @brief Reserves n values in topData, where n = destLast - destFirst
```
jonesmz, line 66: You're getting to deep into the implementation details.

Something like:

Reserves slots for each item in the output range [destFirst, destLast)

src/osp/tasks/top_utils.h
```diff
@@ -81,6 +93,11 @@ TopDataId top_reserve(
     return current;
 }
 
+/**
+ * @brief Using args, constructs an object of type T at the index ID inside topData
```
jonesmz, line 97: Constructs an object of type T at the indicated index.

jonesmz (approved):

src/osp/tasks/top_session.h
```diff
@@ -45,7 +45,12 @@
 #define OSP_AUX_DCDI_A(...) OSP_AUX_DCDI_B(OSP_AUX_DCDI_C(__VA_ARGS__))
 
 /**
- * @brief
+ * @brief Gives the next available section of topData to a session
+ * and declares an ID variable for each element that is given.
+ * The same ID's are also stored inside of the sessions m_data member
+ *
+ * @param count The number of id variables passed to the macro.
+ * @param ids A set of variables that will each be assigned to a unique index of topData.
```
Capital-Asterisk, line 53: some spacing and formatting changes:
```
@brief Gives the next available section of topData to a session
       and declares an ID variable for each element that is given.
       
The same ID's are also stored inside of the sessions m_data member
```
These comments can be used to generate html documentation through doxygen, but I've never actually run it before. I only really care about being able to reading docs in the header code itself, since this is where I'm going to be looking most of the time.

Also, the params here are suppose to be (session, topData, arglist) but it says count and ids? Specifically just the arglist is a count and variable names.

--- extra unneeded information:

There are actually some janky ways of counting the number of arguments instead of passing a number of arguments, but I've yet to implement any. 

I mean jank because there's no way to measure the number of args in __VA_ARGS__. one way I'm thinking of doing this is turning __VA_ARGS__ into a struct of chars or std::byte, then measuring its size:
```c
struct Dummy{
    std::byte __VA_ARGS__; // becomes "std::byte argA, argB, argC, argD;"
};
int argsCount = sizeof(Dummy);
```

Ben-Bingham:
I switched over the @param portions of the macro docs to say "arglist" instead of both count and ID.

As jonesmz said in a previous comment count is needed for every call of the macro, so I will make another pull request that either changes the arguments in the macros to always include the count parameter, or as Capital-Asterisk suggested there are methods to count the number of arguments in a variadic argument, I will try to implement that method first.

If I can get it working with either method I will also update the documentation to still be accurate.

src/osp/tasks/top_session.h
jonesmz: You could use https://www.boost.org/doc/libs/1_69_0/libs/preprocessor/doc/ref/variadic_size.html

But we don't depend on boost and adding that dependency just for a macro seems like a bad idea.

Capital-Asterisk (approved):

PR #284, opened 2024-05-27
Quick fix for the gtest lib issue

Lamakaio:
Quick fix I used for the gtest lib issue. I just forced gtest libs to be in the same directory as the other libs (they weren't), and added an `LD_LIBRARY_PATH` to the tests to search in that library. 

I still don't know why cmake doesn't add a rpath for the tests.

I am not sure if you want to merge this, as it is a bit of a dirty fix. It's easy to just keep it locally anyway.


Capital-Asterisk:
Welcome!

There's a way to add RPATH, like here: https://github.com/TheOpenSpaceProgram/osp-magnum/blob/a5c4ea4d4953bff6e3defd258933346742242b0f/src/CMakeLists.txt#L140-L147

I remember there was some issues with RPATH vs RUNPATH in the past. The binaries built on the ubuntu CI workflow were different from what was built on my system (a different distro), leading to a minor incompatibility.

Lamakaio:
Thank you ! 
I did variations of this, but I'll try that exact syntax tomorrow.

jonesmz:
CMake's rpath handling for unit tests is generally pretty terrible, and definitely does not work on windows.

The solution that I've seen work well in the past is to just ensure that any dll goes into the artifact directory just like how this PR does it.

jonesmz (approved):

Issue #282, opened 2024-05-26
Replace osp::BitVector_t with lgrn::IdSetStl


Capital-Asterisk:
Something to do. This doesn't require too much knowledge of how everything works. Depends on #281, this issue is mostly copied from it.

### Task

Find occurrences of `osp::BitVector_t`.
Figure out which ID type are used for it. (eg: find an occurrence of `container.set(std::size_t(id))`, check what type id is)
Replace the things that would eventually error...
* `osp::BitVector_t` → `lgrn::IdSetStl<IdType>`
* `bitvector_resize(container, size)` → `container.resize(size)`
* `container.set(std::size_t(foo))` → `container.insert(foo)`
* `container.reset(std::size_t(foo))` → `container.erase(foo)`
* `container.test(std::size_t(foo))` → `container.contains(foo)`
* `container.reset()` → `container.clear()`
* `for (std::size_t const idInt : container.ones())` → `for (IdType const id : container)`

Eventually delete `src/osp/core/bitvector.h`

### Context

[Longeron++](https://github.com/Capital-Asterisk/longeronpp) is a library consisting of some OSP 'core' components at the time, that I split off to allow it to be used on different projects and be presented on its own. I made some updates to fix some ugly code.

### osp::BitVector_t replacement

BitVectors are often used in OSP to represent sets of unique integer IDs (eg: `[1, 2, 24, 83]`). This is done for performance and simplicity: easily outperforming std::set or sorted std::vector of integers in many tasks. Iterating 'ones' bits can be done very quickly as we have 64 bit ints (not to mention SIMD). This is able to work in this project specifically as IDs are array indices and are close to zero; the highest ID correlates to the max number of objects.

`osp::BitVector_t` (aka: `lgrn::BitView< std::vector<uint64_t> >`) has an interface similar to std::bitset. The way it's used is kinda ugly:
```cpp
// get these from somewhere. ActiveEnt is osp::StrongId<std::uint32_t, ...>
ActiveEnt foo = ActiveEnt{12}; 
ActiveEnt bar = ActiveEnt{63};

osp::BitVector_t ids;
osp::bitvector_resize(ids, 100); // yucky

// IDs converted to an integer to add to the set. two different methods:
ids.set(std::size_t(foo));
ids.set(bar.value);

// ids now contains [12, 63]

ids.test(foo.value); // = true

for (std::size_t const idInt : ids.ones())
{
[...]

PR #285, opened 2024-05-28
Replace osp::BitVector_t with lgrn::IdSetStl

Lamakaio:
This pull request replaces osp::BitVector_t with lgrn::IdSetStl as outlined in issue #282. It depends on #281 and #284 at the moment. 

All the functional refactoring should be done, but for some reason both this branch and #281 fail to compile on my (admittedly janky) hardware, and I still need to do a pass to check formatting and variable names.

### DIFF
--- a/src/testapp/enginetest.cpp
+++ b/src/testapp/enginetest.cpp
@@ -133,8 +133,8 @@ entt::any setup_scene(osp::Resources& rResources, osp::PkgId const pkg)
 
     // Resize some containers to fit all existing entities
     std::size_t const maxEnts = rScene.m_activeIds.vec().capacity();
-    rScene.m_matPhong.ints()    .resize(maxEnts);
-    rScene.m_basic.m_scnGraph   .resize(maxEnts);
+    rScene.m_matPhong.resize(maxEnts);
+    rScene.m_basic.m_scnGraph.resize(maxEnts);
     rScene.m_scnRdr.resize_active(maxEnts);
     rScene.m_scnRdr.resize_draw();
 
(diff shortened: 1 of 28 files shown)


Capital-Asterisk:
Delightfully surprised to see this addressed rather quickly, thanks!

May I ask what the build issue might be? If it's related to building SDL2, then there exists the `-DOSP_USE_SYSTEM_SDL=ON` option that can be enabled to use SDL2 through a system package manager or other means.

Lamakaio:
I solved the build issue. Somehow my computer tends to crash during builds (not only this project) and it corrupted some compilation artifacts or something, entirely on my end anyway. 

However, I am now having some issues on assertion fail due to insufficient size for the IdSets (trying to insert bit 64 on set of size 64). 

In some places in the code, sets are resized like that  (in testapp/sessions/shapes.cpp)
```cpp
rPhys.m_hasColliders.resize(rBasic.m_activeIds.vec().capacity());
```

The resize function looks like that (in id_set_htl.hpp)
```cpp
void resize(std::size_t n)
    {
        vec().resize(lgrn::div_ceil(n, Base_t::bitview().int_bitsize()), 0);
    }
```

To me, it looks like we're using the capacity of an uint64 vector as the number of individual bits in the IdSet, and that is why it is too small.

However, I cannot figure out why it would work before if that was the case, as my changes should not have changed that behavior.

Anyway, I get assertion fails on insert on the sets that are resized this way (after the refactor).

jonesmz:
> Somehow my computer tends to crash during builds (not only this project) and it corrupted some compilation artifacts or something, entirely on my end anyway.

Try limiting parallelism in the build.

`cmake --build path/to/build --parallel numberParallelJobs``

Lamakaio:
Thank you, that makes a lot of sense !

I fixed it and rebased the PR.

src/testapp/enginetest.cpp
```diff
@@ -133,8 +133,8 @@ entt::any setup_scene(osp::Resources& rResources, osp::PkgId const pkg)
 
     // Resize some containers to fit all existing entities
     std::size_t const maxEnts = rScene.m_activeIds.vec().capacity();
-    rScene.m_matPhong.ints()    .resize(maxEnts);
```
Capital-Asterisk, line 136: looks like this was oversized by 64x previously

Capital-Asterisk (approved):
looks good!

PR #287, opened 2024-05-29
Replace Newton Dynamics with Jolt Physics

Lamakaio:
I am making this PR to track the work on replacing Newton Dynamics with Jolt Physics. 
If it's not the right way to track things, feel free to remove it.

- [x] Add JoltPhysics to the repo as a submodule
- [x] Make it compile with the project
- [x] Write the interface to Jolt, similar to [ospnewton](https://github.com/TheOpenSpaceProgram/osp-magnum/tree/master/src/ospnewton)
- [x] Write the Session for ospjolt, similar to [newton](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/master/src/testapp/sessions/newton.cpp)
- [x] Change scenarios to use Jolt instead of Newton
- [x] Fix remaining bugs
- [x] Find out what I forgot in this list


jonesmz:
Really great start to this! I applaud your communication and consistency thereof!

.gitmodules
```diff
@@ -37,3 +37,6 @@
 [submodule "3rdparty/freetype"]
 	path = 3rdparty/freetype
 	url = https://github.com/freetype/freetype
+[submodule "3rdparty/JoltPhysics"]
+	path = 3rdparty/JoltPhysics
+	url = git@github.com:jrouwe/JoltPhysics.git
```
jonesmz, line 42: If you didn't already do this, please make sure you track an "official" release by git-tag, if that exists for Jolt.

For all the other submodules, except the Magnum graphics related ones, I Just track whatever the latest official release tag is.

For magnum, they just kind of stopped doing releases, so i just periodically bump to the latest commit.

src/adera/activescene/VehicleBuilder.cpp
```diff
@@ -201,15 +201,15 @@ VehicleData VehicleBuilder::finalize_release()
 
         // reserve node-to-machine partitions
         rPerNodeType.nodeToMach.data_reserve(rPerNodeType.m_connectCountTotal);
-        for (NodeId node : rPerNodeType.nodeIds.bitview().zeros())
+        for (NodeId node : rPerNodeType.nodeIds)
```
jonesmz, line 204: I don't disagree with this change, but I'm not really sure how related to the Jolt physics change it is.

Maybe this should be in it's own stand-alone PR?

Lamakaio: It is ! Sorry it is not the cleanest, I have two other PRs open and this is one of them. I kinda figured I would start with the most advanced version of the code I could, as this PR will take me quite some time anyway.

.gitmodules
Lamakaio: It should be tracking v5.0.0 ? hopefully

Lamakaio:
I am mostly finished here, but I get some weird behavior in Release mode only. Valgrind doesn't tell me anything, so I'm not sure where to look. If I could get some additional testing / a review it would be great.  

Jolt produces compilation warnings on Windows with Wall, so I disabled warnings in Jolt (at least, hopefully, I don't have MSVC to test it locally).

Currently I left all the Newton code up, and examples simply have to include Jolt.h or Newton.h to switch physics engine. I'm not sure how relevant that is, though.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+    {
+        using UserData_t = std::array<void*, 6u>;
+        using Func_t = void (*)(JoltBodyId joltBodyId, OspBodyId ospBodyId, ACtxJoltWorld const&, UserData_t, osp::Vector3&, osp::Vector3&) noexcept;
+
+        Func_t      m_func;
+        UserData_t  m_userData;
+    };
+    // The default values are the one suggested in the Jolt hello world exemple for a "real" project.
+    // It might be overkill here.
+    ACtxJoltWorld(  int threadCount = 2,
```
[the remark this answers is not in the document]
Lamakaio: Eh should I fix this ? It was the same in the original code, the pointer is overwritten before being used.

Capital-Asterisk (approved):
Nice work! I'd assume the Newton submodule should be kept around for benchmarking? Fairly simple to just make a copy of each test scenario. I don't mind this being merged in a non-final state as long as it mostly works; I'd like to figure out how to get around doing planet terrain colliders.

src/testapp/sessions/jolt.cpp
```diff
@@ -0,0 +1,747 @@
...
+                bodyCreation.mOverrideMassProperties = EOverrideMassProperties::MassAndInertiaProvided;
+            }
+            else
+            {   
+                bodyCreation.mMotionType = EMotionType::Static;
+                bodyCreation.mObjectLayer = Layers::MOVING;
+            }
+            PhysicsSystem *pJoltWorld = rJolt.m_world.get();
+            BodyInterface &bodyInterface = pJoltWorld->GetBodyInterface();
+            JoltBodyId joltBodyId = bodyInterface.CreateAndAddBody(bodyCreation, EActivation::Activate);
```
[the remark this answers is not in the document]
Lamakaio: Combining the two IDs sounds neat, I'll look into that. 

I think the idea is that accessing a body through BodyInterface is thread-safe. Bodies also don't have individual mutexes, locking is managed globally, so it ends up being quite verbose to lock a body and do something with it. The speed difference is probably negligible, afaik body storage is just an array.

jonesmz:
> Nice work! I'd assume the Newton submodule should be kept around for benchmarking? Fairly simple to just make a copy of each test scenario. I don't mind this being merged in a non-final state as long as it mostly works; I'd like to figure out how to get around doing planet terrain colliders.

I think keeping newton for a short amount of time might be worth doing, but we should definitely be removing it if jolt is working at a similar level or better. Newton doesn't compile on a bunch of targets, and it's code is a bonkers mess. I'd rather just be done with it.

src/ospjolt/activescene/forcefactors.h
```diff
@@ -0,0 +1,39 @@
...
+#include <cstdint>
+
+namespace ospjolt
+{
+
+// Each rigid body is given 64 bits to enable/disable forces
+// These determine which physics calculations are required for a certain
+// rigid body, such as gravity, thurst, or aerodynamics.
+// Forces are assignable at runtime in ACtxJoltWorld::m_factors
+using ForceFactors_t = std::array<uint64_t, 1u>;
```
jonesmz, line 37: Why use an array of size one?

jonesmz: Is this to provide a strong typedef? or is the goal here to generically support any number of items in the array, but start with only 1?

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+
+#include <iostream>
+#include <cstdarg>
+#include <thread>
+
+// Callback for traces, connect this to your own trace function if you have one
+static void TraceImpl(const char *inFMT, ...)
+{
+	// Format the message
+	va_list list;
```
jonesmz, line 64: I'm very, very, very, wary of C++ code using va_list / va_start.

What's driving the use of a C-style variadic function instead of a C++ template function?

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+
+using OspBodyId = uint32_t;
+using JoltBodyId = BodyID;
+
+/**
+ * @brief The different physics layers for the simulation
+ */
+namespace Layers
+{
+	static constexpr ObjectLayer NON_MOVING = 0;
```
jonesmz, line 107: Is there a reason not to use an enum class for these constants? They look a lot like enum values?

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+ * SOFTWARE.
+ */
+#pragma once
+
+#include "forcefactors.h"
+
+#include <osp/activescene/basic.h>
+#include <osp/core/id_map.h>
+
+#include <Jolt/Jolt.h>
```
jonesmz, line 32: from https://github.com/jrouwe/JoltPhysics/blob/master/HelloWorld/HelloWorld.cpp
```
// The Jolt headers don't include Jolt.h. Always include Jolt.h before including any other Jolt header.
// You can use Jolt.h in your precompiled header to speed up compilation.
#include <Jolt/Jolt.h>
```

wow, what an absolutely bullshit decision.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+	virtual bool ShouldCollide(ObjectLayer inObject1, ObjectLayer inObject2) const override
+	{
+		switch (inObject1)
+		{
+		case Layers::NON_MOVING:
+			return inObject2 == Layers::MOVING; // Non moving only collides with moving
+		case Layers::MOVING:
+			return true; // Moving collides with everything
+		default:
+			JPH_ASSERT(false);
```
jonesmz, line 126: Outside the scope of this PR, but we should probably make a more advanced ASSERT implementation for OSP that automatically does fun things like generating coredumps, generating back traces, and so on.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+		case Layers::MOVING:
+			return true;
+		default:
+			JPH_ASSERT(false);
+			return false;
+		}
+	}
+};
+
+//Forward declaration
```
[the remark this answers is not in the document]
jonesmz: Having a comment for forward declarations can be helpful, but only if the comment explains what the class is or provides a reference to the file where it's defined, etc.

Otherwise I agree that there's no need to say that a forward decl is a forward decl.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+struct ACtxJoltWorld
+{
+
+    struct ForceFactorFunc
+    {
+        using UserData_t = std::array<void*, 6u>;
+        using Func_t = void (*)(JoltBodyId joltBodyId, OspBodyId ospBodyId, ACtxJoltWorld const&, UserData_t, osp::Vector3&, osp::Vector3&) noexcept;
+
+        Func_t      m_func;
+        UserData_t  m_userData;
```
jonesmz, line 231: Note that neither of these member variables will have any initialization, so if they are read before something sets them, the read values will be arbitrary.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+
+    TempAllocatorImpl                               m_temp_allocator;
+    ObjectLayerPairFilterImpl                       m_objectLayerFilter;
+    BPLayerInterfaceImpl                            m_bPLInterface;
+    ObjectVsBroadPhaseLayerFilterImpl               m_objectVsBPLFilter;
+    std::unique_ptr<JobSystem>                      m_joltJobSystem;
+
+    std::unique_ptr<PhysicsSystem>                  m_world;
+
+	std::unique_ptr<PhysicsStepListenerImpl>		m_listener;
```
jonesmz, line 288: indentation

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+    lgrn::IdRegistryStl<OspBodyId>                  m_bodyIds;
+    std::vector<ForceFactors_t>                     m_bodyFactors;
+    lgrn::IdSetStl<OspBodyId>                       m_bodyDirty;
+
+    std::vector<JoltBodyId>                         m_ospToJoltBodyId;
+    std::vector<osp::active::ActiveEnt>             m_bodyToEnt;
+    osp::IdMap_t<osp::active::ActiveEnt, OspBodyId> m_entToBody;
+
+    std::vector<ForceFactorFunc>                    m_factors;
+	ShapeStorage_t                                  m_shapes;
```
jonesmz, line 299: indentation

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,306 @@
...
+
+	//mandatory jolt initialization steps
+	//Should this be called on world creation ? not ideal in case of multiple worlds.
+	static void initJoltGlobal() 
+	{
+	
+		RegisterDefaultAllocator();
+
+		// Install trace and assert callbacks
+		Trace = TraceImpl;
```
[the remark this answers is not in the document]
jonesmz: for clarity, yes, but there are some situations where you can accidentally get something that isn't a function pointer that auto-converts to a function pointer, and either compiles when it shouldn't, or doesn't compile but the error message is incomprehensible.

src/ospjolt/activescene/joltinteg_fn.cpp
```diff
@@ -0,0 +1,254 @@
...
+        rCtxWorld.m_bodyIds.remove(bodyId);
+        rCtxWorld.m_bodyToEnt[bodyId] = lgrn::id_null<ActiveEnt>();
+        rCtxWorld.m_entToBody.erase(itBodyId);
+    }
+
+}
+
+TransformedShapePtr_t SysJolt::create_primitive(ACtxJoltWorld &rCtxWorld, osp::EShape shape)
+{
+    Shape* joltShape;
```
jonesmz, line 126: std::unique_ptr

src/ospjolt/activescene/joltinteg_fn.h
```diff
@@ -0,0 +1,138 @@
...
+class SysJolt
+{
+    using ActiveEnt                 = osp::active::ActiveEnt;
+    using ACtxPhysics               = osp::active::ACtxPhysics;
+    using ACtxSceneGraph            = osp::active::ACtxSceneGraph;
+    using ACompTransform            = osp::active::ACompTransform;
+    using ACompTransformStorage_t   = osp::active::ACompTransformStorage_t;
+public:
+
+    using JoltThreadIndex_t = int;
```
jonesmz, line 50: should indicies ever be negative?

src/testapp/scenarios.cpp
```diff
@@ -152,10 +152,10 @@ static ScenarioMap_t make_scenarios()
         droppers        = setup_droppers            (builder, rTopData, scene, commonScene, physShapes);
         bounds          = setup_bounds              (builder, rTopData, scene, commonScene, physShapes);
 
-        newton          = setup_newton              (builder, rTopData, scene, commonScene, physics);
-        nwtGravSet      = setup_newton_factors      (builder, rTopData);
-        nwtGrav         = setup_newton_force_accel  (builder, rTopData, newton, nwtGravSet, sc_gravityForce);
-        physShapesNwt   = setup_phys_shapes_newton  (builder, rTopData, commonScene, physics, physShapes, newton, nwtGravSet);
+        jolt          = setup_jolt              (builder, rTopData, scene, commonScene, physics);
```
jonesmz, line 155: alignment / indentation

src/testapp/scenarios.cpp
```diff
@@ -231,13 +231,13 @@ static ScenarioMap_t make_scenarios()
...
         machRcsDriver   = setup_mach_rcsdriver      (builder, rTopData, scene, parts, signalsFloat);
 
-        newton          = setup_newton              (builder, rTopData, scene, commonScene, physics);
-        nwtGravSet      = setup_newton_factors      (builder, rTopData);
-        nwtGrav         = setup_newton_force_accel  (builder, rTopData, newton, nwtGravSet, sc_gravityForce);
-        physShapesNwt   = setup_phys_shapes_newton  (builder, rTopData, commonScene, physics, physShapes, newton, nwtGravSet);
-        vehicleSpawnNwt = setup_vehicle_spawn_newton(builder, rTopData, application, commonScene, physics, prefabs, parts, vehicleSpawn, newton);
-        nwtRocketSet    = setup_newton_factors      (builder, rTopData);
-        rocketsNwt      = setup_rocket_thrust_newton(builder, rTopData, scene, commonScene, physics, prefabs, parts, signalsFloat, newton, nwtRocketSet);
+        jolt          = setup_jolt              (builder, rTopData, scene, commonScene, physics);
```
jonesmz, line 234: alignment / indentation

src/testapp/sessions/jolt.cpp
```diff
@@ -0,0 +1,747 @@
...
+Session setup_jolt(
+        TopTaskBuilder&             rBuilder,
+        ArrayView<entt::any> const  topData,
+        osp::Session const&         scene,
+        Session const&              commonScene,
+        Session const&              physics)
+{
+    //Mandatory Jolt setup steps (start of program)
+    ACtxJoltWorld::initJoltGlobal();
+	JPH_IF_ENABLE_ASSERTS(AssertFailed = AssertFailedImpl;)
```
jonesmz, line 62: indentation

jonesmz:
To fix the msvc analyzer workflow, we'll need to link against the debug build of the microsoft libraries when building the Debug config.

See: https://stackoverflow.com/questions/35793622/lnk2019unresolved-external-symbol-crtdbgreport-referenced-in-function-crt-rtc

That's controlled by https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

but it doesn't seem to be working properly?

src/ospjolt/activescene/forcefactors.h
```diff
@@ -0,0 +1,39 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2024 Open Space Program Project
```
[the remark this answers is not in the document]
Lamakaio: I copied some things straight from the forcefactor.h in ospnewton which starts in 2019, but honestly I have no idea either.

src/ospjolt/activescene/forcefactors.h
Lamakaio: (not my code, so I can't be 100% sure) 
I think the intent is to be able to have a bigger array, so that more than 64 forces can be supported. Currently some things break if the array is bigger though.

src/ospjolt/activescene/joltinteg.h
Lamakaio: Jolt expects "ObjectLayer" values, which is basically u16. Is there a way to make an enum which can be compared against another type easily in c++ ? For exemple in a switch statement

src/ospjolt/activescene/joltinteg.h
Lamakaio: Somehow, there don't seem to be another way. 

Basically, I updated it to use spdlog, and the problem is, the spdlog function takes 2 arguments then variadics, and jolt expects 1 argument then variadics. So the only way seem to be to "unwrap" the variadics using vsnprintf, then call spdlog::trace.
I just looked into it and I'm pretty annoyed there isn't a simpler way. 

Template function can't be used here, as Jolt expects a generic function pointer that can be used for any type of values.

src/ospjolt/activescene/joltinteg_fn.cpp
```diff
@@ -0,0 +1,254 @@
...
+    return std::make_unique<TransformedShape>(RVec3::sZero(), Quat::sZero(), joltShape, BodyID(), SubShapeIDCreator());
+}
+
+void SysJolt::orient_shape(TransformedShapePtr_t& pJoltShape, osp::EShape ospShape, osp::Vector3 const &translation, osp::Matrix3 const &rotation, osp::Vector3 const &scale)
+{
+    auto rawQuat = Magnum::Quaternion::fromMatrix(rotation).data();
+    Quat joltRotation(rawQuat[0], rawQuat[1], rawQuat[2], rawQuat[3]);
+
+    //scale the shape directly
+    pJoltShape->mShape = pJoltShape->mShape->ScaleShape(Vec3Arg(scale.x(), scale.y(), scale.z())).Get();
```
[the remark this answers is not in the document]
Lamakaio: Is the compound shape (collider) transform supposed to be absolute ? I assumed it was relative to the compound shape transform (mostly because the recursive function starts with identity transform)
That is, the final absolute transform needs to be "compound shape transform" * "relative subshape transform".

I do agree that there are better ways to handle this though. I'll scrap orient_shape and create_primitive to make something that only adds ScaledShape wrapper when it is needed (ie when we are not creating the shape in the same spot, and when the shape is not already scaled).

Lamakaio:
This should be mostly done, except for the windows build issue, which I can't seem to find a solution to. If no new idea arise, I will probably setup the project on my windows machine to do some more in-depth testing.

src/ospjolt/activescene/joltinteg.h
```diff
@@ -0,0 +1,349 @@
...
+
+    uint GetNumBroadPhaseLayers() const override
+    {
+        return BroadPhaseLayers::NUM_LAYERS;
+    }
+
+    BroadPhaseLayer GetBroadPhaseLayer(ObjectLayer inLayer) const override
+    {
+        JPH_ASSERT(inLayer < Layers::NUM_LAYERS);
+        return mObjectToBroadPhase[inLayer];
```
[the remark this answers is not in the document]
Lamakaio: Looks like this is covered by the assert?

The only constructor initializes the array up to Layers::NUM_LAYERS.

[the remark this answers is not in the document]
Capital-Asterisk: I've seen clang-tidy do similar things on custom asserts. Under the 'show paths' button, it says it's assuming that the condition `inLayer < Layers::NUM_LAYERS` is false, and doesn't know the program should just panic in that case.

.gitmodules
alex-sherwin: The Jolt submodule is set to use a git+ssh protocol, and this only works if your client is setup with a valid GitHub SSH credential (e.g. it cannot be cloned anonymously this way)

Like the other dependencies I'm assuming this should be using the HTTPS URI of `https://github.com/jrouwe/JoltPhysics.git` so that it can be cloned anonymously

Lamakaio:
I did one more pass on the code and rebased it. I feel like it's pretty much done on my end.

Capital-Asterisk:
One thing I forgot to note, is that the rockets in the vehicle test scenario have a bit of a pendulum effect on them. Something is up with the center-of-mass or torque calculations. I don't mind this being fixed later.

does this PR still need WIP in the title?

src/ospjolt/activescene/joltinteg.h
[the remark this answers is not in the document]
jonesmz: There's specifically a function attribute to tag functions like that as "If you get here, assume everything after it is unreachable".

but i don't think there's an attribute like that for MSVC... (if someone knows about one, please tell me, as i really would like to use it at work...)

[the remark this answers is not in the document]
jonesmz: Note that these code scanners are scanning based on MSVC, not clang / gcc. 

Github requires "Sarif" output to be able to ingest analyzer results, and clang / gcc don't support that format for their analyzer tools.

jonesmz:
I have no objection to merging this code as is.

@Lamakaio can you please do three (relatively) minor things so we can merge it?

1. Change the title to remove "wip".
2. Change the PR from draft to normal.
3. Rebase against master one last time, and this time squash/fixup your commits into a more "publicly facing" format. You don't need to squish everything into a single commit, just try to reduce the quantity, and maybe add some more elaborated commit messages if you feel like it.

Lamakaio:
Absolutely ! I'll do that tomorrow.

Lamakaio:
Should be all done. I also fixed the pendulum effect, there was a stray minus sign in the center of mass calculations.

jonesmz:
Ok, passing CI, approval from CapitalAsterisk and myself, and nice commit organization.

Lets merge it!

PR #293, opened 2024-06-17
Planeta again

Capital-Asterisk:
Major terrain-related changes:
* Add session to upload terrain mesh to GPU
* Add temporary 'heightmap' function to terrain. (two cosine waves added together)
* Increase number of terrain skeleton subdivision level from 10 to 24 
* Rejiggle terrain test scenario(s): Earth-sized planet for `terrain` scenario, and tiny planet for `terrain_small` scenario
* A bunch of internal changes to planeta that are difficult to explain, to make the stuff above actually work.

Non-terrain-related changes:
* Add `src/osp/core/buffer_format.h`, based on and improved from the "StrideDesc" nonsense from `src/osp/universe/universe.h`

### DIFF
--- a/src/osp/core/array_view.h
+++ b/src/osp/core/array_view.h
@@ -27,6 +27,7 @@
 // IWYU pragma: begin_exports
 #include <Corrade/Containers/ArrayView.h>
 #include <Corrade/Containers/StridedArrayView.h>
+#include <Corrade/Containers/StridedArrayViewStl.h>
 // IWYU pragma: end_exports
 
 #include <iterator>
--- a/src/osp/core/array_view.h
+++ b/src/osp/core/array_view.h
@@ -38,6 +38,8 @@ namespace osp
 using Corrade::Containers::ArrayView;
 using Corrade::Containers::arrayView;
 
+using Corrade::Containers::arrayCast;
+
 /**
  * @brief Wraps a Corrade ArrayView or StridedArrayView to use as a 2D array of equally sized rows
  */
--- a/src/osp/core/array_view.h
+++ b/src/osp/core/array_view.h
@@ -26,6 +26,7 @@
 
 // IWYU pragma: begin_exports
 #include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Containers/StridedArrayView.h>
 // IWYU pragma: end_exports
 
 #include <iterator>
@@ -36,34 +37,52 @@ namespace osp
 using Corrade::Containers::ArrayView;
 using Corrade::Containers::arrayView;
 
+/**
+ * @brief Wraps a Corrade ArrayView or StridedArrayView to use as a 2D array of equally sized rows
+ */
 template<typename T>
 struct ArrayView2DWrapper
 {
-    constexpr ArrayView<T> row(std::size_t const index) const noexcept
+    // yes, I know that StridedArrayView2D exists, but ".row(...)" is more explicit
+
+    constexpr T row(std::size_t const columnIndex) const noexcept
     {
-        return view.sliceSize(index * columns, columns);
+        return view.sliceSize(columnIndex * rowSize, rowSize);
     }
 
-    ArrayView<T> view;
-    std::size_t columns;
+    T view;
+    std::size_t rowSize; ///< Size of each row, same as the number of columns
 };
 
+/**
+ * @brief Returns an interface that treats an ArrayView as a 2D array of equally sized rows.
+ *
+ * This overload auto-converts ArrayView-compatible types.
+ */
 template<typename T>
-constexpr ArrayView2DWrapper<T> as_2d(ArrayView<T> view, std::size_t columns) noexcept
+    requires requires (T &view) { osp::arrayView(view); }
+constexpr decltype(auto) as_2d(T &view, std::size_t rowSize) noexcept
 {
-    return { .view = view, .columns = columns };
+    return ArrayView2DWrapper<decltype(osp::arrayView(view))>{ .view = osp::arrayView(view), .rowSize = rowSize };
 }
 
 /**
- * @brief slice_2d_row
- * @param view
- * @param index
- * @param size
+ * @brief Returns an interface that treats an ArrayView as a 2D array of equally sized rows.
  */
 template<typename T>
-ArrayView<T> slice_2d_row(ArrayView<T> const& view, std::size_t index, std::size_t size) noexcept
+constexpr ArrayView2DWrapper< ArrayView<T> > as_2d(ArrayView<T> view, std::size_t rowSize) noexcept
 {
-    return view.sliceSize(index, size);
+    return { .view = view, .rowSize = rowSize };
+}
+
+/**
+ * @brief Returns an interface that treats an ArrayView as a 2D array of equally sized rows.
+ */
+template<typename T>
+constexpr ArrayView2DWrapper< Corrade::Containers::StridedArrayView1D<T> >
+        as_2d(Corrade::Containers::StridedArrayView1D<T> view, std::size_t rowSize) noexcept
+{
+    return { .view = view, .rowSize = rowSize };
 }
 
 /**
--- a/src/osp/core/buffer_format.h
+++ b/src/osp/core/buffer_format.h
@@ -41,8 +41,8 @@ struct BufAttribFormat
 {
     using View_t        = Corrade::Containers::StridedArrayView1D<T>;
     using ViewConst_t   = Corrade::Containers::StridedArrayView1D<T const>;
-    using Data_t        = Corrade::Containers::ArrayView<unsigned char>;
-    using DataConst_t   = Corrade::Containers::ArrayView<unsigned char const>;
+    using Data_t        = Corrade::Containers::ArrayView<std::byte>;
+    using DataConst_t   = Corrade::Containers::ArrayView<std::byte const>;
 
     constexpr View_t view(Data_t data, std::size_t count) const noexcept
     {
--- a/src/osp/core/buffer_format.h
+++ b/src/osp/core/buffer_format.h
@@ -1,0 +1,139 @@
+/**
+ * Open Space Program
+ * Copyright © 2019-2024 Open Space Program Project
+ *
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+#pragma once
+
+#include <Corrade/Containers/Array.h>
+#include <Corrade/Containers/StridedArrayView.h>
+
+namespace osp
+{
+
+/**
+ * @brief Buffer Attribute Format. Describes how to access element attribute data within a buffer.
+ *
+ * This is useful for SIMD, GPU, and serialization. A SIMD n-body simulation may prefer [XXXYYYZZZ]
+ * to store positions, but GPU mesh vertex positions prefer [XYZXYZXYZ...].
+ */
+template <typename T>
+struct BufAttribFormat
+{
+    using View_t        = Corrade::Containers::StridedArrayView1D<T>;
+    using ViewConst_t   = Corrade::Containers::StridedArrayView1D<T const>;
+    using Data_t        = Corrade::Containers::ArrayView<unsigned char>;
+    using DataConst_t   = Corrade::Containers::ArrayView<unsigned char const>;
+
+    constexpr View_t view(Data_t data, std::size_t count) const noexcept
+    {
+        return stridedArrayView<T>(data, reinterpret_cast<T*>(&data[offset]), count, stride);
+    }
+
+    constexpr ViewConst_t view_const(DataConst_t data, std::size_t count) const noexcept
+    {
+        return stridedArrayView<T const>(data, reinterpret_cast<T const*>(&data[offset]), count, stride);
+    }
+
+    constexpr bool is_not_used() const noexcept { return stride == 0; }
+
+    std::size_t     offset{};
+    std::ptrdiff_t  stride{};
+};
+
+/**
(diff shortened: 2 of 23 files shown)


src/planet-a/chunk_utils.h
```diff
@@ -251,10 +232,10 @@ class ChunkFillSubdivLUT
 
     void fill_tri_recurse(Vector2us top, Vector2us lft, Vector2us rte, std::uint8_t level);
 
-    //std::unique_ptr<ToSubdiv[]> m_data;
     std::vector<ToSubdiv> m_data;
-    std::uint16_t m_fillVrtxCount{0};
-    std::uint16_t m_edgeVrtxCount{0};
+
+    std::uint16_t m_fillVrtxCount{};
+    std::uint16_t m_edgeVrtxCount{};
```
Capital-Asterisk, line 238: Empty braces like this will zero-initialize (initialize to zero) to prevent an uninitialized value.

I don't have this in guidelines.md, but I'd intend that empty braces means "don't care what this value is, but don't keep it uninitialized," whereas `{0}` would imply that the zero is important.

src/osp/core/buffer_format.h
```diff
@@ -0,0 +1,139 @@
...
+     * @endcode
+     *
+     * @param count [in] Number of elements
+     */
+    template <typename ... T>
+    constexpr void insert_interleave(std::size_t count, BufAttribFormat<T>& ... rInterleve)
+    {
+        constexpr std::size_t stride = (sizeof(T) + ...);
+
+        (rInterleve.m_stride = ... = stride);
```
Lamakaio, line 111: Seem to be a typo in multiple variables in this file, interleve -> interleave

Lamakaio (approved):

src/testapp/sessions/magnum.cpp
```diff
@@ -546,4 +549,132 @@
...
 
+
+
+struct ACtxDrawTerrainGL
+{
+    Magnum::GL::Buffer  vrtxBufGL{Corrade::NoCreate};
+    Magnum::GL::Buffer  indxBufGL{Corrade::NoCreate};
+    MeshGlId            terrainMeshGl;
+    bool                enabled{false};
+};
```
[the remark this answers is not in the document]
Capital-Asterisk: changed MeshGlId from an enum class to an osp::StrongId, where it will be defaulted to null properly.

Lamakaio (approved):

src/osp/core/array_view.h
```diff
@@ -36,34 +38,52 @@ namespace osp
 using Corrade::Containers::ArrayView;
 using Corrade::Containers::arrayView;
 
+/**
+ * @brief Wraps a Corrade ArrayView or StridedArrayView to use as a 2D array of equally sized rows
+ */
 template<typename T>
 struct ArrayView2DWrapper
```
jonesmz, line 45: Since this class is only supposed to be used with ArrayView or StridedArrayView, you might want to consider adding template specializations for only those two, and leaving the "default" template declaration of the struct undefined.

The lazy way of just leaving it like this is also fine, of course.

jonesmz: Or a requires clause.

either way.

src/osp/core/buffer_format.h
```diff
@@ -0,0 +1,139 @@
...
+
+    constexpr ViewConst_t view_const(DataConst_t data, std::size_t count) const noexcept
+    {
+        return stridedArrayView<T const>(data, reinterpret_cast<T const*>(&data[offset]), count, stride);
+    }
+
+    constexpr bool is_not_used() const noexcept { return stride == 0; }
+
+    std::size_t     offset{};
+    std::ptrdiff_t  stride{};
```
jonesmz, line 60: empty-brace-init instead of = 0 ?

src/osp/core/math_int64.h
```diff
@@ -54,25 +54,14 @@ constexpr std::uint64_t absdelta(std::int64_t lhs, std::int64_t rhs) noexcept
 /**
  * @brief (distance between a and b) > threshold
  *
- * This function is quick and dirty, max threshold is limited to 1,431,655,765
  */
-constexpr bool is_distance_near(Vector3l const a, Vector3l const b, std::uint64_t const threshold) noexcept
+constexpr bool is_distance_near(Vector3l const a, Vector3l const b, double const threshold) noexcept
```
jonesmz, line 58: why switch to double?

src/osp/core/math_int64.h
```diff
@@ -54,25 +54,14 @@ constexpr std::uint64_t absdelta(std::int64_t lhs, std::int64_t rhs) noexcept
...
 {
-    std::uint64_t const dx = absdelta(a.x(), b.x());
-    std::uint64_t const dy = absdelta(a.y(), b.y());
-    std::uint64_t const dz = absdelta(a.z(), b.z());
+    double const dx( absdelta(a.x(), b.x()) );
+    double const dy( absdelta(a.y(), b.y()) );
+    double const dz( absdelta(a.z(), b.z()) );
```
jonesmz, line 62: Personally i don't find switching from assignment initialization to braced initialization easier to read.

Are you doing this to force an error if the type of the result of absdelta is not the same as the type of the variable?

src/osp/core/math_int64.h
```diff
@@ -54,25 +54,14 @@ constexpr std::uint64_t absdelta(std::int64_t lhs, std::int64_t rhs) noexcept
...
-
-    if (dx > dmax || dy > dmax || dz > dmax)
-    {
-        return false;
-    }
-
-    std::uint64_t const magnitudeSqr = (dx*dx + dy*dy + dz*dz);
-
-    return magnitudeSqr < threshold*threshold;
+    return (dx*dx + dy*dy + dz*dz) < threshold*threshold;
```
jonesmz, line 64: You could check and see if the codegen is better or worse when using https://en.cppreference.com/w/cpp/numeric/math/hypot here.

jonesmz: I could def see an argument that using the sum of squares is faster than doing a square root, so i have no concerns about this code. Just pointing out that there's a special function for a different way to slice this.

src/planet-a/chunk_utils.cpp
```diff
@@ -56,9 +56,17 @@ void ChunkFillSubdivLUT::subdiv_line_recurse(
 {
     Vector2us const mid = (a + b) / 2;
 
-    ChunkLocalSharedId const out = ChunkLocalSharedId(xy_to_triangular(mid.x() - 1, mid.y() - 2));
-
-    m_data.emplace_back(ToSubdiv{id_at(a), id_at(b), out});
+    int                const out     = xy_to_triangular(mid.x() - 1, mid.y() - 2);
```
jonesmz, line 59: not uint32_t ?

src/planet-a/chunk_utils.h
jonesmz: Kind of a philosophical question here, but is there any *actual* difference between "Don't care what the value is, but don't leave it uninitialized" and "don't initialize" ?

Using empty-brace initialization for variables where you don't care what the resulting value is means that automated tools like clang-tidy / msvc-analyzer can't identify "read-from-uninitialized" bugs in the code, which i think is a big risk.

jonesmz: Note that I have no problem with the variables here being initialized to zero if that's what's intended. But "initialize them, but i don't care to what" is, in my opinion, worse than 'don't initialize'.

src/planet-a/geometry.h
```diff
@@ -83,9 +84,14 @@ struct BasicChunkMeshGeometry
 {
     void resize(ChunkSkeleton const& skCh, ChunkMeshBufferInfo const& info);
 
-    std::vector<osp::Vector3>           chunkVbufPos;
-    std::vector<osp::Vector3>           chunkVbufNrm;
-    std::vector<osp::Vector3u>          chunkIbuf;
+    Corrade::Containers::Array<unsigned char> vrtxBuffer; ///< Output vertex buffer
```
jonesmz, line 87: `unsigned char`, specifically?

not `std::byte`, or just `char`?

src/planet-a/icosahedron.h
```diff
@@ -91,31 +91,44 @@ inline constexpr std::array<std::array<uint8_t, 3>, 20> gc_icoNeighbors
 /**
  * @brief Icosahedron Minimum Edge Length vs Subdiv Levels, radius = 1.0
  */
-inline constexpr std::array<float, 10> const gc_icoMinEdgeVsLevel
+inline constexpr std::array<float, 24> const gc_icoMinEdgeVsLevel
 {
-    1.05146222e+0f,  5.46533058e-1f,  2.75904484e-1f,  1.38283174e-1f,  6.91829904e-2f,
-    3.45966718e-2f,  1.72989830e-2f,  8.64957239e-3f,  4.32479631e-3f,  2.16239942e-3f
+    1.05146222e+0f,  5.46533058e-1f,  2.75904484e-1f,  1.38283174e-1f,
```
jonesmz, line 96: could you document where the initial values of this are from? Is it `icosahedron_tables.py` ?

src/planet-a/icosahedron.h
```diff
@@ -91,31 +91,44 @@ inline constexpr std::array<std::array<uint8_t, 3>, 20> gc_icoNeighbors
...
 /**
  * @brief Icosahedron Maximum Edge Length vs Subdiv Levels, radius = 1.0
  */
-inline constexpr std::array<float, 10> const gc_icoMaxEdgeVsLevel
-{
-    1.05146222e+0f,  6.18033989e-1f,  3.24919696e-1f,  1.64647160e-1f,  8.26039665e-2f,
-    4.13372560e-2f,  2.06730441e-2f,  1.03370743e-2f,  5.16860619e-3f,  2.58431173e-3f
-};
+inline constexpr std::array<float, 24> const gc_icoMaxEdgeVsLevel
+{{
```
jonesmz, line 108: double brace init here? that seems like maybe a typo?

src/testapp/scenarios.cpp
```diff
@@ -319,18 +319,107 @@ static ScenarioMap_t make_scenarios()
...
-            .height                 = 5.0,
+            .radius                 = 6371000.0, // Earth radius
+            .height                 = 20000.0,   // Height between Mariana Trench and Mount Everest
+            .skelPrecision          = 10,        // 2^10 units = 1024 units = 1 meter
+            .skelMaxSubdivLevels    = 19,
+            .chunkSubdivLevels      = 4
+        });
+
+        // Set scene position relative to planet to be just on the surface
+        rTerrainFrame.position = Vector3l{0,0,6371000} * 1024;
```
jonesmz, line 339: minor: can the earth radius be extracted out to a constexpr variable for self-documentation purposes?

src/planet-a/chunk_utils.h
jonesmz: And, of course, this whole concept only applies to primative types. Class types / types with constructors implicitly manage their initial state.

jonesmz (approved):

src/osp/core/buffer_format.h
Capital-Asterisk: I think I mentioned this in another review comment: my intention for an empty brace is "don't care what this value is, just not uninitialized." A zero explicitly indicates that the value is important to some business logic somewhere.

src/osp/core/array_view.h
Capital-Asterisk: I guess you mean a concept for array views?

I'll be lazy on this as there's already a short comment explaining what this type is, and it's a little too simple.

src/osp/core/math_int64.h
Capital-Asterisk: The previous implementation here just doesn't work for the Earth-sized distances. I'd rather keep it simple and not write something ultra-optimized here.

src/osp/core/math_int64.h
Capital-Asterisk: I agree. Not particularly a sane move to do this on my end. Curly braces will enforce the type; however, these are those error-prone round braces that have a silly chance of being interpreted as a function declaration. I'm intentionally implicit-casting int64 from absdelta to a double.

src/osp/core/buffer_format.h
Capital-Asterisk: oh, it's this one: https://github.com/TheOpenSpaceProgram/osp-magnum/pull/293/files#r1642043900

src/planet-a/chunk_utils.h
Capital-Asterisk: > Kind of a philosophical question here, but is there any actual difference between "Don't care what the value is, but don't leave it uninitialized" and "don't initialize" ?

If I look at code that looks like this:

```cpp
int value;
do_something(value);
```

`value` is uninitialized and there may be weird inconsistent errors when it's used wrong. static analysis may also scream at it. However, the intention is clear that the initial value of `value` is dont-care.


If I instead do...

```cpp
int value = 0;
do_something(value);
```
there's no risk of using uninitialized values, but the zero looks significant, as if it was a start of a count for example. 


Instead using...
```cpp
int value{};
do_something(value);
```
`value` would be zero-initialized, there's no risk of any weird inconsistent errors. The intention is default or don't-care, nothing is explicitly specified.



> Using empty-brace initialization for variables where you don't care what the resulting value is means that automated tools like clang-tidy / msvc-analyzer can't identify "read-from-uninitialized" bugs in the code, which i think is a big risk.

Are you implying that uninitialized values can help find logic errors?

I believe we should avoid uninitialized values whenever possible. Using uninitialized memory as part of regular calculation logic something I'd mostly consider a bad idea nowadays.

PREfast also tends to scream at every uninitialized value.

Capital-Asterisk: I don't remember where I originally learned this from. I did find this recent similar discussion through google search results: https://www.reddit.com/r/cpp/comments/1dm2y2t/is_empty_brace_initialization_considered_best/

Since I'm noticing I'm throwing around the term "don't care" quite a bit, I would note that this is a term from digital logic; I'm not sure if it's used in other parts of computing.

jonesmz: > Are you implying that uninitialized values can help find logic errors?

Not quite.

I'm saying that from the perspective of a static analyzer

```
int bob;
```
and
```
int alice{};
```

are very different looking.

If the programmer doesn't care about the value, then the programmer doesn't want the value to be read from until something is put there. A static analyzer (e.g. PREfast) can detect this kind of problem with `bob`, but `alice` looks initialized.

My advice is that empty brace initialization for primitive types is a bad idea exactly because of this, and explicitly setting a value is better.

I'd be interested to see an example where PREfast was saying "read from uninitialized!" when that wasn't happening. I suppose it could happen with non-const reference args to functions in other translation units, but that seems like a bug in PREfast that it isn't going and looking at the other translation unit.

Capital-Asterisk: >> Are you implying that uninitialized values can help find logic errors?
>
> Not quite.

The further explanation makes this a yes? In your explanation, `bob` is kept uninitialized to use it to help find a potential logic error.

So from your point of view, would this example also be unacceptable?
```cpp
int value = 0; // This can be anything, will be overwritten immediately
do_something(value);
```

Since this is more about using uninitialized memory (less about zero-initialization), we're not the first to have this discussion:

https://softwareengineering.stackexchange.com/questions/305886/doesnt-always-initialize-variables-lead-to-important-bugs-being-hidden

I'm mostly on the side of just disallowing uninitialized memory. Undefined behaviour that does stuff like being different between debug vs release is far harder to debug IMO, compared to accidentally reading a zero.

Clang-tidy does have an option for this: `cppcoreguidelines-init-variables`

> I'd be interested to see an example where PREfast was saying "read from uninitialized!" when that wasn't happening....

This alert is from this PR: 

https://github.com/TheOpenSpaceProgram/osp-magnum/security/code-scanning/333

It doesn't even care about whether the variable is being read or not. It just doesn't like uninitialized members.

src/planet-a/icosahedron.h
Capital-Asterisk: There's already comment; it's just not visible on the diff.

src/planet-a/icosahedron.h
Capital-Asterisk: I don't remember the exact details, but this is required for some of the other tables in this file for arrays of structs. This is harmless afaik

Issue #289, opened 2024-05-31
Decide on another Renderer


Capital-Asterisk:
PBR? Lights? Actual materials? Actual render passes? 

We currently use Magnum, which is a lightweight abstraction around OpenGL and Vulkan bundled with useful tools, not a rendering engine. Magnum is excellent if we want to write a paper on cool a new GPU rendering technique, but not when we just want to "put a cube here, put a light here."

We can search online for open source PBR shaders and everything else; however, there's so much more infrastructure required to tie everything together: shader compilation, supporting multiple devices, scene environment map, etc...

Do I (or most other contributors I presume) want to put an extra 2 years of effort into an already-solved problem that's not actually related to doing any space-related things better? Probably not.

## Other Rendering Engines

A practical solution arises: ***yoink an existing (open source) rendering engine***

OSP features full separation between scene stuff and rendering, which means it's really easy to write a new rendering backend. OSP may be considered a 'custom game engine' but the only issue is that GAME ENGINES AREN'T REAL! WAKE UP!

Looking around online, there are no open source 'rendering engines' out there on their own, only ones from existing game engines.

### Atom Renderer

I first looked at the [Atom renderer](https://www.docs.o3de.org/docs/atom-guide/) from O3DE. O3DE's platform support, massive codebase size, and high minimum requirements is a deal-breaker for me though.
 
### Godot RenderingServer
 
Dylan recently brought [Godot RenderingServer](https://docs.godotengine.org/en/stable/classes/class_renderingserver.html) to my attention. This apparently contains all of Godot's rendering capability and "bypasses the Scene/Node system entirely." The Scene/Node system is the thing I complain about the most. This aside, Godot is fairly lightweight, has good hardware & platform support, and a great community with lots of people using and testing it. I'm also kind of surprised how easy it is to download the editor, run, and export a project. This looks like our best option.

***Should Godot use OSP as a library or should OSP use Godot as a library?***

Ideally we should use Godot as a library, so we have access to the main function, command line interface, and control over running things in general. Imagine having an option to run a dedicated multiplayer server.

The main challenge here is that Godot is not designed to be used as a library. This is possible (to do without too much additional work) but is still under development: https://github.com/godotengine/godot/pull/90510 . There's also possible issues with how Godot relies on global singletons, so we can't have multiple instances or may have problems reopening it after its closed. 

[...]

PR #294, opened 2024-06-17
GDExtension build prototype

Lamakaio:
This is a heavy refactor of the project architecture and build commands to allow for the compilation of osp-magnum as a Godot GDExtension (see #289). It is very much a prototype intended for experimenting with Godot, not really a template of how things should be (names are whatever went through my head when I wrote them, some things are hardcoded...)

~~Currently I have not tested it, or even tried to make it work, for anything other than Linux x64.~~ It is also based on my Jolt PR (#287 ), and I yeeted all the Newton stuff, as trying to get it to compile and link properly did not look pleasant. ~~The unit tests were also giving me trouble so they're gone for now, I'll work on adding them back.~~ (done)

It adds a new `OSP_BUILD_GDEXTENSION` cmake flag, that when set, will compile osp as a shared library linked against godot-cpp, with (for the moment) the code from [this GDExtension template repo](https://github.com/asmaloney/GDExtensionTemplate) as a "main", with an added call to ospjolt to check that it works. (About the license stuff, the template is Unlicense)

~~For the project architecture, it splits the original code in two directories : `common` and `standalone`. `common` contains almost all the code from osp-magnum, with the exception of the `testapp`, and the opengl drawing stuff (`adera/drawing_gl`, renamed `shader`, and  `osp/drawing_gl`, renamed `drawing_gl`). Standalone contains the rest.~~

A `gdextension` directory was added, for the godot-specific code.

~~The dependencies were modified. Mostly, `common` depends on less features of Magnum, while `standalone` depends on all the graphic stuff + `common`. When `OSP_BUILD_GDEXTENSION` is set, Magnum is compiled without all the graphics stuff and the main sdl app.~~

I put a `CMakeLists.txt` in each subdirectory from the osp-magnum app, and additionally split the magnum-renderer specific code into additional directories `osp_drawing_gl`, `adera_drawing_gl` and `testapp_magnum`. Each directory only depends on what it needs to build. 

`gdextension` additionally depends on `godot-cpp`, which was added to the project as a submodule in `3rdparty`. 

Some libraries (Jolt and Freetype) were switched to static linking, as dynamic linking other libraires for a GDExtension is a bit of a pain. 

As a side note, Godot seems to dislike (some parts of) the C++ standard library. Notably, it doesn't allow exceptions, which the standard library makes use of. It will probably lead to some issues (for example, calling `std::call_once` crashes the application). All the threading stuff may need to be implemented using Godot threads.


Capital-Asterisk:
Off to a great start!

Do you have a good reason why restructure the directories into `common` and `standalone`? I'm not much of a fan of there being too much nesting. If OSP mostly already a library, why put it in a common directory?

My intention with the current subdirectory structure is that `adera`, `osp`, `ospnewton`, and `planet-a` are practically separate libraries; `testapp` is an executable that depends on all of these libraries.

Ideally, each one should have a separate CMakeLists.txt defining them as such, but that isn't the case due to laziness (low priority and not really required, until now?). I guess CMake namespaces can be used too (eg: `osp::adera`)?

Though I'd agree with separating out the OpenGL-dependent stuff. Maybe into `src/osp-magnum-gl` and `src/adera-magnum-gl`? maybe with underscores (this is a bit of an inconsistency right now)? or just a single library for all Magnum GL stuff? It's mostly for organization.



> Godot seems to dislike (some parts of) the C++ standard library...

I figured. I remember reading about similar issues while reading about the planned "LibGodot" feature to allow using Godot as a library. Exceptions are enabled in OSP but we really don't use them (I personally don't like exceptions); it's possible to just disable them with a build flag. If `std::call_once` has issues, I'd presume static local variables would likely have issues too. I find it likely that there's likely other GdExtension C++ projects out there using modern features with the standard library enabled, so I'm fairly confident so far that most issues here are fixable.

Lamakaio:
> Do you have a good reason why restructure the directories into common and standalone? I'm not much of a fan of there being too much nesting. If OSP mostly already a library, why put it in a common directory?

It was mostly due to the opengl stuff. Once I figured I couldn't keep each directory as it was, I just split it in two. Your solution makes a lot more sense. 

> If std::call_once has issues, I'd presume static local variables would likely have issues too.

I'm not sure what causes the issue. std::call_once can raise an exception, but well. 

From what I've seen the Godot people say, the stdc++ library is not officially supported, and they make no effort to test if the features work, but if you give them a very specific and well-documented bug with it, they'll fix it.

.gitmodules
```diff
@@ -37,3 +37,9 @@
 [submodule "3rdparty/freetype"]
 	path = 3rdparty/freetype
 	url = https://github.com/freetype/freetype
+[submodule "3rdparty/JoltPhysics"]
+	path = 3rdparty/JoltPhysics
+	url = https://github.com/jrouwe/JoltPhysics.git
+[submodule "3rdparty/godot-cpp"]
```
alex-sherwin, line 43: This should probably specify a branch (latest LTS? so 4.2 right now), the README for https://github.com/godotengine/godot-cpp?tab=readme-ov-file#versioning indicates tracking master is not a good idea unless you are building Godot from source to match

jonesmz:
Can you make a PR that depends on your Jolt PR, that does nothing but purge newton?

Not that said "remove newton" pr should be merged immediately, but the sooner we can see that it looks like the sooner we can get it done.

jonesmz:
> Ideally, each one should have a separate CMakeLists.txt defining them as such, but that isn't the case due to laziness (low priority and not really required, until now?). I guess CMake namespaces can be used too (eg: osp::adera)?

I agree with @Capital-Asterisk that these should be kept as "stand-alone" libraries unless we run into something that can't be resolved without merging them.

I also agree about the use of cmake namespaces.

jonesmz:
> Though I'd agree with separating out the OpenGL-dependent stuff. Maybe into src/osp-magnum-gl and src/adera-magnum-gl? maybe with underscores (this is a bit of an inconsistency right now)? or just a single library for all Magnum GL stuff? It's mostly for organization.

This seems like a fine suggestion to me.

jonesmz:
> As a side note, Godot seems to dislike (some parts of) the C++ standard library. Notably, it doesn't allow exceptions, which the standard library makes use of. It will probably lead to some issues (for example, calling std::call_once crashes the application). All the threading stuff may need to be implemented using Godot threads.

Well that's just shitty library/application design then.

People can dislike exceptions all they want (C++'s implementation of exceptions has plenty of issues, obviously), but they exist and turning them off for such a large project is just stupid, as it leads to exactly the situation you point out with `std::call_once`.

Though it's important to point out that `std::call_once` should *not* crash if exceptions are disabled, because `std::call_once` does not throw exceptions, it just lets the exceptions from the called-once function be propagated. If the provided function to call-once can't throw an exception, then `std::call_once` should have all the exception handling machinery pruned by the compilers dead-code elimination.

That it's crashing for you (without an exception being thrown, even) strongly indicates that there is an implementation bug in the specific version of the standard library / compiler you're using.

What standard library and compiler are you using? E.g. Operating system, compiler version, version of libstdc++ or libc++

> I figured. I remember reading about similar issues while reading about the planned "LibGodot" feature to allow using Godot as a library. Exceptions are enabled in OSP but we really don't use them (I personally don't like exceptions);

It's true that we don't, but we've been pretty careful to go out of our way to design all of our code in a "can't fail" kind of way, with notable exceptions (heh) of math operations where we ignore floating point issues (though we don't currently bind signals to exceptions, so the program would just crash here), and "well that'll never happen" kinds of situations where the app'll break if the "never happen" thing happens.

I think `std::expected` https://en.cppreference.com/w/cpp/utility/expected is a *massive* improvement over the current status quo, since it provides a mechanism to return error indicators without (almost) all of the terrible ergonomics issues that most previous solutions have had, and it can be used intelligently to bridge the gap between a library that doesn't ever throw exceptions (even going so far as to turn them off while the library is being compiled) and application code / other library code that does build with and understand exceptions.

> it's possible to just disable them with a build flag. If std::call_once has issues, I'd presume static local variables would likely have issues too. I find it likely that there's There are currently other GdExtension C++ projects out there using modern features with the standard library enabled, so I'm fairly confident so far that most issues here are fixable.

function-static local variables are initialized with a two-phase atomic "is initialized?" check.

The first phase is a non-atomic bool lookup (happy simple path, basically), the second is essentially an std::mutex.

https://godbolt.org/z/8bMfoo36W

jonesmz:
> Some libraries (Jolt and Freetype) were switched to static linking, as dynamic linking other libraires for a GDExtension is a bit of a pain.

I have no objection to this, but keep in mind that this does make it pretty difficult to pick up system libraries. Mileage may vary, since I've always wanted to make sure that 100% of our dependencies were always "in-tree", but others have been unhappy with that.

.gitmodules
jonesmz: I'm not fully up to date on how the `.gitmodules` file works so maybe a new feature was added since i last looked, but as far as I understand, there isn't a way to explicitly indicate a branch in this file, and you have to instead rely on specifying the exact commit you want to refer to.

If they've added "Give me the latest in branch X" support, or "Give me whatever tag Y points to" support, i'd be happy to hear about it.

3rdparty/CMakeLists.txt
```diff
@@ -222,30 +218,44 @@ ADD_SUBDIRECTORY(toml11 EXCLUDE_FROM_ALL)
 SET(SPDLOG_ENABLE_PCH ON CACHE BOOL "" FORCE)
 # Not yet supported by github runners
 # SET(SPDLOG_USE_STD_FORMAT ON CACHE BOOL "" FORCE)
+SET(SPDLOG_BUILD_PIC  ON CACHE BOOL "" FORCE)
```
jonesmz, line 221: Could you elaborate on this?

I don't mind it, but i'm curious to know your reasons.

3rdparty/CMakeLists.txt
```diff
@@ -222,30 +218,44 @@ ADD_SUBDIRECTORY(toml11 EXCLUDE_FROM_ALL)
...
-       )
+# SET(FT_DISABLE_ZLIB        ON CACHE BOOL "" FORCE)
+# SET(FT_DISABLE_BZIP2       ON CACHE BOOL "" FORCE)
+# SET(FT_DISABLE_PNG         ON CACHE BOOL "" FORCE)
+# SET(FT_DISABLE_HARFBUZZ    ON CACHE BOOL "" FORCE)
+# SET(FT_DISABLE_BROTLI      ON CACHE BOOL "" FORCE)
+# SET(SKIP_INSTALL_ALL       ON CACHE BOOL "" FORCE)
+# SET(SKIP_INSTALL_LIBRARIES ON CACHE BOOL "" FORCE)
+# SET(BUILD_SHARED_LIBS      OFF CACHE BOOL "" FORCE)
+# ADD_SUBDIRECTORY(freetype EXCLUDE_FROM_ALL)
```
jonesmz, line 237: if i recall correctly, we only need freetype for RmlUi.

So if we're going to go full steam on godot, then we can just drop RmlUi and freetype.

but in general, best not to leave commented out code in the source files.

3rdparty/CMakeLists.txt
```diff
@@ -222,30 +218,44 @@ ADD_SUBDIRECTORY(toml11 EXCLUDE_FROM_ALL)
...
+#         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+#         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+#         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+#        )
 
 SET(RML_SKIP_INSTALL ON CACHE BOOL "" FORCE)
 SET(CUSTOM_CONFIGURATION ON CACHE BOOL "" FORCE)
 #ADD_SUBDIRECTORY(RmlUi EXCLUDE_FROM_ALL)
+SET(ENABLE_ALL_WARNINGS OFF CACHE BOOL "" FORCE)
+SET(BUILD_SHARED_LIBS      OFF CACHE BOOL "" FORCE)
```
jonesmz, line 250: Oh lovely, Jolt doesn't namespace it's options? Rather selfish of them.

CMakeLists.txt
```diff
@@ -43,9 +43,35 @@ IF(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows" AND NOT CMAKE_SYSTEM_VERSION)
...
+#build universally on macOS
+if ( APPLE )
+    set( CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE STRING "" )
+endif()
+
 # Set project name
 PROJECT(OSP-MAGNUM CXX)
 
+# From: https://crascit.com/2016/04/09/using-ccache-with-cmake/
+find_program( CCACHE_PROGRAM ccache )
```
jonesmz, line 55: personally i don't think ccache specific things should live in the cmakelists.txt

In fact, this change will most likely screw up the ccache config that i (think...? did i? i guess i don't actually remember if i submitted it) setup in the continuous integration runners.

Instead, ccache support should live in a CMakePresets.json file

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

CMakeLists.txt
```diff
@@ -136,6 +162,8 @@ IF(MSVC)
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIG> CACHE PATH "")
   # Force multiprocessor compilation for all projects
   add_compile_options($<$<CXX_COMPILER_ID:MSVC>:/MP>)
+  # Force dynamic runtime library to be in debug mode in debug mode.
+  add_compile_options($<$<CONFIG:Debug>:$<$<CXX_COMPILER_ID:MSVC>:/MDd>>)
```
jonesmz, line 166: this is already supposed to be the behavior that cmake has by default.

https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

> If this variable is not set then the [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html#prop_tgt:MSVC_RUNTIME_LIBRARY) target property will not be set automatically. If that property is not set then CMake uses the default value MultiThreaded$<$<CONFIG:Debug>:Debug>DLL to select a MSVC runtime library.

CMakeLists.txt
```diff
@@ -156,9 +184,12 @@ ADD_SUBDIRECTORY(3rdparty)
 ADD_SUBDIRECTORY(src)
 
 # Unit Tests
-ADD_SUBDIRECTORY(test)
-
+IF (NOT OSP_BUILD_GDEXTENSION)
+    ADD_SUBDIRECTORY(test)
```
jonesmz, line 188: is there a problem building unit tests when building a GDEXTENSION ?

src/CMakeLists.txt
```diff
@@ -23,136 +23,32 @@
...
-            spdlog
-            longeron)
-target_link_libraries(osp-magnum-deps INTERFACE ${OSP_MAGNUM_DEPS_LIBS})
-add_dependencies(compile-osp-magnum-deps ${OSP_MAGNUM_DEPS_LIBS})
-
-add_executable(osp-magnum)
-
-target_compile_features(osp-magnum PUBLIC cxx_std_20)
+#compile as position-independant to avoid link errors on static link
+add_compile_options("-fPIC")
```
jonesmz, line 35: Use https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html instead

src/adera/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+# Link it to the interface library that describes the dependencies
+target_link_libraries(Adera PRIVATE ${DEPS_LIBS})
+
+# Gather paths to OSP headers and sources
+file(GLOB_RECURSE CPP_FILES CONFIGURE_DEPENDS *.cpp)
+file(GLOB_RECURSE H_FILES CONFIGURE_DEPENDS *.h)
+target_sources(Adera PRIVATE "${CPP_FILES}" "${H_FILES}")
+
+# Enforce conformance mode for adera
+target_compile_options(Adera PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
```
jonesmz, line 57: i recommend always leaving a trailing newline in files.

Otherwise unix based tools (and the github review viewer) give you angry red badges.

src/adera_drawing_gl/flat_shader.h
```diff
@@ -24,7 +24,7 @@
  */
 #pragma once
 
-#include <osp/drawing_gl/rendergl.h>
+#include <osp_drawing_gl/rendergl.h>
```
jonesmz, line 27: i guess i'm not following why you're changing the folder hierarchy like this?

is having a subdirectory causing a problem?

src/gdextension/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+
+target_compile_options(osp-gdextension PRIVATE "-static")
+
+
+# Install library, extension file, and support files in ${CMAKE_INSTALL_PREFIX}/${PROJECT_NAME}
+set( INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/osp-gdextension/" )
+
+add_subdirectory( templates )
+
+install( TARGETS osp-gdextension
```
jonesmz, line 69: is the "install" step needed?

.gitmodules
Lamakaio: I'll put the commit number, I'm pretty sure it's specified in the releases.

3rdparty/CMakeLists.txt
Lamakaio: Honestly, I had a link error due to positionning in spdlog, I saw there was a flag, I enabled it, it fixed the error. 
Then again, I'm not sure how this all works, so if there is a preferred solution that's perfectly fine.

CMakeLists.txt
Lamakaio: Oh I didn't know this was a thing, thank you. I just added it since it was in the template and the GdExtension takes a while to compile without it on my machine.

CMakeLists.txt
Lamakaio: I'm pretty sure this is something for the Jolt PR, it was what was causing the windows compile error in debug mode in the CI. Adding that line fixed it. Maybe something in Jolt breaks it somehow ... 

If that's the case, it might be better to upstream a fix in the long term.

src/adera_drawing_gl/flat_shader.h
Lamakaio: Mostly, I don't want to compile the drawing_gl part when building the GdExtension, to not depend on Magnum GL, SDL, and the like. 

I figured it would be clearer to separate the common code, magnum-standalone-specific code, and Godot code in different directories instead of hiding it down the folder hierarchy. Also saves quite a few CMakeLists.txt files.

src/gdextension/CMakeLists.txt
Lamakaio: GdExtension have a required folder structure, as well as a template that needs to be copied in the right place, and the install step generates it properly. 

Also, compiling with the cmake option to install the extension directly in you Godot project directory is super convenient imo. Godot reloads extensions immediately wh n the files change.

jonesmz:
Since you re-arranged your commits for the jolt physics PR, can you rebase this PR to use the re-arranged commits as parent, instead of the original version?

E.g. just rebase on master

Lamakaio:
> Since you re-arranged your commits for the jolt physics PR, can you rebase this PR to use the re-arranged commits as parent, instead of the original version?
> 
> E.g. just rebase on master

I made a new PR that removes newton, and rebased this PR on the previous one, as you had suggested. This should now be (mostly) clean.

Lamakaio:
Small progress report: I got all the resources (gltf files and meshes) to load, via Magnum, in Godot. Basically the `load_a_bunch_of_stuff` function.

Turns out Godot doesn't really support std::cout, std::cerr, and the like. Which makes basically every logging function crash the program (and sometimes the engine).

I kinda got around the issue by making Corrade logs go to stringstreams, that are then printed once per frame if they are not empty using the Godot print function. There might be a better way to do this for OSP logs, currently I just comment them when they cause an issue. 

My goal for now is really to get some examples working in the most ugly way, just to get an idea of what problems we might encounter with Godot.

src/gdextension/render.h
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * @brief OpenGL specific rendering components for rendering a scene
+ */
+struct ACtxSceneRenderGd
+{
+    MeshGdEntStorage_t     m_meshId;
+    TexGdEntStorage_t      m_diffuseTexId;
+    InstanceGdEntStorage_t m_instanceId;
+    godot::RID             m_scenario;
```
Capital-Asterisk, line 108: scene?

"scenario" more refers to a loadable scene

Lamakaio: It's mostly because the object behind the RID is a scenario for Godot. Which is not the same as a scene (that also exists in Godot). It has nothing to do with the OSP scenario, but I guess it should be more clear somehow.

src/gdextension/input.cpp
```diff
@@ -0,0 +1,232 @@
...
+        //Add user input
+        rUserInput.config_register_control(k, holdable, std::move(controls));
+    }
+}
+
+// Map for all the keys
+
+// Pair holds device and enum
+using button_pair_t = std::pair<int, int>;
+const std::map<std::string_view, button_pair_t, std::less<>> gc_buttonMap = {
```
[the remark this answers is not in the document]
Lamakaio: The goal was to use the same config file for both for now yes. Probably not needed though you're right.

src/gdextension/flying_scene.h
```diff
@@ -0,0 +1,95 @@
...
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.mainLoop);
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.inputs);
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.renderSync);
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.renderResync);
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.sceneUpdate);
+        m_testApp.m_pExecutor->signal(m_testApp, m_signals.sceneRender);
+    }
+
+protected:
+    static void _bind_methods();
```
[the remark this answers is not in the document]
Lamakaio: I'm not even sure it's in the macro, but the editor calls it when it loads the plugin to know which method it should call for what. Bound methods can also be called from gdscript.

Capital-Asterisk:
I thought I sent this earlier?

I already fixed all the things broken due to the framework changes. I have this branch that I just screw around in: https://github.com/Capital-Asterisk/osp-magnum/commit/1f5295786868c45b6ef24830aba122c838740a50 . This builds directly off of the gdextension-build branch, so you can just add my account as a git remote and `git reset --hard 1f52957`

Lamakaio:
I took your commits, added the aliasing fix, and axed all the useless rendering stuff (so now, osp is clearly just syncing it's entities with the rendered entities)

Need to look into why it doesn't build on windows anymore though. I think we had the same problem before ?

Lamakaio:
I think I'm mostly done (if the latest push passes CI) with what I want to include in this PR, so I'll mark it as ready. 

I cleaned up the code (which mostly means removed a lot of useless stuff here), addressed some very old comments, and solved all the errors and warning Godot was spouting out. 
Or almost, somehow SurfaceTool is leaking when it is by all means a local object. It might be a Godot bug. 

The only issue left is the total lack of color and general rendering prettyness. But as this PR is starting to be quite big already, it might be wiser to add that in a secondary PR ?

src/gdextension/render.cpp
```diff
@@ -0,0 +1,389 @@
...
+        }
+        if ( primitive == godot::RenderingServer::PRIMITIVE_TRIANGLES )
+        {
+            //st.deindex();
+            st.generate_normals();
+            //st.generate_tangents();
+            
+        }
+
+        godot::Array meshArray = st.commit_to_arrays();
```
Lamakaio, line 241: This is the SurfaceTool that's leaking. Calling `clear` does not change anything.

Capital-Asterisk: I presume a RID is being leaked? though I notice there's no mention of RIDs in `godot/scene/resources/surface_tool.cpp`

Capital-Asterisk: Unrelated to the leak: the way the indexed mesh is being turned into just vertices is throwing off the smooth shading. It looks like there's methods for directly adding an index buffer to SurfaceTool. I implied this somewhat in a previous comment, but it's likely that SurfaceTool isn't needed entirely.

Capital-Asterisk (approved):
> But as this PR is starting to be quite big already, it might be wiser to add that in a secondary PR

I think it's fair to get this merged quite soon, so others can start trying it out. Ideally, pull requests should be bite-sized; just getting a Godot window open would have been satisfactory.

Though honestly, I don't think we should care too much about master, since there's nothing at stake for screwing something up. IMO, these code reviews are best as learning oppurtunities more than anything in our case.

I can't seem to figure out how to get the framework changes out of the way to make this more readable; best solution I can find so far is using compare instead:

https://github.com/TheOpenSpaceProgram/osp-magnum/compare/master...Lamakaio:osp-magnum:gdextension-build

src/gdextension/render.cpp
Lamakaio: It's not a RID, it's the object itself. There is apparently an "ObjectDB" in Godot, and it detects that all the surface tool used to create meshes are still allocated when the app closes. 

I'll try tinkering it a bit, but I remember getting weird results when using other ways to create meshes.

Lamakaio: > Unrelated to the leak: the way the indexed mesh is being turned into just vertices is throwing off the smooth shading. It looks like there's methods for directly adding an index buffer to SurfaceTool. I implied this somewhat in a previous comment, but it's likely that SurfaceTool isn't needed entirely.

It turns out not ignoring normals fixes the smooth shading. Who could have thought.

PR #296, opened 2024-06-24
Remove newton dynamics from the project

Lamakaio:
This PR simply removes Newton Dynamics from the project. It is unused since the Jolt integration. It is not intended to be merged immediately, but is a useful base to start tinkering with gdextension (see #294, that depends on this PR).

I included a small fix for center of mass (again) that slipped through the Jolt PR. It's literally 1 line, and doesn't affect much, so I don't think it warrants its own commit, but if it does, I'll split it.


jonesmz (approved):
I really have no feedback. Looks good to me, and good riddance.

Lamakaio:
It got integrated in the godot PR, so no use rebasing it here.

PR #300, opened 2024-09-27
Fix aliasing issue in framework

Lamakaio:
A proposal for a fix of aliasing issues in framework, using some memcpy calls. They should be optimized away because the compiler is smart. 

bit_cast also exists, but apparently it cannot be used to cast between arrays with elements of different size. 

I bundled a small fix for jolt that was still pending, because I don't want to make a whole pull request for 1 line.

### DIFF
--- a/src/osp/framework/framework.h
+++ b/src/osp/framework/framework.h
@@ -32,6 +32,7 @@
 #include "../core/keyed_vector.h"
 #include "../tasks/tasks.h"
 
+#include <cstring>
 #include <entt/core/any.hpp>
 
 #include <Corrade/Containers/ArrayViewStl.h>
@@ -336,13 +337,17 @@ struct Framework
             // Access struct members of out.pl and out.di as if they're arrays, in order to write
             // DataIds and PipelineIds to them.
 
+            //Using this reference to avoid strict aliasing UB : https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8
+            //Apparently the memcpy calls get optimized away. 
+
             FeatureInterface const &rInterface = m_fiinstData[fiId];
+
             if constexpr ( ! std::is_empty_v<typename FI_T::Pipelines> )
             {
                 auto const plMembers = arrayCast<PipelineDefBlank_t>(  arrayCast<std::byte>( arrayView(&out.pl, 1) )  );
                 for (std::size_t i = 0; i < plMembers.size(); ++i)
                 {
-                    plMembers[i].m_value = rInterface.pipelines[i];
+                    std::memcpy(&plMembers[i].m_value,  &rInterface.pipelines[i], sizeof(PipelineId));
                 }
             }
 
@@ -351,7 +356,7 @@ struct Framework
                 auto const diMembers = arrayCast<DataId>(  arrayCast<std::byte>( arrayView(&out.di, 1) )  );
                 for (std::size_t i = 0; i < diMembers.size(); ++i)
                 {
-                    diMembers[i] = rInterface.data[i];
+                    std::memcpy(&diMembers[i], &rInterface.data[i], sizeof(DataId));
                 }
             }
         }


Lamakaio:
It looks like a problem with the way I compute the sizes of intermediate array, which triggers static analysis somehow. Not sure what to do about it.

src/adera_app/features/jolt.cpp
```diff
@@ -623,7 +623,7 @@ static void rocket_thrust_force(BodyId const bodyId, ACtxJoltWorld const& rJolt,
 
     JPH::BodyID joltBodyId = BToJolt(bodyId);
     Quaternion const rot = QuatJoltToMagnum(bodyInterface.GetRotation(joltBodyId));
-    RVec3 joltCOM = bodyInterface.GetCenterOfMassPosition(joltBodyId) - bodyInterface.GetPosition(joltBodyId);
+    RVec3 joltCOM = bodyInterface.GetPosition(joltBodyId) - bodyInterface.GetCenterOfMassPosition(joltBodyId);
```
Capital-Asterisk, line 626: For the pendulum effect the rockets are having?

The issue here is that GetCenterOfMassPosition takes (world) rotation in account when it shouldn't. I worked this out on https://github.com/Capital-Asterisk/osp-magnum/commit/463fe6e553fc5ffe2a4c192de964fcf13575836f , part of the commit I sent on the gdextension pull request. This can just be cherry-picked.

src/osp/framework/framework.h
```diff
@@ -336,23 +337,35 @@ struct Framework
...
             FeatureInterface const &rInterface = m_fiinstData[fiId];
+
             if constexpr ( ! std::is_empty_v<typename FI_T::Pipelines> )
             {
-                auto const plMembers = arrayCast<PipelineDefBlank_t>(  arrayCast<std::byte>( arrayView(&out.pl, 1) )  );
-                for (std::size_t i = 0; i < plMembers.size(); ++i)
+                constexpr size_t members_size = sizeof(typename FI_T::Pipelines)/sizeof(PipelineDefBlank_t);
+                PipelineDefBlank_t plMembers[members_size] ;
+                std::memcpy(plMembers, &out.pl, sizeof(typename FI_T::Pipelines));                
+                //auto const plMembers = arrayCast<PipelineDefBlank_t>(  arrayCast<std::byte>( arrayView(&out.pl, 1) )  );
```
Capital-Asterisk, line 350: no need to keep commented out code

Capital-Asterisk (commented):
The [old code with the funny comment](https://github.com/TheOpenSpaceProgram/osp-magnum/blob/fa254e85bfda3f6b6d521b55bd8e2559bd0e501d/src/osp/tasks/builder.h#L101C1-L119) worked because it worked with an `unsigned char` array first, then used it to initialize a struct afterwards.

This memcpy approach looks a little cleaner. 

About bit_cast, an alternative approach may be to use arrays of PipelineDefBlank_t that are bit_casted at the end with `return  { .pl = std::bit_cast<typename FI_T::Pipelines>(...) ... };`. (bit_casting an array to a struct of the same size)

src/osp/framework/framework.h
```diff
@@ -336,23 +337,35 @@ struct Framework
...
+                for (std::size_t i = 0; i < members_size; ++i)
                 {
                     plMembers[i].m_value = rInterface.pipelines[i];
```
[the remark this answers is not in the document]
Lamakaio: I could test it.
I thought the cast to ArrayView was the problem. The compiler might understand a memcpy of size 4 exactly the same way it does a simple assignment. 

What makes it work here for me is that the cast from opaque struct to array is done through memcpy and not ArrayView ?

src/adera_app/features/jolt.cpp
Lamakaio: Oh well perfect then.

Lamakaio:
I changed the memcpy to just replace the array assignement. Is this what you suggested @Capital-Asterisk ? 

If that looks good I'll rewrite the commit history to remove the Jolt stuff and get a single commit.

src/osp/framework/framework.h
```diff
@@ -336,22 +337,26 @@ struct Framework
...
                 for (std::size_t i = 0; i < plMembers.size(); ++i)
                 {
-                    plMembers[i].m_value = rInterface.pipelines[i];
+                    memcpy(&plMembers[i].m_value,  &rInterface.pipelines[i], sizeof(PipelineId));
                 }
             }
 
             if constexpr ( ! std::is_empty_v<typename FI_T::DataIds> )
-            {
+            {
```
Capital-Asterisk, line 355: very minor: excess spacing

src/osp/framework/framework.h
```diff
@@ -336,22 +337,26 @@ struct Framework
...
+
             FeatureInterface const &rInterface = m_fiinstData[fiId];
+
             if constexpr ( ! std::is_empty_v<typename FI_T::Pipelines> )
             {
                 auto const plMembers = arrayCast<PipelineDefBlank_t>(  arrayCast<std::byte>( arrayView(&out.pl, 1) )  );
                 for (std::size_t i = 0; i < plMembers.size(); ++i)
                 {
-                    plMembers[i].m_value = rInterface.pipelines[i];
+                    memcpy(&plMembers[i].m_value,  &rInterface.pipelines[i], sizeof(PipelineId));
```
Capital-Asterisk, line 350: minor: std::memcpy. iirc, this is C's memcpy from `<string.h>`, instead of `<cstring>`

Capital-Asterisk (commented):
> Is this what you suggested @Capital-Asterisk ?

looks good!

Lamakaio:
Should be good !

Capital-Asterisk (approved):

PR #310, opened 2025-11-01
Fix and clean up some framework and Jolt integration issues

Capital-Asterisk:
Hit some serious issues while trying to request floating origin translations and creating/deleting entities on the same frame. This is a tough case for game engines and simulations in general, prone to bugs involving 'flickering' or accidentally spawning objects or doing calculations using old positions.

OSP framework is specifically designed to handle these cases elegantly, but just need some tweaking to get it right.

### EStgCont and pipeline changes

old: Delete → Resize → New → Modify → ScheduleC → Ready → ReadyWorkaround → ...loop
new: Modify → ReadyB4New → Delete → Resize → New → Ready → ...loop

* Removed weird workaround. I don't remember what this was for, but it's not needed anymore
* Made 'Schedule' task non-mandatory

#### comment from `src/adera_app/feature_interfaces.h`:

Continuous Containers, data that persists and is modified over time
```cpp
enum class EStgCont : uint8_t
{
    Modify      = 0,    ///< Modify existing elements
    ReadyB4New  = 1,    ///< Container is ready to use after modify, but before adding new elements
    Delete      = 2,    ///< Remove elements from a container
    Resize_     = 3,    ///< Resize the container to fit more elements
    New         = 4,    ///< Add new elements
    Ready       = 5,    ///< Container is ready to use
};
```
This pipeline is intended to organize access to variables associated with entities that can be
added and removed.

Things that were considered (about game engines and simulations in general):
* When a new entity is created, it should NOT be moved (physics updated) right away before it is first rendered. It should not appear offset from where it was requested to be initialized.
* It shouldn't be possible to create and immediately delete an entity on the same frame before it is rendered. The entity should live at least one frame for other systems to be able to process and respond to entities being added or removed.
* Creating an entity and simultaneously requesting a floating origin translation is tough case. Make sure all positions are translated before requesting to spawn new entities.

More explanations on the current order:
* Modify comes before Delete and New, because it's janky to modify elements directly after they're added. It only cares about 'what currently exists'.
* ReadyB4New is there to read results of Modify, without considering new or deleted elements. Elements can be requested to be deleted here. This also handles the case for requesting a floating origin translation like mentioned above.
* Delete runs before New, because Delete makes empty spaces in buffers for new elements to fill
* Resize comes before New, to resize/reallocate containers to a known new maximum size before adding new elements.

### task order report readability improvements:

image
see #305 for previous appearance

> This is a timeline of Tasks with dependencies based on which variables they access (read/write/other stuff). This determine what order they're suppose to run in (i.e.: all tasks that read from a queue must wait until all tasks that write to it are finished). Names of the tasks are on the right, and the time axis goes downwards.

### other stuff:
* Removed `using namespace JPH`
* Fixed cylinder jolt collider being too big
* Fixed rocket jolt torque calculations not handling rotations correctly

### DIFF
--- a/src/ospjolt/activescene/joltinteg.h
+++ b/src/ospjolt/activescene/joltinteg.h
@@ -37,19 +37,22 @@ JPH_SUPPRESS_WARNING_PUSH
 JPH_SUPPRESS_WARNINGS
 
 #include <Jolt/RegisterTypes.h>
+
+#include <Jolt/Physics/Body/BodyCreationSettings.h>
+#include <Jolt/Physics/Collision/BroadPhase/BroadPhaseLayer.h>
+#include <Jolt/Physics/Collision/Shape/BoxShape.h>
+#include <Jolt/Physics/Collision/Shape/CompoundShape.h>
+#include <Jolt/Physics/Collision/Shape/CylinderShape.h>
+#include <Jolt/Physics/Collision/Shape/MutableCompoundShape.h>
+#include <Jolt/Physics/Collision/Shape/RotatedTranslatedShape.h>
+#include <Jolt/Physics/Collision/Shape/ScaledShape.h>
+#include <Jolt/Physics/Collision/Shape/SphereShape.h>
+#include <Jolt/Physics/PhysicsStepListener.h>
+#include <Jolt/Physics/PhysicsSystem.h>
+
 #include <Jolt/Core/Factory.h>
 #include <Jolt/Core/JobSystemThreadPool.h>
-#include <Jolt/Physics/PhysicsSystem.h>
-#include <Jolt/Physics/Collision/Shape/BoxShape.h>
-#include <Jolt/Physics/Collision/Shape/SphereShape.h>
-#include <Jolt/Physics/Collision/Shape/CylinderShape.h>
-#include <Jolt/Physics/Collision/Shape/RotatedTranslatedShape.h>
-#include <Jolt/Physics/Body/BodyCreationSettings.h>
 #include <Jolt/Core/TempAllocator.h>
-#include <Jolt/Physics/Collision/Shape/CompoundShape.h>
-#include <Jolt/Physics/Collision/Shape/ScaledShape.h>
-#include <Jolt/Physics/Collision/Shape/MutableCompoundShape.h>
-#include <Jolt/Physics/PhysicsStepListener.h>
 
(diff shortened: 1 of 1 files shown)


src/adera_app/feature_interfaces.h
```diff
@@ -76,56 +76,63 @@ inline osp::PipelineTypeInfo const gc_infoForEStgIntr
...
-    ScheduleC   = 4,
-
-    Ready       = 5,
-    ///< Container is ready to use
-
-    ReadyWorkaround = 6,
+    Modify      = 0,    ///< Modify existing elements
+    ReadyB4New  = 1,    ///< Container is ready to use after modify, but before adding new elements
+    Delete      = 2,    ///< Remove elements from a container
+    Resize_     = 3,    ///< Resize the container to fit more elements
```
jonesmz: Why the underscore?

src/adera_app/feature_interfaces.h
```diff
@@ -76,56 +76,63 @@ inline osp::PipelineTypeInfo const gc_infoForEStgIntr
...
+ * * It shouldn't be possible to create and immediately delete an entity on the same frame before
+ *   it is rendered. The entity should live at least one frame for other systems to be able to
+ *   process and respond to entities being added or removed.
+ * * Creating an entity and simultaneously requesting a floating origin translation is tough case.
+ *   Make sure all positions are translated before requesting to spawn new entites.
+ *
+ * More explainations on the current order:
+ * * Modify comes before Delete and New, because it's janky to modify elements directly after
+ *   they're added. It only cares about 'what currently exists'.
+ * * ReadyB4New is there to read results of Modify, without considering new or deleted elements.
```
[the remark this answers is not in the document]
Capital-Asterisk: I'm not sure of the best way to explain it, but there's now two ready states. A few tasks that use to be synced with 'Ready' are changed to 'ReadyB4New'; it's been split into two. It's still 'ready to use' and the associated container stays constant.

src/adera_app/feature_interfaces.h
Capital-Asterisk: conflict with another enum named `Resize`. These are just `using enum`-ed in for readability. I thought about making individual 'stages' global variables instead, where each pipeline can list off which stages they use, so they can share the same stage names. The metaprogramming would be harder though.

